PR #2, opened 2015-09-11
mapi-964 use "update or create"

ScalaWilliamTG:


alexduf:
Looks good to me :+1:

PR #6, opened 2015-09-18
Mapi 945 platform generic

davidfurey:


common/src/main/scala/models/Platform.scala
```diff
@@ -0,0 +1,26 @@
...
+import play.api.libs.json._
+
+sealed trait Platform
+
+case object Android extends Platform { override def toString = "android" }
+case object iOS extends Platform { override def toString = "ios" }
+case object WindowsMobile extends Platform { override def toString = "windows-mobile" }
+
+object Platform {
+  def fromString(s: String): Option[Platform] = s match {
```
alexduf: You either want to drop the Option and rely on the exception when it does not match, or you're mission case _ => None (or PartialFunction.condOpt)

alexduf:
Cool! Looks good to me :+1:

PR #7, opened 2015-09-18
Improve hub client

davidfurey:


common/src/main/scala/gu/msnotifications/RawWindowsRegistration.scala
```diff
@@ -1,8 +1,17 @@
 package gu.msnotifications
 
+import models.MobileRegistration
 /**
  * [[https://msdn.microsoft.com/en-us/library/microsoft.servicebus.notifications.windowsregistrationdescription.aspx]]
  */
+object RawWindowsRegistration {
+  def fromMobileRegistration(m: MobileRegistration) = {
```
agulowaty: `mr` intead `m` ? :smile:

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
```diff
@@ -22,33 +20,45 @@ final class NotificationHubClient(notificationHubConnection: NotificationHubConn
...
       .withHeaders("X-WNS-Type" -> azureWindowsPush.wnsType)
       .withHeaders("ServiceBusNotification-Format" -> "windows")
       .withHeaders(serviceBusTags: _*)
       .post(azureWindowsPush.xml)
-      .map(processResponse).map(_.map(_ => ()))
+      .map { response =>
+        if (response.status == 200)
+          \/.right(())
+        else
+          \/.left(HubFailure.fromResponseCode(response.status))
```
alexduf: I would play safe and re-use the status message from the response instead of deducing it from the code. I know it's supposed to be fairly standard but you never know.

common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -0,0 +1,93 @@
...
+        )
+      else
+        HubParseFailed(
+          body = response.body,
+          reason = "Failed to find any XML"
+        )
+    }
+  }
+
+  def parse[T](response: WSResponse)(implicit reader: XmlReads[T]): HubResult[T] =
```
alexduf: :100: Clever

alexduf:
Aside from the comment about the response code, that PR looks good to me :+1:

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
```diff
@@ -22,33 +20,45 @@ final class NotificationHubClient(notificationHubConnection: NotificationHubConn
...
 
-    request(s"$notificationsHubUrl/messages/?api-version=2015-01")
+    request(s"/messages/")
       .withHeaders("X-WNS-Type" -> azureWindowsPush.wnsType)
       .withHeaders("ServiceBusNotification-Format" -> "windows")
       .withHeaders(serviceBusTags: _*)
       .post(azureWindowsPush.xml)
-      .map(processResponse).map(_.map(_ => ()))
+      .map { response =>
+        if (response.status == 200)
```
agulowaty: are we sure azure service will always return 200 on post ? or maybe it'll be 201 Created  which is also .right case for us.

notification/app/controllers/Main.scala
```diff
@@ -0,0 +1,22 @@
...
+import play.api.Logger
+import play.api.libs.ws.WSClient
+import play.api.mvc.{Action, Controller}
+import services.NotificationConfiguration
+import scala.concurrent.ExecutionContext
+
+final class Main @Inject()(wsClient: WSClient,
+  msNotificationsConfiguration: NotificationConfiguration)(
+  implicit executionContext: ExecutionContext
+  ) extends Controller {
```
agulowaty: this formatting looks weird

common/src/main/scala/models/Platform.scala
```diff
@@ -0,0 +1,26 @@
...
+
+case object Android extends Platform { override def toString = "android" }
+case object iOS extends Platform { override def toString = "ios" }
+case object WindowsMobile extends Platform { override def toString = "windows-mobile" }
+
+object Platform {
+  def fromString(s: String): Option[Platform] = s match {
+    case "android" => Some(Android)
+    case "ios" => Some(iOS)
+    case "windows-mobile" => Some(WindowsMobile)
```
agulowaty: case _ => None missing ? else branch on line 20 is probably dead code since MatchError will be thrown instead None

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
davidfurey: Based on their docs, it should return 200:
https://msdn.microsoft.com/en-us/library/azure/dn223272.aspx

notification/app/controllers/Main.scala
davidfurey: Happy to change this if you suggest an improvement

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
davidfurey: The API docs implied that there would never be a response body, but I've re-read and I think it might, so I will change this as you suggest

agulowaty: :ok_hand:

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
agulowaty: :ok_hand:

notification/app/controllers/Main.scala
agulowaty:
```
final class Main @Inject()(wsClient: WSClient, msNotificationsConfiguration: NotificationConfiguration)
  (implicit executionContext: ExecutionContext) 
  extends Controller {
```

davidfurey: Thanks

agulowaty:
:+1:

alexduf:
Still looking good to me too :+1:

PR #8, opened 2015-09-18
Split registration and push into separate services

davidfurey:


notification/app/controllers/Main.scala
```diff
@@ -0,0 +1,22 @@
...
+import services.NotificationConfiguration
+import scala.concurrent.ExecutionContext
+
+final class Main @Inject()(wsClient: WSClient,
+  msNotificationsConfiguration: NotificationConfiguration)(
+  implicit executionContext: ExecutionContext
+  ) extends Controller {
+
+
+  private val logger = Logger("main")
```
alexduf: I would either remove that unused variable, or use Logger(this.getClass) instead of a string. That would avoid having a logger copied/pasted with the wrong name.

notification/conf/bootstrap.conf
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
```
alexduf: Should that file be named bootstrap.sh instead of bootstrap.conf?

PR #11, opened 2015-10-01
Multi provider prep

davidfurey:


common/src/main/scala/models/Notification.scala
```diff
@@ -0,0 +1,14 @@
+package models
```
agulowaty: Notification-related models shouldn't be in `common` project, they are not relevant to `registration` module

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
```diff
@@ -15,11 +16,13 @@ object NotificationHubClient {
  * https://msdn.microsoft.com/en-us/library/azure/dn223264.aspx
  */
 final class NotificationHubClient(notificationHubConnection: NotificationHubConnection, wsClient: WSClient)
-                                 (implicit executionContext: ExecutionContext) {
+                                 (implicit executionContext: ExecutionContext) extends NotificationProvider {
 
   import notificationHubConnection._
   import NotificationHubClient.HubResult
 
+  val name = "WNS"
```
agulowaty: naming convention for constant vals is to UpperCamelCase so it should be `Name` instead

common/src/main/scala/notifications/providers/NotificationProvider.scala
```diff
@@ -0,0 +1,11 @@
+package notifications.providers
+
+import models.Push
+
+import scala.concurrent.Future
+import scalaz.\/
+
+trait NotificationProvider {
```
agulowaty: *Provider suffix is unfortunate - it's the interface of NotificationSender not a provider

common/src/main/scala/models/Notification.scala
davidfurey: NotificationHubClient is in common, and it needs the notification model

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
davidfurey: name does not fit the style guides definition of a constant, since it is a member of a trait

alexduf:
Looking good :+1:

common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
agulowaty: ok

agulowaty:
:+1:

PR #12, opened 2015-10-02
MAPI-979: /registrations endpoint prep

agulowaty:
- migrating to spec2
- Introduced separated interface for register() action on NotificationHubClient and companion case classes for interface.
- specs2 play support added
- extended Registration with set of Topics
- renamed routes for registrations


davidfurey:
:+1:

PR #13, opened 2015-10-02
Add reporting to notification push

davidfurey:
In-memory only for now.  Also need to add topic registration tracking
to registration end-point


common/src/main/scala/tracking/InMemoryNotificationReportRepository.scala
```diff
@@ -0,0 +1,29 @@
...
+    Future.successful(\/.right(()))
+  }
+
+  def getByUuid(uuid: String): Future[RepositoryResult[NotificationReport]] = {
+    Future.successful(db.find(_.notification.uuid == uuid) \/> RepositoryError("Notification report not found"))
+  }
+
+  def getByDateRange(from: DateTime, to: DateTime): Future[RepositoryResult[List[NotificationReport]]] = {
+    Future.successful(\/.right(db.filter({report =>
+      (report.sentTime isAfter from) && (report.sentTime isBefore to)
```
alexduf: I think this filter should be inclusive, as in "from <= report.sentTime <= to "

notification/app/services/NotificationConfiguration.scala
```diff
@@ -32,7 +33,10 @@ final class NotificationConfiguration @Inject()(wsClient: WSClient)(implicit exe
 
   private def notificationHub = notificationHubOR.fold(error => throw new Exception(error.message), identity)
 
-  def notificationHubClient = new NotificationHubClient(notificationHub, wsClient)
+  val notificationReportRepository = new InMemoryNotificationReportRepository
```
alexduf: If we want to have a state on that object, I would suggest that NotificationConfiguration use the @Singleton tag.
That being said I would prefer our state to be outside of the configuration (and ideally injected)

notification/app/services/NotificationConfiguration.scala
```diff
@@ -32,7 +33,10 @@ final class NotificationConfiguration @Inject()(wsClient: WSClient)(implicit exe
...
+  val topicSubscriptionRepository = new InMemoryTopicSubscriptionsRepository
```
alexduf: same as above

registration/app/services/RegistrationConfiguration.scala
```diff
@@ -32,7 +33,8 @@ final class RegistrationConfiguration @Inject()(wsClient: WSClient)
 
   private def notificationHub = notificationHubOR.fold(error => throw new Exception(error.message), identity)
 
-  def notificationHubClient = new NotificationHubClient(notificationHub, wsClient)
+  val topicSubscriptionRepository = new InMemoryTopicSubscriptionsRepository
```
alexduf: Same again

alexduf:
Following our conversation, the configuration object will be refactored as a separate PR. :+1:

common/src/main/scala/tracking/InMemoryNotificationReportRepository.scala
agulowaty: http://joda-time.sourceforge.net/apidocs/org/joda/time/ReadableInterval.html#contains(org.joda.time.ReadableInstant) can be used instead isAfter and isBefore

davidfurey: Thanks for that @agulowaty

PR #14, opened 2015-10-02
First pass at configuration refactor

davidfurey:


registration/app/controllers/Main.scala
```diff
@@ -3,49 +3,29 @@ package controllers
...
 import play.api.Logger
 import play.api.libs.json.Json
 import play.api.mvc.{Action, BodyParsers, Controller, Result}
 import services._
-import scala.async.Async
 import scala.concurrent.{Future, ExecutionContext}
-import scalaz.{-\/, \/-}
+import scalaz.{\/, -\/, \/-}
 import BodyParsers.parse.{json => BodyJson}
+import notifications.providers.{Error => ProviderError}
```
agulowaty: don't see benefit in introducing that alias

notification/app/controllers/Main.scala
```diff
@@ -2,25 +2,30 @@ package controllers
...
 import services._
-import tracking.InMemoryNotificationReportRepository
 import scala.concurrent.ExecutionContext
 import BodyParsers.parse.{json => BodyJson}
 import scala.concurrent.Future
 import scalaz.{-\/, \/-}
 
-final class Main @Inject()(wsClient: WSClient, msNotificationsConfiguration: NotificationConfiguration)
+final class Main @Inject()(
+  wsClient: WSClient,
```
agulowaty: this dependency shouldn't be injected here

registration/app/services/NotificationRegistrarSupport.scala
```diff
@@ -0,0 +1,19 @@
...
+import javax.inject.Inject
+
+import notifications.providers.{NotificationRegistrar, WindowsNotificationProvider}
+import play.api.libs.ws.WSClient
+
+import scala.concurrent.ExecutionContext
+
+final class NotificationRegistrarSupport @Inject()(wsClient: WSClient, configuration: Configuration)(implicit executionContext: ExecutionContext) {
+
+  val providerConf = configuration.notificationHubConfiguration.fold(error => throw new Exception(error.message), identity)
```
agulowaty: I don't think we should expose it as public, controllers shouldn't have an access here

registration/app/controllers/Main.scala
davidfurey: "Error" felt like a very generic class to import, but I am happy to remove the alias if you don't see the benefit

alexduf: "Error" is indeed a very generic name. What about importing the package and using providers.Error whenever needed?

notification/app/services/Configuration.scala
```diff
@@ -0,0 +1,34 @@
...
+case class ErrorMessage(message: String)
+
+final class Configuration @Inject()()(implicit executionContext: ExecutionContext) {
+
+  private lazy val conf = ConfigurationFactory.getConfiguration(
+    applicationName = "notification",
+    webappConfDirectory = "gu-conf"
+  )
+
+  private def getConfigurationProperty(name: String): ErrorMessage \/ String =
```
alexduf: I doubt any part of the application should work if it's not configured properly, should we just throw an exception here instead of returning Either and letting every client throw their own exception?

registration/app/controllers/Main.scala
agulowaty: If it's too generic why not to rename it in place i.e. in its own definition ? ;)

registration/app/services/Configuration.scala
```diff
@@ -0,0 +1,32 @@
...
+  connectionString: String,
+  hubName: String
+)
+
+case class ErrorMessage(message: String)
+
+final class Configuration @Inject()()(implicit executionContext: ExecutionContext) {
+
+  private lazy val conf = ConfigurationFactory.getConfiguration(
+    applicationName = "notification",
```
alexduf: "registration" ? or are we sharing the same config file (which is actually not a bad idea)

agulowaty:
:+1:

alexduf:
:+1:

PR #15, opened 2015-10-05
Mapi 902 reports endpoints

davidfurey:
In-memory database for now - persistence still to do.


report/test/controllers/ReportIntegrationSpec.scala
```diff
@@ -0,0 +1,99 @@
...
+import play.api.inject.guice.GuiceApplicationBuilder
+import play.api.inject.bind
+import play.api.test._
+import services.{NotificationReportRepositorySupport, Configuration}
+import tracking.InMemoryNotificationReportRepository
+import scala.concurrent.ExecutionContext.Implicits.global
+
+class ReportIntegrationSpec extends PlaySpecification with Mockito {
+
+  trait ReportTestScope extends Scope {
```
agulowaty: I'd move test scope to the bottom, test definitions are more important than setup

report/app/services/Configuration.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.gu.conf.ConfigurationFactory
+
+import scala.concurrent.ExecutionContext
+
+case class NotificationHubConfiguration(
+  connectionString: String,
+  hubName: String
+)
+
+class Configuration @Inject()()(implicit executionContext: ExecutionContext) {
```
alexduf: I may have missed something but I think we don't need the ExecutionContext here

alexduf:
Aside from a minor comment, it looks good to me :+1:

PR #16, opened 2015-10-06
MAPI-983 - registrations create and update

agulowaty:
Implementing registration and registration update with tags.
w/o topics as for now

### DIFF
--- a/common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
+++ b/common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
@@ -1,11 +1,6 @@
 package gu.msnotifications
 
-import models._
-import org.joda.time.DateTime
-import notifications.providers.{NotificationRegistrar, NotificationSender, RegistrationResponse => RegistrarResponse}
 import play.api.libs.ws.WSClient
-import tracking.Repository.RepositoryResult
-import tracking.{RepositoryResult, TopicSubscriptionsRepository}
 
 import scala.concurrent.{ExecutionContext, Future}
 import scalaz.\/
@@ -25,8 +20,13 @@ final class NotificationHubClient(
   import notificationHubConnection._
   import NotificationHubClient.HubResult
 
-  def register(rawWindowsRegistration: RawWindowsRegistration): Future[HubResult[RegistrationResponse]] = {
-    request("/registrations/")
+  private object Endpoints {
+    val Registrations = "/registrations"
+    val Messages = "/messages/"
+  }
+
+  def create(rawWindowsRegistration: RawWindowsRegistration): Future[HubResult[RegistrationResponse]] = {
+    request(Endpoints.Registrations)
       .post(rawWindowsRegistration.toXml)
       .map(XmlParser.parse[RegistrationResponse])
   }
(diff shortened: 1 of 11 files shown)


common/src/main/scala/gu/msnotifications/NotificationHubClient.scala
```diff
@@ -25,8 +20,13 @@ final class NotificationHubClient(
   import notificationHubConnection._
   import NotificationHubClient.HubResult
 
-  def register(rawWindowsRegistration: RawWindowsRegistration): Future[HubResult[RegistrationResponse]] = {
-    request("/registrations/")
+  private object Endpoints {
+    val Registrations = "/registrations"
```
davidfurey: I think this is missing a trailing slash

common/src/main/scala/gu/msnotifications/package.scala
```diff
@@ -30,4 +30,15 @@ package object msnotifications {
...
+
+  case class Tags(tags: Set[String] = Set.empty) {
+    import Tags._
+    def asSet = tags
+    def findUserId: Option[UserId] = tags.find(_.startsWith(UserTagPrefix)).map(UserId(_))
+  }
+
+  object Tags {
+    val UserTagPrefix = "user:"
+    def withUserId(userId: UserId) = Tags(Set(s"${UserTagPrefix}${userId.userId}}"))
```
davidfurey: Extra } at end of string

common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -65,7 +66,11 @@ object RegistrationResponse {
   }
 }
 case class RegistrationResponse(registration: WNSRegistrationId, tags: List[String], channelUri: String, expirationTime: DateTime) {
-  def toRegistrarResponse : RegistrarResponse = ???
+  def toRegistrarResponse : RegistrarResponse = RegistrarResponse(
+    deviceId = channelUri,
+    WindowsMobile,
+    userId = Tags(tags.toSet).findUserId.get
```
davidfurey: Option.get is bad.  Perhaps toRegistrarResponse could return an Option[RegistrarResponse].  No need for a space between method name and :

common/src/main/scala/notifications/providers/WindowsNotificationProvider.scala
```diff
@@ -23,10 +25,23 @@ class WindowsNotificationProvider(wsClient: WSClient, connectionString: String,
...
+
+    def updateRegistration(regId: WNSRegistrationId) = hubClient
+      .update(regId, RawWindowsRegistration.fromMobileRegistration(registration))
+      .map(hubResultToRegistrationResponse)
+
+    val channelUri = registration.deviceId
+    hubClient.registrationsByChannelUri(channelUri).flatMap {
+      case \/-(Nil) => createNewRegistration
+      case \/-(existing :: Nil) => updateRegistration(existing.registration)
+      case \/-(List(_, _)) => Future { TooManyRegistrationsForChannel(channelUri).left }
```
davidfurey: You could replace List(_, _) with _
Or with first :: second :: tail
Instead of wrapping TooManyRegistrations in a Future, use Future.successful(...)
same for e.left

common/src/main/scala/gu/msnotifications/Responses.scala
agulowaty: I didn't want to make it Optional - RegistrationResponse without userId tag inside is considered as serious problem and I want to throw Exception here if you don't mind.

common/src/main/scala/notifications/providers/WindowsNotificationProvider.scala
agulowaty: thanks for that

common/src/main/scala/gu/msnotifications/Responses.scala
davidfurey: In that case, return Error \/ RegistrarResponse.  Encoding the possibility of error into the type forces the caller into handling the unexpected behaviour whereas a theoretical exception can be easily ignored

davidfurey:
:+1:

PR #18, opened 2015-10-07
Mapi 986 Report Persistence

davidfurey:
Adds a Dynamodb implementation of NotificationReportRepository

### DIFF
--- a/common/src/main/scala/aws/AsyncDynamo.scala
+++ b/common/src/main/scala/aws/AsyncDynamo.scala
@@ -17,6 +17,10 @@ object AsyncDynamo {
   def keyLT(s: String) = new Condition()
     .withComparisonOperator(ComparisonOperator.LT)
     .withAttributeValueList(new AttributeValue(s))
+
+  def keyBetween(a: String, b: String) = new Condition()
+    .withComparisonOperator(ComparisonOperator.BETWEEN)
+    .withAttributeValueList(new AttributeValue(a), new AttributeValue(b))
 }
 
 class AsyncDynamo(client: AmazonDynamoDBAsyncClient) {
--- a/common/src/main/scala/aws/AsyncDynamo.scala
+++ b/common/src/main/scala/aws/AsyncDynamo.scala
@@ -1,0 +1,27 @@
+package aws
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDBAsyncClient
+import com.amazonaws.services.dynamodbv2.model._
+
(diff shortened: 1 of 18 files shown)


common/src/test/scala/tracking/DynamoNotificationReportRepositoryTest.scala
```diff
@@ -0,0 +1,132 @@
...
+import org.specs2.mock.Mockito
+import org.specs2.mutable.Specification
+import org.specs2.specification.BeforeAfterAll
+import org.specs2.specification.Scope
+import tracking.Repository.RepositoryResult
+
+import scala.collection.JavaConversions._
+import scala.concurrent.Future
+
+class DynamoNotificationReportRepositoryTest(implicit ev: ExecutionEnv) extends Specification with Mockito with ShouldMatchers with BeforeAfterAll {
```
agulowaty: Suffix should be *Spec instead according to our convention

common/src/main/scala/aws/AsyncDynamo.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  def keyBetween(a: String, b: String) = new Condition()
+    .withComparisonOperator(ComparisonOperator.BETWEEN)
+    .withAttributeValueList(new AttributeValue(a), new AttributeValue(b))
+}
+
+class AsyncDynamo(client: AmazonDynamoDBAsyncClient) {
+  import AWSAsync._
+  // These work but are red because Intellij is broken
+  def putItem(putItemRequest: PutItemRequest): Future[PutItemResult] = wrapAsyncMethod(client.putItemAsync, putItemRequest)
```
agulowaty: `putItem` 3x in one line, it could be rewritten to:

```
def put(request: PutItemRequest)
```

Minor, but less noise introduced

common/src/test/scala/tracking/DynamoNotificationReportRepositoryTest.scala
agulowaty: Why ShouldMatchers are imported ?

common/src/main/scala/aws/AsyncDynamo.scala
davidfurey: I'll change the variable name, but not the function name because I want to keep AsyncDynamo similar to the underlying AmazonDynamoDBAsyncClient.  It is supposed to be a thin wrapper

agulowaty:
:+1:

PR #19, opened 2015-10-07
Mapi 900 registration backup

alexduf:
Adds support for backups

### DIFF
--- a/backup/src/main/scala/Boot.scala
+++ b/backup/src/main/scala/Boot.scala
@@ -34,5 +34,5 @@ object Boot {
     }
   }
 
-  def main (args: Array[String]): Unit = runBatch[Backup]("./backup")
+  def main (args: Array[String]): Unit = runBatch[Backup](args.headOption.getOrElse("./backup"))
 }
--- a/backup/src/main/scala/Boot.scala
+++ b/backup/src/main/scala/Boot.scala
@@ -1,0 +1,32 @@
+import java.io.File
+
+import batches.{Backup, Batch}
+import play.api.inject.guice.GuiceApplicationLoader
+import play.api.{Logger, ApplicationLoader, Mode, Environment}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import scala.reflect.ClassTag
+import scala.util.{Failure, Success}
+
+object Boot {
+  def runBatch[T<:Batch](rootPath: String)(implicit classTag: ClassTag[T]): Unit = {
+    val env = Environment(new File(rootPath), this.getClass.getClassLoader, Mode.Prod)
+    val ctx = ApplicationLoader.createContext(env)
+    val app = new GuiceApplicationLoader().load(ctx)
+    val className = classTag.runtimeClass.getName
+
+    val backup = app.injector.instanceOf[T]
+
+    Logger.info(s"Starting $className")
+    backup.execute().onComplete {
+      case Success(_) =>
+        Logger.info(s"End of $className")
+        app.stop()
+      case Failure(error) =>
+        Logger.error(error.getMessage, error)
+        app.stop()
+    }
+  }
+
+  def main (args: Array[String]): Unit = runBatch[Backup]("./backup")
+}
--- a/backup/src/main/scala/Boot.scala
+++ b/backup/src/main/scala/Boot.scala
@@ -15,15 +15,21 @@ object Boot {
     val app = new GuiceApplicationLoader().load(ctx)
     val className = classTag.runtimeClass.getName
 
-    val backup = app.injector.instanceOf[T]
+    try {
+      val backup = app.injector.instanceOf[T]
 
-    Logger.info(s"Starting $className")
-    backup.execute().onComplete {
-      case Success(_) =>
-        Logger.info(s"End of $className")
-        app.stop()
-      case Failure(error) =>
-        Logger.error(error.getMessage, error)
+      Logger.info(s"Starting $className")
+      backup.execute().onComplete {
+        case Success(_) =>
+          Logger.info(s"End of $className")
+          app.stop()
+        case Failure(error) =>
+          Logger.error(error.getMessage, error)
+          app.stop()
+      }
+    } catch {
+      case e: Exception =>
+        Logger.error(e.getMessage, e)
         app.stop()
     }
   }
--- a/backup/src/main/scala/Boot.scala
+++ b/backup/src/main/scala/Boot.scala
@@ -9,7 +9,7 @@ import scala.reflect.ClassTag
 import scala.util.{Failure, Success}
 
 object Boot {
-  def runBatch[T<:Batch](rootPath: String)(implicit classTag: ClassTag[T]): Unit = {
+  def runBatch[T <: Batch](rootPath: String)(implicit classTag: ClassTag[T]): Unit = {
     val env = Environment(new File(rootPath), this.getClass.getClassLoader, Mode.Prod)
     val ctx = ApplicationLoader.createContext(env)
     val app = new GuiceApplicationLoader().load(ctx)
--- a/backup/src/main/scala/batches/Backup.scala
+++ b/backup/src/main/scala/batches/Backup.scala
@@ -17,6 +17,8 @@ import scalaz.{-\/, \/-}
 
 class Backup @Inject() (conf: BackupConfiguration, ws: WSClient)(implicit ec: ExecutionContext) extends Batch {
 
+  val batchExecutionDeadline = DateTime.now(DateTimeZone.UTC).plusHours(3)
+
   def execute(): Future[Unit] = {
     val storageAccount = CloudStorageAccount.parse(conf.storageConnectionString)
     val blobClient = storageAccount.createCloudBlobClient()
@@ -27,7 +29,7 @@ class Backup @Inject() (conf: BackupConfiguration, ws: WSClient)(implicit ec: Ex
 
     val sharedAccessPolicy = new SharedAccessBlobPolicy()
     sharedAccessPolicy.setPermissions(util.EnumSet.of(READ, WRITE, LIST))
-    sharedAccessPolicy.setSharedAccessExpiryTime(DateTime.now(DateTimeZone.UTC).plusHours(3).toDate)
+    sharedAccessPolicy.setSharedAccessExpiryTime(batchExecutionDeadline.toDate)
     val arg = container.generateSharedAccessSignature(sharedAccessPolicy, "")
     val containerUri = container.getUri + "?" + arg
 
--- a/backup/src/main/scala/batches/Backup.scala
+++ b/backup/src/main/scala/batches/Backup.scala
@@ -36,7 +36,6 @@ class Backup @Inject() (conf: BackupConfiguration, ws: WSClient)(implicit ec: Ex
       outputContainerUri = Some(containerUri)
     )
 
-    Logger.info(notificationJob.toXml.toString())
     val hubClient = new NotificationHubClient(conf.notificationHubConnection, ws)
     hubClient.submitNotificationHubJob(notificationJob).map {
       case \/-(result) => result.items.foreach { job =>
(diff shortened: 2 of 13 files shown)


backup/src/main/scala/Boot.scala
```diff
@@ -0,0 +1,38 @@
...
+import batches.{Backup, Batch}
+import play.api.inject.guice.GuiceApplicationLoader
+import play.api.{Logger, ApplicationLoader, Mode, Environment}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+import scala.reflect.ClassTag
+import scala.util.{Failure, Success}
+
+object Boot {
+  def runBatch[T<:Batch](rootPath: String)(implicit classTag: ClassTag[T]): Unit = {
```
agulowaty: I'd put spaces in `[T <: Batch]`

backup/src/main/scala/batches/BackupConfiguration.scala
```diff
@@ -0,0 +1,40 @@
...
+trait BackupConfiguration {
+  def storageConnectionString: String
+  def containerName: String
+  def directoryName: String
+  def backupRetention: Int
+  def notificationHubConnection: NotificationHubConnection
+}
+
+@Singleton
+class BackupConfigurationFile extends BackupConfiguration {
```
agulowaty: Better name would be `FileBackupConfiguration`

backup/src/main/scala/batches/Backup.scala
```diff
@@ -0,0 +1,70 @@
...
+import gu.msnotifications.{NotificationHubJobType, _}
+import org.joda.time.{DateTime, DateTimeZone}
+import play.api.libs.ws.WSClient
+import play.api.Logger
+import collection.JavaConversions._
+
+import scala.concurrent.{ExecutionContext, Future}
+import scalaz.{-\/, \/-}
+
+class Backup @Inject() (conf: BackupConfiguration, ws: WSClient)(implicit ec: ExecutionContext) extends Batch {
```
agulowaty: Any tests ?

backup/src/main/scala/batches/BackupConfiguration.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import javax.inject.Singleton
+
+import com.google.inject.ImplementedBy
+import com.gu.conf.ConfigurationFactory
+import gu.msnotifications.NotificationHubConnection
+
+
+@ImplementedBy(classOf[BackupConfigurationFile])
+trait BackupConfiguration {
```
davidfurey: What is the benefit of this over just having a BackupConfiguration class?

common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -5,18 +5,20 @@ import gu.msnotifications.NotificationHubClient.HubResult
...
 
 trait XmlReads[T] {
-  def reads(xml: scala.xml.Elem): HubResult[T]
+  def reads(xml: Elem): HubResult[T]
 }
 
 object XmlParser {
-  private def getXml(response: WSResponse): HubResult[scala.xml.Elem] = {
+  private def getXml(response: WSResponse): HubResult[Elem] = {
+    Logger.info(response.xml.toString)
```
davidfurey: I presume this was for debugging and can be removed?

common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -37,18 +39,47 @@ object XmlParser {
...
 
-  implicit class RichXmlElem(xml: scala.xml.Elem) {
+  implicit class RichXmlElem(xml: Elem) {
 
     def getStrings(s: String): Seq[String] = (xml \ s).map(_.text)
 
     def getString(s: String): HubResult[String] =
       getStrings(s).headOption \/> HubParseFailed(body = xml.toString(), reason = s"Missing field $s")
 
+    def getOptionString(s: String): HubResult[Option[String]] =
```
agulowaty: `get` naming convention in this class reminds me POJOs ;) I'm not a fan of that + I prefer to name all *option* accessors like 
```
def stringOption(path: String)
```

backup/src/main/scala/batches/Backup.scala
```diff
@@ -0,0 +1,70 @@
+package batches
```
agulowaty: I'll use singular noun here i.e. `batch`

alexduf: We have ```controllers``` and ```services```, I figured I would stay consistent

backup/src/main/scala/batches/Backup.scala
alexduf: All the Microsoft classes are final, in my opinopn the overhead of creating classes that delegates to Microsoft's code to allow testing doesn't justify creating a unit test here.
I'm opening it for debate though, if anyone has a very string opinion against it.

backup/src/main/scala/batches/BackupConfiguration.scala
alexduf: None, it's a remain of me trying to write unit tests (see previous comment). I'll make it a class next commit.

common/src/main/scala/gu/msnotifications/Responses.scala
alexduf: Indeed!

alexduf:
can @davidfurey or @agulowaty have another look? I pushed a new commit

backup/src/main/scala/batches/Backup.scala
agulowaty: And especially in that case I'd distinguish packages that were imposed by framework (Play) and our own created to encapsulate business logic so my vote goes for singular name for batch.

backup/src/main/scala/batches/Backup.scala
agulowaty: For me this ticket is no-go without tests. If we cannot unit-test it, we should make an effort to write automated tests with mocked Windows Notification Hub WS interface.

Odhran: Wouldn't these type of tests be overkill here? I'd prefer pragmatic decisions about when to test.

agulowaty: So what's pragmatic decision here ? Not having automated tests for that sub-module at all ?

backup/src/main/scala/batches/Backup.scala
```diff
@@ -0,0 +1,70 @@
...
+    val storageAccount = CloudStorageAccount.parse(conf.storageConnectionString)
+    val blobClient = storageAccount.createCloudBlobClient()
+    val container = blobClient.getContainerReference(conf.containerName)
+    container.createIfNotExists()
+
+    cleanup(container)
+
+    val sharedAccessPolicy = new SharedAccessBlobPolicy()
+    sharedAccessPolicy.setPermissions(util.EnumSet.of(READ, WRITE, LIST))
+    sharedAccessPolicy.setSharedAccessExpiryTime(DateTime.now(DateTimeZone.UTC).plusHours(3).toDate)
```
davidfurey: Why 3 hours?

common/src/main/scala/gu/msnotifications/NotificationHubJob.scala
```diff
@@ -0,0 +1,83 @@
...
+  createdAt: Option[DateTime],
+  updatedAt: Option[DateTime]
+)
+
+object NotificationHubJob {
+
+  private def parseProperties(seq: NodeSeq): HubResult[Map[String, String]] = {
+    \/-(seq.map { elem =>
+      (elem \ "Key" text) -> (elem \ "Value" text)
+    }.toMap)
```
davidfurey: I think using .right is more readable than \/-

notification/conf/bootstrap.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+/var/lib/dpkg/info/ca-certificates-java.postinst configure
```
davidfurey: Should we do this as part of baking our AMIs?

common/src/main/scala/gu/msnotifications/NotificationHubJob.scala
agulowaty: `scalaz.syntax.either._`

backup/src/main/scala/batches/Backup.scala
davidfurey: This is a effectively a bash script written in Scala.  What monitoring can we put in place so that we are confident that the backup is working on an ongoing basis?  I am more worried about that than about a code change breaking it.

backup/src/main/scala/batches/Backup.scala
alexduf: Because I had to draw a line somewhere. I'm happy with 1hour or 1 day.

davidfurey: So is it effectively the maximum length of time you expect the backup to take?

alexduf: Yes, more precisely, it's how long I'm allowing the hub job to write on the storage.

notification/conf/bootstrap.sh
alexduf: You're right, I did that here: https://github.com/guardian/mobile-platform/pull/197. I'll remove that line

alexduf:
The branch is rebased which explains the odd history. I've added the monitoring to send the logs via cloudwatch and an email at the end of the batch. Can anyone have another look?

common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -6,17 +6,17 @@ import models.WindowsMobile
 import notifications.providers.{RegistrationResponse => RegistrarResponse, UserIdNotInTags}
 import org.joda.time.DateTime
 import play.api.libs.ws.WSResponse
-
-import scala.util.Try
-import scalaz.\/
+import scala.util.{Failure, Success, Try}
+import scala.xml.Elem
+import scalaz.{-\/, \/-, \/}
```
agulowaty: it was mentioned in previous PRs by @davidfurey - `import scalaz.syntax.either._` gives opportunity to convert:
`\/-(Some(dateTime))` to `Some(dateTime).right` - IMHO its more readable.

backup/src/main/scala/batches/Backup.scala
```diff
@@ -0,0 +1,70 @@
...
+    val arg = container.generateSharedAccessSignature(sharedAccessPolicy, "")
+    val containerUri = container.getUri + "?" + arg
+
+    val notificationJob = NotificationHubJobRequest(
+      jobType = NotificationHubJobType.ExportRegistrations,
+      outputContainerUri = Some(containerUri)
+    )
+
+    Logger.info(notificationJob.toXml.toString())
```
agulowaty: this had to be removed

backup/src/main/scala/batches/Backup.scala
```diff
@@ -0,0 +1,70 @@
...
+    val hubClient = new NotificationHubClient(conf.notificationHubConnection, ws)
+    hubClient.submitNotificationHubJob(notificationJob).map {
```
agulowaty: it's like fire-and-forget usage, how we gonna observer if job has been completed ?

backup/src/main/scala/batches/Backup.scala
```diff
@@ -0,0 +1,70 @@
...
+      }
+    }
+
+    val tooOld = DateTime.now().minusDays(conf.backupRetention).toDate
+    val directory = container.getDirectoryReference(conf.directoryName)
+    val allBlobs = listAllFiles(directory)
+    val blobsToDelete = allBlobs.filter(_.getProperties.getLastModified.before(tooOld))
+    blobsToDelete.foreach { blob =>
+      blob.deleteIfExists()
+      Logger.info(s"deleted old backup: ${blob.getName}")
```
agulowaty: `com.microsoft.azure.storage.blob.BlobProperties#getLength` offers size of CloudBlob, I think it would be useful to observe how it varies (at least each backup should have size >= the previous)

backup/src/main/scala/batches/Backup.scala
alexduf: It is fire-and-forget

backup/src/main/scala/batches/Backup.scala
alexduf: I have to draw a line somewhere when it comes to how many features we want the batch to have. I won't include that, even though that's a good suggestion.

davidfurey:
:+1:

PR #21, opened 2015-10-08
MAPI-989: send topics as tags to hub

agulowaty:
Topics in Registration to be sent in RawWindowsRegistration to Azure hub.

### DIFF
--- a/common/src/main/scala/gu/msnotifications/RawWindowsRegistration.scala
+++ b/common/src/main/scala/gu/msnotifications/RawWindowsRegistration.scala
@@ -8,7 +8,10 @@ object RawWindowsRegistration {
   def fromMobileRegistration(m: Registration) = {
     RawWindowsRegistration(
       channelUri = m.deviceId,
-      tags = Tags.withUserId(m.userId).asSet
+      tags = Tags()
+        .addUserId(m.userId)
+        .addTopics(m.topics)
+        .asSet
     )
   }
 }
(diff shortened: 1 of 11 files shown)


common/src/main/scala/gu/msnotifications/RawWindowsRegistration.scala
```diff
@@ -8,7 +8,10 @@ object RawWindowsRegistration {
   def fromMobileRegistration(m: Registration) = {
     RawWindowsRegistration(
       channelUri = m.deviceId,
-      tags = Tags.withUserId(m.userId).asSet
+      tags = Tags()
+        .addUserId(m.userId)
```
davidfurey: This would be clearer if the methods were withUserId and withTopics

common/src/main/scala/gu/msnotifications/package.scala
```diff
@@ -33,12 +33,19 @@ package object msnotifications {
...
     def asSet = tags
-    def findUserId: Option[UserId] = tags.find(_.startsWith(UserTagPrefix)).map(UserId(_))
+
+    def findUserId: Option[UserId] = tags
+      .find(_.startsWith(UserTagPrefix))
+      .map { s => UserId(s.split(':')(1)) }
+
+    def addUserId(userId: UserId) = copy(tags + s"${UserTagPrefix}${userId.userId}")
+
+    def addTopics(topics: Set[Topic]) = copy(tags ++ topics.map(_.name))
```
davidfurey: This is throwing away the Topic's 'type'.
Also, it would be good if a topic with a name starting with "user:" didn't break the system.

common/src/main/scala/gu/msnotifications/package.scala
```diff
@@ -11,34 +12,40 @@ package object msnotifications {
...
-    def fromUserId(u: UserId): WNSTopic = {
-      WNSTopic(s"user:${u.userId}")
+    def fromUserId(u: UserId): WNSTag = {
+      WNSTag(s"user:${u.userId}")
     }
+
   }
 
-  case class Tags(tags: Set[String] = Set.empty) {
+  case class Tags(tags: Set[WNSTag] = Set.empty) {
```
davidfurey: Should this be called WNSTags?

davidfurey:
:+1:

common/src/main/scala/gu/msnotifications/package.scala
agulowaty: `Tag` will be sufficient probably, I'll simplify that in next PR

PR #22, opened 2015-10-12
MAPI-980 - invalid topics filtering

agulowaty:
Topics filtered against auditors before sending registration request to provider.

Still to come in next PRs:
- rename WNSTag to Tag
- return list of topics in response
- update README

### DIFF
--- a/common/src/main/scala/auditor/package.scala
+++ b/common/src/main/scala/auditor/package.scala
@@ -1,0 +1,29 @@
+import java.net.URL
+
+import models.Topic
+import play.api.libs.json.Json
+
+import scala.concurrent.{ExecutionContext, Future}
+
+package object auditor {
+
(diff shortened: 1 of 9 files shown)


common/src/main/scala/auditor/package.scala
```diff
@@ -0,0 +1,29 @@
...
+
+package object auditor {
+
+  case class AuditorGroupConfig(hosts: Set[String])
+
+  case class Auditor(host: URL)
+
+  case class AuditorGroup(auditors: Set[Auditor]) {
+    def queryEach[T](query: Auditor => Future[T])(implicit ec: ExecutionContext): Future[Set[T]] =
+      Future.sequence(auditors.map(query(_)))
```
davidfurey: You shouldn't need the (_) here

davidfurey:
:+1:

PR #23, opened 2015-10-13
Scalastyle

davidfurey:


notification/app/controllers/Main.scala
```diff
@@ -29,7 +29,7 @@ final class Main @Inject()(
     case error: providers.Error => InternalServerError(error.reason)
   }
 
-  def healthCheck = Action {
+  def healthCheck: Action[AnyContent] = Action {
```
agulowaty: In such cases I got mixed feelings about using return type declarations. It makes no sense here at all and in similar cases we shouldn't blindly follow scalastyle warnings.

agulowaty:
:+1:

alexduf:
made redundant by #38

PR #24, opened 2015-10-13
MAPI-1009 - return topics in registration response

agulowaty:
- Rename WNSTag to Tag
- Fixed topic validation - Auditor query returns expired topics instead validated list.
- Implemented RegistrationResponse returning topics in REST api,
- README doc registration section updated.


common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -65,12 +65,17 @@ object RegistrationResponse {
   }
 }
 case class RegistrationResponse(registration: WNSRegistrationId, tags: List[String], channelUri: String, expirationTime: DateTime) {
-  def toRegistrarResponse = for {
-    userId <- Tags.fromUris(tags.toSet).findUserId \/> UserIdNotInTags()
-  } yield RegistrarResponse(
+  def toRegistrarResponse: \/[UserIdNotInTags, RegistrarResponse] = {
```
davidfurey: This can be written more clearer as UserIdNotInTags \/ RegistrarResponse

davidfurey:
:+1: apart from the one comment

PR #25, opened 2015-10-14
Fix test suite with in-memory dynamo db

agulowaty:
- localdynamodb sbt config attached only to common project and removed testWithDb task.


davidfurey:
Great work!
:+1:

PR #28, opened 2015-11-16
Mapi 1084 uuid

alexduf:
This PR focuses on using UUIDs instead of String wherever possible. I believe it will bring type safety and consistency for our future changes.


agulowaty:
:+1:

PR #29, opened 2015-11-16
Mapi 1084 integration

alexduf:
This PR contains the changes that were necessary to get the integration test working. See the list of commits as they are quite varied.

### DIFF
--- a/common/src/main/scala/gu/msnotifications/package.scala
+++ b/common/src/main/scala/gu/msnotifications/package.scala
@@ -4,11 +4,11 @@ import models.{Topic, UserId}
 
 package object msnotifications {
 
-  private def encodeBase16(string: String): String =
-    string.getBytes("UTF-8").map("%02X".format(_)).mkString.toLowerCase
+  private def encodeBase16(source: String): String =
+    source.getBytes("UTF-8").map("%02X".format(_)).mkString.toLowerCase
 
-  private def decodeBase16(string: String): String = {
-    val bytes = string.grouped(2).map(unit => java.lang.Byte.parseByte(unit, 16))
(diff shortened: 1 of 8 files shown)


agulowaty:
:+1:

common/src/main/scala/gu/msnotifications/package.scala
```diff
@@ -1,24 +1,24 @@
...
 
-  private def encode(string: String): String =
-    Base64.getUrlEncoder.encodeToString(string.getBytes("UTF-8"))
+  private def encodeBase16(string: String): String =
+    string.getBytes("UTF-8").map("%02X".format(_)).mkString.toLowerCase
 
-  private def decode(string: String): String =
-    new String(Base64.getUrlDecoder.decode(string), "UTF-8")
+  private def decodeBase16(string: String): String = {
+    val bytes = string.grouped(2).map(unit => java.lang.Byte.parseByte(unit, 16))
```
davidfurey: Is there another name that you could use instead of unit to avoid confusion with Unit

common/src/main/scala/gu/msnotifications/package.scala
```diff
@@ -1,24 +1,24 @@
...
-import java.util.Base64
-
-import models.{TopicType, Topic, UserId}
+import models.{Topic, UserId}
 
 package object msnotifications {
 
-  private def encode(string: String): String =
-    Base64.getUrlEncoder.encodeToString(string.getBytes("UTF-8"))
+  private def encodeBase16(string: String): String =
```
davidfurey: Is there a better variable that could be used instead of "string"?  To me, even "s" would be an improvement

davidfurey:
:+1:

PR #30, opened 2015-11-16
Mapi 1084 atom

alexduf:
This PR focuses on the XML response parsing.

### DIFF
--- a/backup/src/main/scala/batches/Backup.scala
+++ b/backup/src/main/scala/batches/Backup.scala
@@ -40,9 +40,7 @@ class Backup @Inject() (conf: Configuration, ws: WSClient)(implicit ec: Executio
 
     val hubClient = new NotificationHubClient(conf.notificationHubConnection, ws)
     hubClient.submitNotificationHubJob(notificationJob).map {
-      case \/-(result) => result.items.foreach { job =>
-        Logger.info(s"Success, job created with id ${job.jobId}\n$job")
-      }
+      case \/-(job) => Logger.info(s"Success, job created with id ${job.jobId}\n$job")
       case -\/(failure) => Logger.error(failure.toString)
     }
   }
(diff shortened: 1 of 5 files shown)


backup/src/main/scala/batches/Backup.scala
```diff
@@ -40,9 +40,7 @@ class Backup @Inject() (conf: Configuration, ws: WSClient)(implicit ec: Executio
 
     val hubClient = new NotificationHubClient(conf.notificationHubConnection, ws)
     hubClient.submitNotificationHubJob(notificationJob).map {
-      case \/-(result) => result.items.foreach { job =>
-        Logger.info(s"Success, job created with id ${job.jobId}\n$job")
-      }
+      case \/-(job) => Logger.info(s"Success, job created with id ${job.jobId}\n$job")
```
davidfurey: Do you really want a newline character in a log message?

common/src/main/scala/gu/msnotifications/Responses.scala
```diff
@@ -108,6 +108,26 @@ case class RegistrationResponse(registration: WNSRegistrationId, tags: List[Stri
...
+
+  implicit def reader[T](implicit reader: XmlReads[T]): XmlReads[AtomEntry[T]] = new XmlReads[AtomEntry[T]] {
+    def reads(xml: Elem) =  for {
+      title <- xml.textNode("title")
+      item <- getItem(xml)(reader)
+    } yield AtomEntry(title, item)
+  }
+
+  private def getItem[T](xml: Elem)(implicit reader: XmlReads[T]): HubResult[T] = {
+    val results = (xml \ "content").flatMap(_.child).collect {
```
davidfurey: Could you use collectFirst instead?

alexduf: Never used collectFirst before, sounds like a good idea!

davidfurey:
:+1:

PR #31, opened 2015-11-16
Let's rename packages

alexduf:
We have a couple of name collisions, so let's introduce (very) simple packages.
Also I renamed gu.msnotification to azure as the package name wasn't making sense to me


registration/test/registration/controllers/MainControllerSpec.scala
```diff
@@ -1,9 +1,9 @@
-package controllers
+package rregistration.controllersegistration.controllers
```
davidfurey: typo?

alexduf: I have no idea what happen here :D I'll fix it.

davidfurey:
:+1:

PR #32, opened 2015-11-24
extract the logic that isn't shared and move it to their respective owners

alexduf:
If we don't do that change, MAPI-1103 will include all the logic in the common module which I would like to avoid.

I'm only moving things in this PR, no new code has been written


notification/app/notification/services/WindowsNotificationSender.scala
```diff
@@ -0,0 +1,36 @@
...
+    count <- getCounts(push.destination)
+  } yield {
+      result map { _ =>
+        NotificationReport.create(
+          sentTime = DateTime.now,
+          notification = push.notification,
+          statistics = NotificationStatistics(Map(WindowsMobile -> count.toOption))
+        )
+      }
+    }
```
davidfurey: indentation is a bit out here

davidfurey:
Seems fine to me

PR #33, opened 2015-11-25
Mapi 1103 payloads

alexduf:


notification/app/notification/models/azure/Notification.scala
```diff
@@ -0,0 +1,98 @@
...
+  def thumbnailUrl: Option[URL]
+  def topic: Set[Topic]
+}
+
+object Notification {
+
+  def toAzureRawPush(push: Push) = {
+    val body = Json.stringify(Json.toJson(push.notification))
+    push.destination match {
+      case Left(topic: Topic) => AzureRawPush("wns/raw", body, Some(Set(Tag.fromTopic(topic))))
```
agulowaty: "wns/raw" is default value in constructor
BTW: why wnsType is exposed as ctor parameter ?

common/src/main/scala/models/JsonUtils.scala
```diff
@@ -34,4 +38,12 @@ object JsonUtils {
     override def writes(o: Option[T]): JsValue =
       o map format.writes getOrElse JsNull
   }
+
+  implicit val urlFormat = new Format[URL] {
+    override def writes(o: URL): JsValue = JsString(o.toExternalForm)
+    override def reads(json: JsValue): JsResult[URL] = json match {
+      case JsString(url) => Try(JsSuccess(new URL(url))).getOrElse(JsError(s"Invalid url: $url"))
```
agulowaty: Do we really need java.net.URL semantics here ? It establishes dangerous stream handler that will resolve actual Internet host  on .equals for instance.

notification/app/notification/models/azure/Notification.scala
alexduf: Good point, it doesn't need to be here.

common/src/main/scala/models/Notification.scala
```diff
@@ -1,15 +1,102 @@
...
+
+object BreakingNewsNotification {
+  implicit val jf = Json.format[BreakingNewsNotification]
 }
 
-case class Notification(
-  uuid: UUID,
+case class ContentNotification(
+  id: UUID,
+  `type`: String = "content",
```
agulowaty: how about actually introducing some type for that instead String ?

notification/app/notification/services/AzureRawPushConverter.scala
```diff
@@ -0,0 +1,76 @@
...
+      tags = tags(push.destination)
+    )
+  }
+
+  private def tags(destination: Either[Topic, UserId]) = destination match {
+    case Left(topic: Topic) => Some(Set(Tag.fromTopic(topic)))
+    case Right(user: UserId) => Some(Set(Tag.fromUserId(user)))
+  }
+
+  private def toUrl(link: Link): URL = link match {
```
agulowaty: to avoid pattern matching here it could be defined on trait and case classes of Link

agulowaty:
:+1:

common/src/main/scala/models/JsonUtils.scala
alexduf: I wasn't aware of that, that is a serious problem indeed, will fix. Thanks!

notification/app/notification/services/AzureRawPushConverter.scala
alexduf: I thought about it, but there's not easy way of using the configuration object.

alexduf:
I merged it to unblock other tickets, but that would be nice if you could have another quick look @agulowaty, just in case I missed something else. Thanks :) !

notification/app/notification/services/AzureRawPushConverter.scala
agulowaty: with:
```
Link.toUrl(endpoint: Option[String])
```
```
val link = Internal("capidId")
link.toUrl(endpoint: conf.endpoint)
```

agulowaty:
looks good, :+1:

PR #34, opened 2015-11-27
Add return value when sending a notification

alexduf:
- Add traces
 - return 201
 - test for 201 in the azure client

### DIFF
--- a/common/src/main/scala/azure/NotificationHubClient.scala
+++ b/common/src/main/scala/azure/NotificationHubClient.scala
@@ -53,7 +53,7 @@ class NotificationHubClient(notificationHubConnection: NotificationHubConnection
       .withHeaders(serviceBusTags: _*)
       .post(azureWindowsPush.body)
       .map { response =>
-        if (response.status >= 200 && response.status < 300)
+        if ((200 until 300).contains(response.status))
           \/.right(())
         else
           \/.left(XmlParser.parseError(response))
--- a/common/src/main/scala/azure/NotificationHubClient.scala
+++ b/common/src/main/scala/azure/NotificationHubClient.scala
@@ -53,7 +53,7 @@ class NotificationHubClient(notificationHubConnection: NotificationHubConnection
       .withHeaders(serviceBusTags: _*)
       .post(azureWindowsPush.body)
       .map { response =>
-        if (response.status >= 200 || response.status < 300)
+        if (response.status >= 200 && response.status < 300)
           \/.right(())
         else
           \/.left(XmlParser.parseError(response))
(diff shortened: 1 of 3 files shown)


common/src/main/scala/azure/NotificationHubClient.scala
```diff
@@ -53,7 +53,7 @@ class NotificationHubClient(notificationHubConnection: NotificationHubConnection
       .withHeaders(serviceBusTags: _*)
       .post(azureWindowsPush.body)
       .map { response =>
-        if (response.status == 200)
+        if (response.status >= 200 || response.status < 300)
```
davidfurey: I think you meant if (response.status >= 200 && response.status < 300)

common/src/main/scala/models/PushResult.scala
```diff
@@ -0,0 +1,12 @@
+package models
+
+import java.util.UUID
+
+import play.api.libs.json.Json
+
+case class PushResult(id: UUID)
+
+object PushResult {
+  import JsonUtils._
```
davidfurey: What does this import do?

alexduf: A json formatter for UUID

common/src/main/scala/azure/NotificationHubClient.scala
```diff
@@ -53,7 +53,7 @@ class NotificationHubClient(notificationHubConnection: NotificationHubConnection
       .withHeaders(serviceBusTags: _*)
       .post(azureWindowsPush.body)
       .map { response =>
-        if (response.status == 200)
+        if (response.status >= 200 && response.status < 300)
```
agulowaty: `(200 until 300).contains(response.status)` could be more clear I think

common/src/main/scala/models/PushResult.scala
alexduf: Nope, I'm wrong, it's useless

agulowaty: maybe we should make it more explicit ?

notification/app/notification/controllers/Main.scala
```diff
@@ -39,13 +40,15 @@ final class Main @Inject()(
       case \/-(report) =>
         notificationReportRepository.store(report) map {
           case \/-(_) =>
-            Ok("Ok")
+            Logger.info(s"Notification was sent: $push")
```
agulowaty: this will output whole Push object to info logger ?

alexduf: Yes, I think that could be useful for audit. We only keep 7 days of logs anyway.

alexduf:
Any one to have another quick look ?

agulowaty:
:+1:

PR #38, opened 2015-11-27
Scalastyle rebased

alexduf:
bringing this branch back to life


common/src/main/scala/azure/package.scala
```diff
@@ -5,8 +5,9 @@ package object azure {
   private def encodeBase16(source: String): String =
     source.getBytes("UTF-8").map("%02X".format(_)).mkString.toLowerCase
 
+  val base = 16
```
davidfurey: Could call this hex or something more meaningful?

davidfurey:
:+1:

PR #39, opened 2015-12-01
provide a more meaningful name for the variable

alexduf:


agulowaty:
16 = 16 ;) anyway, :+1:

PR #41, opened 2015-12-01
Scalastyle

alexduf:
Enable scalastyle as a test step, and fix the dynamodb crashing the jvm when running testQuick and testOnly.

Let's agree that if it gets in our way we either modify the xml file or get rid of it completely.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
 import com.gu.riffraff.artifact.RiffRaffArtifact.autoImport._
 
+lazy val testScalastyle = taskKey[Unit]("testScalastyle")
+
 lazy val common = project
   .settings(localdynamodb.settings)
   .settings(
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -17,7 +19,14 @@ lazy val common = project
       "com.amazonaws" % "aws-java-sdk" % "1.9.31",
       "com.gu" %% "configuration" %  "4.1"
     ),
-    test in Test <<= (test in Test).dependsOn(DynamoDBLocal.Keys.startDynamoDBLocal)
+    test in Test <<= (test in Test).dependsOn(DynamoDBLocal.Keys.startDynamoDBLocal),
+    testOnly in Test <<= (testOnly in Test).dependsOn(DynamoDBLocal.Keys.startDynamoDBLocal),
+    testQuick in Test <<= (testQuick in Test).dependsOn(DynamoDBLocal.Keys.startDynamoDBLocal),
+    scalastyleFailOnError := true,
+    testScalastyle := org.scalastyle.sbt.ScalastylePlugin.scalastyle.in(Compile).toTask("").value,
```
agulowaty: 1. shouldn't be `ScalastylePlugin.scalastyle.in(Test)` instead ?
2. how about importing org.scalastyle.sbt.ScalastylePlugin ?

alexduf: 1. No, we're checking the style of the production code
2. You're right I'll make it more pretty

agulowaty:
:+1:

davidfurey:
:+1:

PR #44, opened 2015-12-08
Mapi 1140

alexduf:
- Move classes where they are more appropriate (````Tag````, ````Tags````, ````Push```` and ````PushResult````)
 - Create a new type ````Destination````, that is easier to manage than ````Either[Set[Topic], UserId]````
 - Modify the endpoint to ignore the potential topic passed as a parameter
 - Add a unit test for a collection of topics
 - Cleanup the test logs as the awsclient is way too verbose


notification/app/notification/models/Destination.scala
```diff
@@ -0,0 +1,11 @@
+package notification.models
+
+import models.{UserId, Topic}
+
+object Destination {
+  type Destination = Either[Set[Topic], UserId]
+
+  def apply(userId: UserId): Destination = Right(userId)
+  def apply(topic: Topic): Destination = Left(Set(topic))
+  def apply(topics: Set[Topic]): Destination = Left(topics)
```
agulowaty: could we push to empty list of topics ?

alexduf: That's a good point, I'll add a check in the controller

notification/conf/routes
```diff
@@ -1,3 +1,4 @@
 GET         /healthcheck                        @notification.controllers.Main.healthCheck
 POST        /push/user/:userId                  @notification.controllers.Main.pushUser(userId: java.util.UUID)
+POST        /push/topic                         @notification.controllers.Main.pushTopics
```
agulowaty: what should be in the body ?

notification/app/notification/models/Push.scala
```diff
@@ -0,0 +1,9 @@
+package notification.models
+
+import models.Notification
+import notification.models.Destination.Destination
+import play.api.libs.json._
+
+case class Push(notification: Notification, destination: Destination) {
+  def tagQuery: Option[String] = None
```
agulowaty: why None ?

notification/app/notification/controllers/Main.scala
```diff
@@ -55,8 +56,11 @@ final class Main @Inject()(
     }
   }
 
-  def pushTopic(topic: Topic): Action[Notification] = AuthenticatedAction.async(BodyJson[Notification]) { request =>
-    val push = Push(request.body, Left(topic))
+  @deprecated("A push notification can be sent to multiple topics, this is for backward compatibility only", since = "07/12/2015")
+  def pushTopic(topic: Topic): Action[Notification] = pushTopics
+
+  def pushTopics: Action[Notification] = AuthenticatedAction.async(BodyJson[Notification]) { request =>
+    val push = Push(request.body, Left(request.body.topic))
```
agulowaty: Field `.topic` should be pluralised in: https://github.com/guardian/mobile-n10n/blob/f2b89e81b5a1b32c08d4da604641410f0ef8f1bd/common/src/main/scala/models/Notification.scala#L15

notification/app/notification/services/WindowsNotificationSender.scala
```diff
@@ -30,9 +33,21 @@ class WindowsNotificationSender(hubClient: NotificationHubClient, configuration:
...
 
-
-  private def getCounts(destination: Either[Topic, UserId]): Future[RepositoryResult[Int]] = destination match {
-    case Left(topic: Topic) => topicSubscriptionsRepository.count(topic)
+  private def count(destination: Destination): Future[RepositoryResult[Int]] = destination match {
+    case Left(topics: Set[Topic]) => sumOf(topics)
     case Right(_: UserId) => Future.successful(RepositoryResult(1))
   }
+
+  private def sumOf(topics: Set[Topic]): Future[RepositoryResult[Int]] = {
```
agulowaty: `sumOfSuccessfull` ?

notification/test/notification/services/AzureRawPushConverterSpec.scala
```diff
@@ -31,17 +31,22 @@ class AzureRawPushConverterSpec extends Specification with Mockito {
...
+    "Convert a list of topics into a list of tag" in new PushConverterScope {
+      val topics = Set(Topic(Breaking, "uk"), Topic(Breaking, "us"))
+      val expected = Tags(Set(Tag("topic:Base16:627265616b696e672f756b"), Tag("topic:Base16:627265616b696e672f7573")))
+      azureRawPushConverter.toTags(Destination(topics)) shouldEqual Some(expected)
     }
   }
 
   trait PushConverterScope extends Scope {
-    def configuration = {
+    def configuration: Configuration = {
```
agulowaty: could be val

notification/app/notification/controllers/Main.scala
alexduf: I know but the client is already released.

notification/app/notification/services/WindowsNotificationSender.scala
alexduf: It's either the sum of everything or it returns an error, so sumOf seems relevant.

notification/app/notification/models/Destination.scala
alexduf: the check has been added

agulowaty: if at least one topic is required to define Destination, maybe vararg like
```
def apply(topic: Topic, topics: Topic*)
```
will suite better ?

agulowaty:
:+1:

PR #47, opened 2015-12-16
Using URIs

alexduf:
To be consistent across all products, let's use URI. Malformed URI will be caught during the parsing of the message.


common/src/main/scala/models/JsonUtils.scala
```diff
@@ -34,4 +38,13 @@ object JsonUtils {
     override def writes(o: Option[T]): JsValue =
       o map format.writes getOrElse JsNull
   }
+
+  implicit val uriFormat = new Format[URI] {
+    override def reads(json: JsValue): JsResult[URI] = json match {
+      case JsString(uri) => Try(new URI(uri)).map(u => JsSuccess(u)).getOrElse(JsError(s"Invalid URI: $uri"))
+      case _ => JsError("Unknown URI type")
```
davidfurey: I think this error would be more helpful if it said something like "URI must be a string"

agulowaty:
:+1:

PR #49, opened 2016-01-04
disable notifications

alexduf:
Whilst we fix the problem, let's stop the bleeding.


davidfurey:
:+1: once the tests pass

PR #55, opened 2016-02-19
[MAPI-1291] call frontend on notification

agulowaty:


agulowaty:
Tested with uk news alert:
https://internal-frontend.code.dev-guardianapis.com/news-alert/alerts

notification/app/notification/services/frontend/FrontendAlerts.scala
```diff
@@ -0,0 +1,46 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scalaz.\/
+import scalaz.syntax.either._
+
+class FrontendAlerts(config: FrontendAlertsConfig, wsClient: WSClient) extends TrackingObserver {
+  val logger = Logger(classOf[FrontendAlerts])
+
+  def sendBreakingNewsAlert(alert: NewsAlert): Future[String \/ Unit] = wsClient
+    .url(config.endpoint + "/alert")
```
davidfurey: Why not s"${config.endpoint}/alert"

notification/app/notification/services/frontend/package.scala
```diff
@@ -0,0 +1,49 @@
...
+import models.Link._
+import models.{BreakingNewsNotification}
+import org.joda.time.DateTime
+import play.api.libs.json.Json
+
+import scala.PartialFunction.condOpt
+
+package object frontend {
+
+  case class NewsAlert(
```
davidfurey: What is the advantage of putting this inside a package object?

agulowaty: probably none - I moved it to upper level

notification/app/notification/controllers/Main.scala
```diff
@@ -39,17 +44,23 @@ final class Main @Inject()(
   }
 
   private def pushGeneric(push: Push) = {
+    def notifyAboutPush(report: NotificationReport): Future[\/[List[TrackingError], Unit]] = {
```
davidfurey: I think this method is too long

notification/app/notification/controllers/Main.scala
```diff
@@ -39,17 +44,23 @@ final class Main @Inject()(
...
+      val observers = Seq(notificationReportRepository, frontendAlerts)
+      val noErrors = ().right[List[TrackingError]]
+      Future.fold(observers.map(_.notificationSent(report)))(noErrors) { (aggr, trackerResult) => trackerResult match {
```
davidfurey: I think this line does too much

agulowaty: refactored, please take a look again

notification/app/notification/controllers/Main.scala
```diff
@@ -72,5 +62,28 @@ final class Main @Inject()(
...
+          case -\/(errors) => Created(Json.toJson(PushResult.fromReport(report).withTrackingErrors(errors.map(_.reason))))
+        }
+      case -\/(error) =>
+        logger.error(s"Notification ($push) could not be sent: $error")
+        Future.successful(handleErrors(error))
+    }
+  }
 
+  private def notifyAboutPush(observers: Seq[TrackingObserver], report: NotificationReport) = {
+    def notifyObservers = observers.map { _.notificationSent(report) }
```
davidfurey: Why not a val?

notification/app/notification/controllers/Main.scala
```diff
@@ -72,5 +62,28 @@ final class Main @Inject()(
...
+    val noErrors = ().right[List[TrackingError]]
```
davidfurey: This might be clear as:
```
val NoErrors = ().right[List[TrackingError]]
Future.fold(notifyObservers)(NoErrors) {
  case (NoErrors, -\/(err)) => List(err).left
  case (-\/(errors), -\/(err)) => (err :: errors).left
  case (aggr, _) => aggr
}
```

or even more radically using something like:
```  
private def aggregateFailures(ls: Seq[TrackerResult[Unit]]): Seq[TrackingError] \/ Unit = ls collect {
  case -\/(err) => err
} match {
  case errors if errors.nonEmpty => errors.left
  case Nil => ().right
}
```

agulowaty: I like the first idea, second one is actually expanded fold

agulowaty:
@davidfurey: may you take a look again ?

davidfurey:
Looks fine to me, but can you get a :+1:  from someone else too? @alexduf

notification/app/notification/services/frontend/FrontendAlerts.scala
```diff
@@ -0,0 +1,46 @@
...
+import play.api.libs.json.Json
+import play.api.libs.ws.WSClient
+import play.mvc.Http.Status.CREATED
+import tracking.{TrackingError, TrackerResult, TrackingObserver}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+import scalaz.\/
+import scalaz.syntax.either._
+
+class FrontendAlerts(config: FrontendAlertsConfig, wsClient: WSClient) extends TrackingObserver {
```
alexduf: Why implementing that call as a tracking? To me it should be a first class citizen, and be implemented as a ````NotificationSender````

alexduf: Unless I missed anything?

agulowaty: i wanted to distinguish notifications being sent to devices and notifications (alerts) about "push" + contract defined by `NotificationSender` is different and should be implemented by components (providers) that are communicating with devices

alexduf: I don't think the fact the notification goes to a mobile device is reflected in the contract of NotificationSender, in fact we are notifying front-end and I still think we should treat it as something we notify more than something we report to

agulowaty: I'll create another PR with that in mind, thanks Alex.

PR #58, opened 2016-03-07
[MAPI-1306] use dynamodb for reporting

agulowaty:
@davidfurey @DiegoVazquezNanini 

Still to come: 
Updating duplicated reports in Dynamo


common/src/main/scala/aws/AsyncDynamo.scala
```diff
@@ -21,6 +23,11 @@ object AsyncDynamo {
   def keyBetween(a: String, b: String): Condition = new Condition()
     .withComparisonOperator(ComparisonOperator.BETWEEN)
     .withAttributeValueList(new AttributeValue(a), new AttributeValue(b))
+
+  def apply(region: Regions, credentialsProvider: AWSCredentialsProvider = new InstanceProfileCredentialsProvider()): AsyncDynamo = {
```
davidfurey: Can you use DefaultAWSCredentialsProviderChain instead of InstanceProfileCredentialsProvider?

DiegoVazquezNanini:
:+1:

PR #59, opened 2016-03-15
MAPI-1329 - Aggregate notification reports

agulowaty:
Let's send one report to database aggregating multiple `NotificationSender` responses.
Since we'll probably have multiple senders for different providers I flattened `NotificationStatistics` - now it represents stats created by single sender.

cc @davidfurey, @alexduf


alexduf:
:+1:

PR #60, opened 2016-03-15
Update sbt-riffraff-artifact and change build.sbt to build all artefacts

DiegoVazquezNanini:
Updates sbt-riffraff-artefacts plugin to the latest version and change build.sbt to build the 4 artefacts on one go.


alexduf:
:+1: I think it looks good!

PR #61, opened 2016-03-18
MAPI-1330 - enabled Dynamo in topic registrations stats

agulowaty:


agulowaty:
@davidfurey @alexduf @DiegoVazquezNanini

common/src/main/scala/tracking/DynamoTopicSubscriptionsRepository.scala
```diff
@@ -13,14 +14,14 @@ import scala.collection.JavaConversions._
 
 class DynamoTopicSubscriptionsRepository(client: AsyncDynamo, tableName: String)
   (implicit ec: ExecutionContext) extends TopicSubscriptionsRepository {
-
+  val logger = Logger(classOf[DynamoTopicSubscriptionsRepository])
   val TopicIdField = "topicId"
   val TopicSubscriberCountField = "topicSubscriberCount"
 
-  override def subscribe(topic: Topic): Future[RepositoryResult[Unit]] =
+  override def deviceSubscribed(topic: Topic): Future[RepositoryResult[Unit]] =
```
DiegoVazquezNanini: Is it necessary to rename these as part of this pull request, or would it be better to make a new PR instead?

agulowaty: should I create separate PR just for renaming ?

DiegoVazquezNanini: Yes, please. 
Can you please explain why you want to do this renaming?

agulowaty: Naming was wrong in that sense `TopicSubscriptionsRepository` is not responsible for subscribing to topics (as methods suggested). I wanted to emphasise fact it is `notified` on topic subscription instead.

DiegoVazquezNanini: Ok, thanks for the clarification.

agulowaty:
any comments ?

DiegoVazquezNanini:
:+1: LGTM

PR #62, opened 2016-03-22
MAPI-1340 - lenient topic validation

agulowaty:
cc @pvighi @alexduf @DiegoVazquezNanini

### DIFF
--- a/registration/app/registration/controllers/Main.scala
+++ b/registration/app/registration/controllers/Main.scala
@@ -30,26 +30,26 @@ final class Main @Inject()(notificationRegistrarSupport: RegistrarSupport, topic
   def register(lastKnownDeviceId: String): Action[Registration] = Action.async(BodyJson[Registration]) { request =>
     val registration = request.body
 
-    def validate(topics: Set[Topic]) = {
+    def validate(topics: Set[Topic]): Future[Set[Topic]] =
       topicValidator
         .removeInvalid(topics)
-        .onSuccess {
-          case \/-(validTopics) =>
-            logger.debug(s"Successfully validated topics in registration (${registration.deviceId}), topics valid: [$validTopics]")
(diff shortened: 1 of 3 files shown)


registration/app/registration/services/topic/TopicValidator.scala
```diff
@@ -17,11 +17,12 @@ import scalaz.syntax.either._
...
   def removeInvalid(topics: Set[Topic]): Future[TopicValidatorError \/ Set[Topic]]
 
-  trait TopicValidatorError {
-    def reason: String
-  
-    def topicsQueried: Set[Topic]
-  }
+}
+
+trait TopicValidatorError {
```
agulowaty: made public to mock in Spec

registration/app/registration/controllers/Main.scala
```diff
@@ -30,17 +30,27 @@ final class Main @Inject()(notificationRegistrarSupport: RegistrarSupport, topic
...
+    }
+
+    def registerWith(registrar: NotificationRegistrar) =
+      registrar
+        .register(lastKnownDeviceId, registration)
+        .map {processResponse}
 
     registrarFor(registration) match {
+      case \/-(registrar) =>
+        validate(registration.topics)
```
davidfurey: Should the return value of this be ignored?

agulowaty: yes, problems are logged inside method and we don't want to chain it with registration flow

pvighi: So we are not removing the invalid topics even if we get a correct response from auditors ? or am i missing something?

agulowaty: good point, fixed

davidfurey:
:+1:

PR #64, opened 2016-04-26
added error logging to hub client

agulowaty:
Our Windows 10 client recently received bunch of errors caused by failed responses from Notification Hubs. This PR should help us diagnose failed hub requests.

Also removed not used method that gets title from atom feed.

cc @davidfurey @alexduf @DiegoVazquezNanini


davidfurey:
👍

PR #65, opened 2016-04-26
setting correct Content-Type headers for registration ops

agulowaty:
Windows hub notification does not accept some of our registration requests (especially with longer tags in it). I doubt this PR helps fixing that, but it's worth to have them set correctly

cc @alexduf @davidfurey


alexduf:
:+1:

PR #66, opened 2016-04-28
MAPI-1381 - hashing Tag instead Base16 encoding

agulowaty:
Azure Hub does not accept Tags longer than 120 chars.

- removed decodedTopics for Tags and Base16 encoding functions and introduced MD5 as encoding
- changing topic subscription repo to have Topic.id as primary key
- topic subscription repository to unsubscribe topics based on id without full Topic knowledge

FIXMEs to be addressed in next PR


agulowaty:
@alexduf @davidfurey @DiegoVazquezNanini

alexduf:
I think it's missing the fact that our users will still have the base16 tag in the db for a little while. Unless I missed something, if we release that tomorrow, only people who have re-registered will receive the notification.
Could we send the notification to both topics (md5 and base16) for one or two weeks before getting rid of the base 16?

agulowaty:
Good point Alex, I'll do so.

agulowaty:
@alexduf - check latest change set

alexduf:
Looking good ! :+1:

PR #67, opened 2016-04-28
MAPI-1381 - FIXME removed for subscription tracking

agulowaty:
- WindowsNotificationRegistrar: moved bits responsible for calculating diffs and communicating with repository to separate class

cc @alexduf @davidfurey


alexduf:
Looks good to me :+1:

PR #68, opened 2016-05-05
Fixed notification types

pvighi:
Notification type was wrong for goal alerts.
Note that the notification is just a randomly generated UUID in notification-football. In the legacy system this id is derived from match info (score, match minute, things like that) so that we cannot send the same notification twice for the same goal. 

Maybe before merging this we could make sure we generate goal alerts in a deterministic way from match info to preserve the safety of the legacy notifications.

I think this is also an issue for content alerts and no one complained so this might not be a big problem

see https://theguardian.atlassian.net/browse/MAPI-1390


agulowaty:
👍

alexduf:
Linking the related PR for mobile-notification-api-client: https://github.com/guardian/mobile-notifications-api-client/pull/37

PR #69, opened 2016-05-06
Improved rejected notification reason message

agulowaty:
* added reason message to frontend alerts notification sender.

cc @pvighi


alexduf:
:+1:

PR #70, opened 2016-05-10
MAPI-1394 - topic encoded in Base16 are sent as separate hub request

agulowaty:
* the same notification is sent twice - one with differently encoded topic
* protection against duplicated messages addressed to user added

cc @alexduf @davidfurey


alexduf:
Looks good :+1:

PR #71, opened 2016-05-31
MAPI-1421: upgrading to Play 2.5

agulowaty:
* Removed charset asserts in ReportIntegrationSpec since Play 2.5 no longer sends utf-8 as default. 

cc @alexduf @davidfurey


alexduf:
That was rather painless! :+1:

agulowaty:
@alexduf: I've added scripts for application.conf s3 copy - please take a look again (tested on CODE)

alexduf:
Still looks good!

agulowaty:
I fixed logging in backup project - logback implementation was missing on classpath and application needs to accept external config file as parameter

PR #79, opened 2016-07-28
iOS live app via azure

davidfurey:
Support iOS live app notifications using azure notifications hub (daily edition is NOT included in this PR)


common/src/main/scala/models/Registration.scala
```diff
@@ -1,6 +1,6 @@
 package models
 
-case class Registration(deviceId: String, platform: Platform, userId: UserId, topics: Set[Topic])
+case class Registration(deviceId: String, platform: Platform, userId: UniqueDeviceIdentifier, topics: Set[Topic])
```
agulowaty: should we change field name as well ?

davidfurey: It would be nice to, but that would be a breaking change for the windows 10 client

common/src/main/scala/azure/APNSRawPush.scala
```diff
@@ -0,0 +1,9 @@
+package azure
+
+import azure.apns.Body
+
+case class APNSRawPush(body: Body, tags: Option[Tags]) {
+  def tagQuery: Option[String] = tags.map { set =>
```
agulowaty: this is duplicated for each RawPush probably and is specific to azure not APNS or GCM

common/src/main/scala/models/Registration.scala
agulowaty: I know, but its best time now to deprecate old name now at the same time introducing new since we don't have other platforms yet.

davidfurey: Fair point, I'll see how cleanly I can support both names for now.

common/src/main/scala/azure/APNSRawPush.scala
davidfurey: Yes, I'll refactor this

registration/app/registration/services/NotificationRegistrarProvider.scala
```diff
@@ -16,12 +16,16 @@ case class UnsupportedPlatform(platform: String) extends RequestError {
...
 }
 
-final class NotificationRegistrarProvider(windowsNotificationRegistrar: WindowsNotificationRegistrar, gcmNotificationsRegistrar: GCMNotificationRegistrar)
+final class NotificationRegistrarProvider(
+  windowsNotificationRegistrar: WindowsNotificationRegistrar,
+  gcmNotificationRegistrar: GCMNotificationRegistrar,
+  apnsNotificationRegistrar: APNSNotificationRegistrar)
   (implicit executionContext: ExecutionContext) extends RegistrarProvider {
 
   override def registrarFor(registration: Registration): NotificationsError \/ NotificationRegistrar = registration match {
```
alexduf: `registration.platform match {` ?

alexduf: maybe that would look nicer if the pattern match was happening on `registration.platform`?

common/src/main/scala/azure/apns/Body.scala
```diff
@@ -0,0 +1,23 @@
...
+
+import scala.collection.breakOut
+
+case class Body(aps: APS, customProperties: Map[String, String])
+object Body {
+  implicit val jf = new Format[Body] {
+    override def writes(o: Body): JsValue =
+      Json.toJson(o.customProperties.mapValues(JsString(_)) + ("aps" -> Json.toJson(o.aps)))
+
+    override def reads(json: JsValue): JsResult[Body] = for {
```
alexduf: If you're writing your own `reads` function, why passing by that extra step of deserializing through `BodyWithoutCustom`?

I think something in the effect of `(json / "aps").as[APS]` may do

notification/app/notification/models/ios/MessageTypes.scala
```diff
@@ -0,0 +1,6 @@
+package notification.models.ios
+
+object MessageTypes {
+  final val NewsAlert = "m"
+  final val GoalAlert = "g"
```
alexduf: aren't you missing content alerts?

davidfurey: I think content alerts are sent with the NewsAlert type.

pvighi: I think content alerts use "m" as well

alexduf: alright then!

common/src/main/scala/models/Link.scala
```diff
@@ -1,15 +1,24 @@
...
+  case class External(url: String) extends Link {
+    def webUri: URI = new URI(url)
+  }
   object External {
     implicit val jf = Json.format[External]
   }
 
-  case class Internal(contentApiId: String, git: GuardianItemType) extends Link
+  case class Internal(contentApiId: String, shortUrl: Option[String], git: GuardianItemType) extends Link {
+    def webUri: URI = new URI(s"http://www.theguardian.com/$contentApiId") // todo: remove hardcoded domain/protocol
```
alexduf: You probably want to address that todo?

common/src/main/scala/azure/NotificationHubClient.scala
```diff
@@ -70,43 +70,14 @@ class NotificationHubClient(notificationHubConnection: NotificationHubConnection
       }
   }
 
-  def sendWNSNotification(azureWindowsPush: WNSRawPush): Future[HubResult[Unit]] = {
-    val serviceBusTags = azureWindowsPush.tagQuery.map(tagQuery => "ServiceBusNotification-Tags" -> tagQuery).toList
-    logger.debug(s"Sending WNS Raw Notification: $azureWindowsPush")
+  def sendNotification[T](rawPush: RawPush[T]): Future[HubResult[Unit]] = {
```
alexduf: :+1: looks great like that!

common/src/main/scala/models/Registration.scala
```diff
@@ -1,9 +1,23 @@
...
-  implicit val jf = Json.format[Registration]
+  implicit val writes = Json.writes[Registration]
+
+  implicit val reads = new Reads[Registration] {
+    override def reads(json: JsValue): JsResult[Registration] =
+      replaceUserId(json).flatMap(basicReader.reads)
+
+    private val basicReader = Json.reads[Registration]
+
+    private def replaceUserId(json: JsValue) = json.validate[JsObject] map { o =>
```
alexduf: I know you don't like comments, but here a comment would be nice to explain this is for backward compatibility with win10 :smile:

davidfurey: I was trying to convince myself that a deprecation warning would be more appropriate, but a comment is probably best.

pvighi: or you could rename the method "replaceUserIdForBackWardsCompatibilityWithWindowsWhichSendsAUniqueDeviceIdentifierButCallsItUserId"

alexduf:
nice :+1:

PR #80, opened 2016-08-26
Fix ios notifications

davidfurey:
The Guardian iOS app expects the alert to be a string, not an object

### DIFF
--- a/notification/app/notification/models/ios/Notification.scala
+++ b/notification/app/notification/models/ios/Notification.scala
@@ -27,7 +27,9 @@ case class BreakingNewsNotification(
   def payload: Body = Body(
     aps = APS(
       category = Some(category),
-      alert = Some(Right(message))
+      alert = Some(Right(message)),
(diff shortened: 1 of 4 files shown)


notification/app/notification/models/ios/Notification.scala
```diff
@@ -27,9 +27,9 @@ case class BreakingNewsNotification(
   def payload: Body = Body(
     aps = APS(
       category = Some(category),
-      alert = Some(Alert(
-        body = Some(message)
-      ))
+      alert = Some(Right(message)),
+      `content-available` = Some(1),
```
alexduf: I think this is specific to the daily edition

davidfurey: Breaking news has this set too (as seen with debugger attached to device)

alexduf: Alright then!

alexduf:
:+1:

PR #81, opened 2016-09-01
Replace scalaz with cats

davidfurey:
Mainly so that we can start using [XorT](http://eed3si9n.com/herding-cats/stacking-future-and-either.html) instead of FutureEither

### DIFF
--- a/registration/app/registration/services/NotificationRegistrarProvider.scala
+++ b/registration/app/registration/services/NotificationRegistrarProvider.scala
@@ -5,11 +5,11 @@ import models.{Android, iOS, Registration, WindowsMobile}
 import registration.services.azure.{APNSNotificationRegistrar, GCMNotificationRegistrar, WindowsNotificationRegistrar}
 
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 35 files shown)


registration/app/registration/services/NotificationRegistrarProvider.scala
```diff
@@ -5,11 +5,11 @@ import models.{Android, iOS, Registration, WindowsMobile}
...
 
 import scala.concurrent.ExecutionContext
-import scalaz.\/
-import scalaz.syntax.either._
+import cats.data.Xor
+import cats.implicits._
 
 trait RegistrarProvider {
-  def registrarFor(registration: Registration): \/[NotificationsError, NotificationRegistrar]
+  def registrarFor(registration: Registration): Xor[NotificationsError, NotificationRegistrar]
```
alexduf: `NotificationsError Xor NotificationRegistrar` ?

alexduf:
I'm not seeing a massive benefit here but hey I didn't like scalaz in the first place :smiley: 
:+1:

PR #82, opened 2016-09-01
Add unregister endpoint

davidfurey:
This is primarily to allow us to roll backwards/forwards between pushy and n10n.  

When a device registers to pushy, it will be unregistered from n10n.
When a device registers to n10n, it will be unregistered from pushy.

### DIFF
--- a/registration/app/registration/controllers/Main.scala
+++ b/registration/app/registration/controllers/Main.scala
@@ -3,6 +3,8 @@ package registration.controllers
 import java.util.UUID
 
 import azure.HubFailure.{HubInvalidConnectionString, HubParseFailed, HubServiceError}
+import cats.data.{Xor, XorT}
+import cats.implicits._
 import error.{NotificationsError, RequestError}
 import models._
 import play.api.Logger
@@ -10,6 +12,7 @@ import play.api.libs.json.Json
 import play.api.mvc.BodyParsers.parse.{json => BodyJson}
 import play.api.mvc.{Action, AnyContent, Controller, Result}
 import registration.models.LegacyRegistration
+import registration.services.azure.UdidNotFound
 import registration.services.{UnsupportedPlatform, _}
 import registration.services.topic.TopicValidator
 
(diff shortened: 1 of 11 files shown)


registration/app/registration/controllers/Main.scala
```diff
@@ -35,6 +38,21 @@ final class Main(
     registerCommon(lastKnownDeviceId, request.body).map(processResponse)
   }
 
+  def unregister(platform: Platform, udid: UniqueDeviceIdentifier): Action[AnyContent] = Action.async {
+
+    def registrarFor(platform: Platform) = XorT.fromXor[Future](
```
alexduf: so here it's returning a `XorT[NotificationRegistrar]` ?

registration/app/registration/services/NotificationRegistrarProvider.scala
```diff
@@ -1,15 +1,18 @@
...
 import scala.concurrent.ExecutionContext
 import cats.data.Xor
 import cats.implicits._
 
 trait RegistrarProvider {
-  def registrarFor(registration: Registration): Xor[NotificationsError, NotificationRegistrar]
+  def registrarFor(registration: Registration): Xor[NotificationsError, NotificationRegistrar] =
+    registrarFor(registration.platform)
+
+  def registrarFor(platform: Platform): Xor[NotificationsError, NotificationRegistrar]
```
alexduf: Are `Xor[A,B]` and `A Xor B` equivalent? if so should we stick to using only one?

registration/app/registration/controllers/Main.scala
davidfurey: It is returning an `XorT[Future, NotificationsError, NotificationRegistrar]`

registration/app/registration/services/NotificationRegistrarProvider.scala
davidfurey: Yes, they are equivalent and we should stick to just one.  Do you have any preference?

alexduf: not really. I'm used to Xor[A, B] rather than fancy syntax, but that's because I'm old school. I'm fine with the other

davidfurey: [docs](http://typelevel.org/cats/tut/xor.html) also use `Xor[A,B]` so lets go with that

registration/test/registration/controllers/MainControllerSpec.scala
```diff
@@ -37,6 +38,20 @@ class MainControllerSpec extends PlaySpecification with JsonMatchers with Mockit
                                                     /("name" -> "science")
     }
 
+    "return 204 and empty response for unregistration of udid" in new registrations {
+      val Some(result) = route(FakeRequest(DELETE, "/registrations/ios/gia:00000000-0000-0000-0000-000000000000"))
```
alexduf: I'm surprised to see `00000000-0000-0000-0000-000000000000` being a valid UUID!

alexduf:
All of my comments where either questions or out-of-scope work, so go for it :+1:

registration/test/registration/controllers/MainControllerSpec.scala
davidfurey: Nil UUID: https://tools.ietf.org/html/rfc4122#section-4.1.7

PR #83, opened 2016-09-02
Keep iOS uuids uppercase when sending to pushy

davidfurey:
Fixes duplicate notifications bug for iOS Beta users


alexduf:
just to double check if I understand correctly:
this handle the case only in memory but will store it in Azure using lower case only, is that right?

davidfurey:
Yes, the version in Azure will just be user:0e980097-59fd-4047-b609-366c6d5bb1b3 (lowercase, without the "gia:" prefix)

alexduf:
Great :+1:

PR #87, opened 2016-09-09
Newsstand Registration

davidfurey:
Allows newsstand devices to  register for notifications

Subsequent PR will have changes to notifications microservice to allow sending of newsstand notifications

### DIFF
--- a/common/src/main/scala/models/UniqueDeviceIdentifier.scala
+++ b/common/src/main/scala/models/UniqueDeviceIdentifier.scala
@@ -28,20 +28,14 @@ object UniqueDeviceIdentifier {
   }
 
   def apply(id: UUID): UniqueDeviceIdentifier =
-    new UniqueDeviceIdentifier(id)
+    new UniqueDeviceIdentifierImpl(id)
 }
 
-class UniqueDeviceIdentifier(val id: UUID) {
+case class UniqueDeviceIdentifierImpl(id: UUID) extends UniqueDeviceIdentifier
 
-  override def toString: String = s"UniqueDeviceIdentifier($id)"
+trait UniqueDeviceIdentifier {
+
+  def id: UUID
 
   def legacyFormat: String = id.toString
-
-  override def equals(that: Any): Boolean = that match {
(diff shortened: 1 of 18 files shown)


common/src/main/scala/models/UniqueDeviceIdentifier.scala
```diff
@@ -11,36 +11,35 @@ object UniqueDeviceIdentifier {
...
+  def apply(id: UUID): UniqueDeviceIdentifier =
+    new UniqueDeviceIdentifier(id)
 }
+
+class UniqueDeviceIdentifier(val id: UUID) {
+
+  def legacyFormat: String = id.toString
+
+  override def equals(that: Any): Boolean = that match {
+    case that: UniqueDeviceIdentifier if that.toString == this.toString => true
```
alexduf: I'm not sure why you are using toString here. Could you be using `that.id == this.id` instead?

another way to deal with it would be to have a trait `UniqueDeviceIdentifier` defined as follow:

```trait UniqueDeviceIdentifier {
  def id: UUID
}```

Then the `IOSUdid` can be a case class extending that trait.
```
case class IOSUDid(underlying: String, id: UUID) extends UniqueDeviceIdentifier
```

Unless I missed something in which case let me know :)

davidfurey: I thought intuitive equality would be `UniqueDeviceIdentifier != IosUdid` even if the UUID was the same. Is there a downside to using the toString as the basis for equality?

alexduf: It's just that I see `toString` as being a technical method almost for debugging purposes. It's not explicit when you look at a class definition that `toString` will be used as part of the object's contract. The [java doc](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#toString()) states its result as being "easy for a person to read"

So I agree that the code will definitely work but it's relying on something not exactly meant for it.

Also it forces you to implement hashcode as well, whereas using a trait implemented by case classes will give you the whole equality (and hashing function) for free.

PR #90, opened 2016-09-19
Batch topic subscriptions

davidfurey:
Update topic subscriptions repository every 60s.  This should avoid the flood of dynamodb writes when we send a breaking news alert

### DIFF
--- a/common/src/main/scala/tracking/BatchingTopicSubscriptionsRepository.scala
+++ b/common/src/main/scala/tracking/BatchingTopicSubscriptionsRepository.scala
@@ -1,0 +1,94 @@
+package tracking
+import akka.actor.ActorSystem
+import models.Topic
+import tracking.Repository.RepositoryResult
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import akka.actor.{Actor, Props}
+import akka.pattern.ask
+import akka.util.Timeout
+import cats.implicits._
+
+import scala.concurrent.duration.FiniteDuration
+
+object BatchingTopicSubscriptionsRepository {
+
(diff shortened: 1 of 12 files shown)


common/src/main/scala/tracking/BatchingTopicSubscriptionsRepository.scala
```diff
@@ -0,0 +1,94 @@
...
+    var events = collection.mutable.Buffer[SubscriptionEvent]()
+
+    override def receive: Actor.Receive = {
+      case event: SubscriptionEvent =>
+        events.append(event)
+
+      case Flush =>
+        if (events.nonEmpty) {
+          val requestor = sender
+          flush() foreach { _ => requestor ! () }
```
alexduf: why changing the name of the `sender` variable here?

davidfurey: Turns out that sender is a def:
http://doc.akka.io/japi/akka/2.4/akka/actor/Actor.html#sender--

alexduf: so you are saving calling that def in a loop, right?

davidfurey: `flush()` returns a future, so I am avoiding getting the wrong value for `sender` since the actor might be processing a new message (and therefore have a new `sender`) by the time the future has completed.

alexduf: Ah! got you, it makes perfect sense then

PR #91, opened 2016-09-19
Limit number of topics to 200

davidfurey:
Because the payload to Azure gets to large with more than 200 topics

### DIFF
--- a/common/src/main/scala/models/TopicType.scala
+++ b/common/src/main/scala/models/TopicType.scala
@@ -2,17 +2,19 @@ package models
 
 import play.api.libs.json._
 
-sealed trait TopicType
+case class TopicType(name: String, priority: Int) {
+  override def toString: String = name
+}
 
 object TopicTypes {
-  case object Breaking extends TopicType { override def toString: String = "breaking" }
-  case object Content extends TopicType { override def toString: String = "content" }
-  case object TagContributor extends TopicType { override def toString: String = "tag-contributor" }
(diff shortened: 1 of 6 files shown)


DiegoVazquezNanini:
:+1: very neat

common/src/main/scala/models/TopicType.scala
```diff
@@ -2,17 +2,19 @@ package models
...
 object TopicTypes {
-  case object Breaking extends TopicType { override def toString: String = "breaking" }
-  case object Content extends TopicType { override def toString: String = "content" }
-  case object TagContributor extends TopicType { override def toString: String = "tag-contributor" }
-  case object TagKeyword extends TopicType { override def toString: String = "tag-keyword" }
-  case object TagSeries extends TopicType { override def toString: String = "tag-series" }
-  case object TagBlog extends TopicType { override def toString: String = "tag-blog" }
-  case object FootballTeam extends TopicType { override def toString: String = "football-team" }
-  case object FootballMatch extends TopicType { override def toString: String = "football-match" }
+  object Breaking extends TopicType("breaking", 100) // scalastyle:off magic.number
```
alexduf: Don't hesitate to remove the rule in the scala style conf if that gets in your way

alexduf: Or to drop it completely

davidfurey: I mostly like the rule.  Except when it is annoying.  I'll leave it there for now.

PR #93, opened 2016-09-20
Increase logging for tracking/dynamodb operations

davidfurey:
We’re writing much more than expected to dynamodb, this should help us
understand why


DiegoVazquezNanini:
:+1:

PR #96, opened 2016-09-22
Refuse to send duplicate notifications

davidfurey:
Goal alerts sent some duplicates. Hopefully we can fix this at source at some point and remove this check

### DIFF
--- a/notification/test/notification/controllers/MainSpec.scala
+++ b/notification/test/notification/controllers/MainSpec.scala
@@ -12,7 +12,7 @@ import org.specs2.matcher.JsonMatchers
 import org.specs2.mock.Mockito
 import org.specs2.specification.Scope
 import play.api.test.PlaySpecification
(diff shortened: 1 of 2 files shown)


notification/test/notification/controllers/MainSpec.scala
```diff
@@ -48,20 +55,7 @@ class MainSpec(implicit ec: ExecutionEnv) extends PlaySpecification with Mockito
       val response = main.pushTopics()(request)
 
       status(response) must equalTo(CREATED)
-      there was one(reportRepository).store(expectedReport)
-    }
-
-    "notify reporting repository about added notifications and propagate reporting error" in new MainScope {
```
pvighi: is this test not needed anymore?

davidfurey: Yes. If the reporting repository fails the notification will fail, since it needs the reporting repo for the de-duplication.

PR #97, opened 2016-09-23
Timeout requests after 30 seconds

davidfurey:
I'm hoping this might fix the CLOSE_WAIT issue

### DIFF
--- a/registration/app/registration/controllers/Main.scala
+++ b/registration/app/registration/controllers/Main.scala
@@ -1,5 +1,7 @@
 package registration.controllers
 
+import akka.actor.ActorSystem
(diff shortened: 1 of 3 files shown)


registration/app/registration/controllers/Main.scala
```diff
@@ -135,4 +139,14 @@ final class Main(
       InternalServerError(other.reason)
   }
```
DiegoVazquezNanini: What do you think about moving the filter to a different file?

davidfurey: I'd like to, it would be good to, and started doing it, but it depends on quite a few things in this class

PR #99, opened 2016-09-27
Disable mutable-content flag

davidfurey:
This seems to be causing issues for some users.  Disabling until we work out what is causing the problem


davidfurey:
cc @fgauchet

DiegoVazquezNanini:
:+1:

PR #100, opened 2016-09-27
Expose notification reports from Azure

davidfurey:


common/src/main/scala/models/JsonUtils.scala
```diff
@@ -47,4 +47,14 @@ object JsonUtils {
 
     override def writes(o: URI): JsValue = JsString(o.toString)
   }
+
+  def stringFormat[T](fromString: String => Option[T]): Format[T] = new Format[T] {
```
alexduf: this is a good idea 👍

report/app/report/controllers/Report.scala
```diff
@@ -37,4 +41,10 @@ final class Report(
     }
   }
 
+  def notification(id: UUID): Action[AnyContent] = AuthenticatedAction.async {
+    XorT(reportRepository.getByUuid(id)).semiflatMap(reportEnricher.enrich).fold(
+      { error => InternalServerError(error.message) },
```
alexduf: [Pedantic] Am i right to say you don't need the curly braces here?

davidfurey: You are right, and it looks nicer without them

PR #101, opened 2016-09-28
Return correct response for legacy clients

davidfurey:
Returning the correct response will allow legacy clients to update
their local list of subscribed topics.


alexduf:
It looks good to me! 👍

PR #102, opened 2016-09-29
Connect directly to content api for content notification expiry

davidfurey:
In preparation for replacing content notifications with a lambda


davidfurey:
Depends on https://github.com/guardian/mobile-platform/pull/460

common/src/main/scala/auditor/package.scala
```diff
@@ -1,26 +1,63 @@
...
+          case false => Some(topic)
+        }
+      }.map(_.flatten)
+    }
+
+    def isLiveBloggingNow(id: String)(implicit ec: ExecutionContext): Future[Boolean] = {
+      val url = s"${config.url}/$id?show-fields=liveBloggingNow&api-key=${config.apiKey}"
+      wsClient.url(url).get().map { response =>
+        val liveBloggingNowField = response.json \ "response" \ "content" \ "fields" \ "liveBloggingNow"
+        liveBloggingNowField.validate[String].asOpt.flatMap(s => Try(s.toBoolean).toOption).getOrElse(false)
```
alexduf: I think you can use `validateOpt[Boolean].getOrElse(false)` here

common/src/main/scala/auditor/package.scala
```diff
@@ -1,26 +1,63 @@
...
+      } recover { case _ => true }
```
alexduf: why defaulting to false above, but recovering to true here, I feel like I missed something?

common/src/main/scala/auditor/package.scala
```diff
@@ -1,26 +1,63 @@
...
+
+    override def expiredTopics(topics: Set[Topic])(implicit ec: ExecutionContext): Future[Set[Topic]] = {
+      val contentTopics = topics.filter(_.`type` == TopicTypes.Content)
+
+      Future.traverse(contentTopics) { topic =>
+        isLiveBloggingNow(topic.name).map {
+          case true => None
+          case false => Some(topic)
+        }
+      }.map(_.flatten)
```
DiegoVazquezNanini: flatMap instead?

davidfurey: I don't think that will work

common/src/main/scala/auditor/package.scala
davidfurey: The "liveBloggingNow" field is a string not a boolean in the json, so I don't think it would parse

DiegoVazquezNanini:
:+1: Looks good to me, but maybe someone else should take a second look.
Cheers

common/src/main/scala/auditor/package.scala
davidfurey: If the field does not exist then it isn't a live blog but if we can't connect to content-api we have no idea if it is a live blog or not, so 'true' is a safer guess.

common/src/main/scala/auditor/package.scala
```diff
@@ -1,26 +1,63 @@
...
-import models.Topic
+import models.{Topic, TopicTypes}
+import play.api.Logger
 import play.api.libs.json.Json
+import play.api.libs.ws.WSClient
 
 import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
 
 package object auditor {
```
davidfurey: Fair.  I think there is probably scope for a general refactor of this project soon, now that it has gotten a bit larger.

alexduf: Yeah I think that's a good idea, starting by putting general package names (my bad :/)

common/src/main/scala/auditor/package.scala
alexduf: Ah I see

common/src/main/scala/auditor/package.scala
alexduf: That makes sense then!

PR #104, opened 2016-10-06
Elections notifications

davidfurey:
Follow up PR will be required later, but this should allow US team to start testing

### DIFF
--- a/common/src/main/scala/models/elections/ElectionResults.scala
+++ b/common/src/main/scala/models/elections/ElectionResults.scala
@@ -1,0 +1,25 @@
+package models.elections
+
+import java.net.URI
+
+import models.JsonUtils
+import play.api.libs.json.Json
+
+case class CandidateResults(
+  name: String,
+  states: List[String],
+  electoralVotes: Int,
+  popularVotes: Int,
+  avatar: URI
+)
+
+object CandidateResults {
+  import JsonUtils._
+  implicit val jf = Json.format[CandidateResults]
+}
+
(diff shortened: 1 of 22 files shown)


common/src/main/scala/models/elections/ElectionResults.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import java.net.URI
+
+import models.JsonUtils
+import play.api.libs.json.Json
+
+case class CandidateResults(
+  name: String,
+  states: List[String],
+  electoralVotes: Int,
```
alexduf: Should these be `Long` instead? _(or am I misunderstanding the US election system ?)_

notification/app/notification/services/azure/WNSPushConverter.scala
```diff
@@ -82,4 +83,10 @@ class WNSPushConverter(conf: Configuration) {
     addedTime = gan.addedTime,
     debug = conf.debug
   )
+
+  private def toElectionResult(el: ElectionNotification) = wns.ElectionNotification(
```
alexduf: If this isn't already done, I think it would be worth checking with the win10 team to see if this isn't going to cause any issue.

common/src/main/scala/models/elections/ElectionResults.scala
alexduf: Ahah for some reason I was missing three zeros for max_int. My bad, please ignore

alexduf:
👍

PR #105, opened 2016-10-10
Enterprise notifications

davidfurey:
iOS have a different certificate for 'enterprise' (internal beta) builds.  So we need to use a different notification hub for these devices.


DiegoVazquezNanini:
:+1: Look good

PR #111, opened 2016-11-11
Revert "Disable mutable-content flag"

davidfurey:
Reverts guardian/mobile-n10n#99

@fgauchet has decided that enough users have upgraded to the version of iOS that resolves this issue.


fgauchet:
👍

PR #113, opened 2016-11-11
Add iosUseMessage option for content notifications

davidfurey:
Temporary workaround so that we can send a longer message to android
and iOS


davidfurey:
We should revert this on Monday

PR #114, opened 2016-11-21
Migrate users from weekend-roundup to weekend-reading

davidfurey:
A/B test showed this variant was better

@maxspencer


pvighi:
looks ok to me 👍

davidfurey:
iOS doesn't allow the registrations service to automatically subscribe it to a topic, just to unsubscribe.

PR #115, opened 2016-11-22
Send weekend-reading notifications to weekend-round-up too

davidfurey:
These were part of an a/b test, we have decided that the weekend-reading format is better.

@maxspencer

### DIFF
--- a/common/src/main/scala/models/Notification.scala
+++ b/common/src/main/scala/models/Notification.scala
@@ -14,6 +14,7 @@ sealed trait Notification {
   def message: String
   def importance: Importance
   def topic: Set[Topic]
+  def withTopics(topics: Set[Topic]): Notification
 }
 
 object Notification {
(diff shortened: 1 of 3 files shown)


DiegoVazquezNanini:
What is the difference between this PR and the previous one?
You are now adding the withTopics function to all notification types and appending the weekendRoundUpTopic if the user suscribes to weekendReadingTopic?

davidfurey:
@DiegoVazquezNanini 

**Previous PR**
Attempted to unsubscribe users from weekend-round-up and then subscribe them to weekend-reading (when they called the registration endpoint).  But it didn't work

**This PR**
Does not change users registrations at all.  Users who are subscribed to weekend-reading stay subscribed to weekend-reading, users subscribed to weekend-round-up stay subscribed to weekend-roundup.  

When a notification is sent to the weekend-reading topic, it is forwarded to weekend-round-up as well.  So both sets of users receive the notification.

DiegoVazquezNanini:
Thanks for the clarification @davidfurey

PR #116, opened 2016-12-07
Add extra slashes to topic serialisation

davidfurey:
Move to n10n created a regression since android expects an extra slash
in topic names.  This should fix it.


common/src/main/scala/models/Topic.scala
```diff
@@ -7,6 +7,7 @@ import cats.data.Xor
 
 case class Topic(`type`: TopicType, name: String) {
   override def toString: String = s"${`type`}/$name"
+  def androidToString: String = s"${`type`}//$name"
```
alexduf: [Nitpicking] I think I would rather see the logic duplicated at the time it's needed than putting platform specific logic on a generic model

PR #118, opened 2016-12-16
Validate football matches directly with PA

davidfurey:
Football endpoint is being decommissioned
Depends on https://github.com/guardian/mobile-platform/pull/484


davidfurey:
We should.  https://theguardian.atlassian.net/browse/MAPI-1506

PR #121, opened 2016-12-20
Remove topics field from android breaking news

davidfurey:
Including this field causes the app to filter out these notifications because it doesn’t model breaking news as breaking//<edition>


alexduf:
I'm confused about why we need to remove the `topics` attribute. Shouldn't we revert https://github.com/guardian/mobile-n10n/pull/116 instead?

PR #122, opened 2017-01-09
Add LRU cache around is match ended

lmath:
Naturally, the first thing I tried to determine was what are the consequences of using stale values from the cache? It looks like that just means associating an expired topic with a device, and it would stay there until removed by a later update. It looks to me like this should at worst just mean maintaining this association for no reason. 

I don't actually know when device registration/updates happen though so it's not clear to me when it's sensible to expire the data. Default capacity for the cache is 500, so it seems to me I should have some kind of TTL and/or TTI unless I'd like the data to be stale most of the time. 

I made a guess: TTL 5 minutes, TTI 1 minute. Thoughts?

@alexduf @davidfurey

### DIFF
--- a/common/src/main/scala/auditor/package.scala
+++ b/common/src/main/scala/auditor/package.scala
@@ -4,6 +4,7 @@ import pa.PaClient
 import play.api.Logger
 import play.api.libs.json.Json
 import play.api.libs.ws.WSClient
+import spray.caching.{Cache, LruCache}
 
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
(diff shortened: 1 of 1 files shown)


davidfurey:
I don't think the TTI needs to be less than the TTL in this case.  since we probably won't fill a cache with capacity 500.  

TTL of 1 minute, 5 minute or even 30 minutes should be fine.  It just changes how long it takes for the match to automatically drop off the users list of notification topics.

davidfurey:
👍

lmath:
Hmm, yes, a TTI as well is overcomplicating things then. I'll amend to a TTL of 5 minutes.

PR #123, opened 2017-01-27
Super bowl notification

davidfurey:
Adds new live-event notification for superbowl, which is hopefully generic enough to be used for other live events.

### DIFF
--- a/common/src/main/scala/azure/APNSRawPush.scala
+++ b/common/src/main/scala/azure/APNSRawPush.scala
@@ -2,11 +2,16 @@ package azure
 
 import azure.apns.Body
 import play.api.http.Writeable
-import play.api.libs.json.{Json, JsValue}
+import play.api.libs.json.{JsValue, Json}
+import play.api.libs.ws.{WSRequest, WSResponse}
 import utils.WriteableImplicits._
 
-case class APNSRawPush(body: Body, tags: Option[Tags]) extends RawPush[Body] {
+import scala.concurrent.Future
+
+case class APNSRawPush(body: Body, tags: Option[Tags]) extends RawPush {
   override def format: String = "apple"
 
-  override def writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
+  implicit private val writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
+
+  override def post(request: WSRequest): Future[WSResponse] = request.post(body)
 }
--- a/common/src/main/scala/azure/APNSRawPush.scala
+++ b/common/src/main/scala/azure/APNSRawPush.scala
@@ -11,7 +11,7 @@ import scala.concurrent.Future
 case class APNSRawPush(body: Body, tags: Option[Tags]) extends RawPush {
   override def format: String = "apple"
 
-  implicit private val writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
+  private val writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
 
-  override def post(request: WSRequest): Future[WSResponse] = request.post(body)
+  override def post(request: WSRequest): Future[WSResponse] = request.post(body)(writeable)
 }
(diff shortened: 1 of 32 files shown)


common/src/main/scala/azure/APNSRawPush.scala
```diff
@@ -2,11 +2,16 @@ package azure
...
-case class APNSRawPush(body: Body, tags: Option[Tags]) extends RawPush[Body] {
+import scala.concurrent.Future
+
+case class APNSRawPush(body: Body, tags: Option[Tags]) extends RawPush {
   override def format: String = "apple"
 
-  override def writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
+  private val writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
+
+  override def post(request: WSRequest): Future[WSResponse] = request.post(body)(writeable)
```
pvighi: any reason in particular to move the post method here?

common/src/main/scala/azure/apns/LiveEventProperties.scala
```diff
@@ -0,0 +1,17 @@
+package azure.apns
+
+import play.api.libs.json.Json
+
+case class LiveEventProperties(
+  title: String,
+  body: String,
+  richviewbody: String,
```
pvighi: **very minor detail:** 
Does richviewbody have to be called like this when converted to json or can we make it camel case ?

common/src/main/scala/models/Notification.scala
```diff
@@ -134,3 +136,25 @@ object ElectionNotification {
...
+  id: UUID,
+  `type`: NotificationType = LiveEventAlert,
+  sender: String,
+  title: String,
+  message: String,
+  expandedMessage: Option[String],
+  shortMessage: Option[String],
+  importance: Importance,
+  link1: Link, // where does the title for this come from?
+  link2: Link, // where does the title for this come from?
```
pvighi: I'm a little confused by these two comments. Are both links where the title comes from ?

notification/app/notification/controllers/Main.scala
```diff
@@ -34,7 +34,7 @@ final class Main(
 
   override def isPermittedTopic(apiKey: String): Topic => Boolean = {
     if (configuration.electionRestrictedApiKeys.contains(apiKey)) {
-      topic => topic.`type` == ElectionResults
+      topic => List(ElectionResults, LiveNotification).contains(topic.`type`)
```
pvighi: Should election restricted api keys (never heard of them but the name seems self explanatory) also be allowed to use this new topic type ?

common/src/main/scala/azure/APNSRawPush.scala
davidfurey: I ran into issues with RawPush[T] needing to be covariant in T because of `def body: T` and contravariant in T because of ` def writeable: Writeable[T]`.  Moving the post into RawPush removed the need for writeable to be directly exposed in the interface which avoided the problem

common/src/main/scala/azure/apns/LiveEventProperties.scala
davidfurey: This is to match what iOS are expecting

common/src/main/scala/models/Notification.scala
davidfurey: I should remove the comment.  I wanted to know where the text that would displayed for each link came from.  It turns out it is baked into the app.

notification/app/notification/controllers/Main.scala
davidfurey: Yes, we need to rename electionRestrictedApiKeys to mobileInnovationLabRestrictedApiKeys.  It is so that the team in the US cannot accidentally send breaking news or other notifications that they don't need to.

notification/app/notification/services/azure/NotificationsHubSender.scala
```diff
@@ -34,23 +36,24 @@ abstract class NotificationsHubSender(
...
-        count <- count(push.destination)
-      } yield {
-        result.fold(
-          e => NotificationRejected(NotificationHubSenderError(e.some).some).left,
-          id => report(id, count.toOption).right
-        )
-      }
-    } else {
-      Future.successful(report(None, None).right)
+    converter.toRawPush(push) match {
```
pvighi: This looks like it's only needed if shouldSendToApps(push.notification) == true. 
I guess if it's not too much extra processing it's worth it to simplify this code

pvighi:
LGTM 👍

DiegoVazquezNanini:
I would be good to have better names than Link1 and Link2, but I guess IOS needs them. Right?

:+1: LGTM

davidfurey:
I agree, link1 and link2 aren't ideal, but it is for iOS

PR #124, opened 2017-01-27
Survey notification

davidfurey:
Adds a new notification type for sending out surveys


common/src/main/scala/models/Notification.scala
```diff
@@ -25,13 +25,16 @@ object Notification {
       case n: ContentNotification => ContentNotification.jf.writes(n)
       case n: GoalAlertNotification => GoalAlertNotification.jf.writes(n)
       case n: ElectionNotification => ElectionNotification.jf.writes(n)
+      case n: LiveEventNotification => LiveEventNotification.jf.writes(n)
```
DiegoVazquezNanini: SurveyNotification missing

davidfurey:
Not required for now, may re-open later

PR #125, opened 2017-01-27
Category should not be ITEM_CATEGORY for external links

davidfurey:
This should allow us to send external links (for surveys) to iOS


DiegoVazquezNanini:
What happens if it's category = "ITEM_CATEGORY" on an External link?

davidfurey:
@DiegoVazquezNanini item_category makes the usual breaking news buttons show up - save and share, which don't make sense for external links.

pvighi:
👍

PR #132, opened 2017-05-15
The new ELBs don't seem to like the fancy close-my-connection response.

alexduf:
We currently are observing a reboot-loop on our services, and it started on the 24th of march, unrelated to our changes.
I can only assume AWS has changed the way ELBs are handling healthcheck and don't like getting these responses back anymore.

I have deployed that branch already and am currently monitoring it.


alexduf:
It looks like the problem is still happening, so I made the wrong diagnosis

PR #133, opened 2017-05-23
parse android's version of the legacy topic.

alexduf:
This is a temporary workaround that should be removed once android has spread their fix wide enough.

cc @mohammad-haque


NathanielBennett:
:+1:

PR #136, opened 2017-08-07
Configure newsstand notifications

NathanielBennett:
Configures the newsstand notification sender to use the correct endpoint

### DIFF
--- a/notification/app/notification/NotificationApplicationLoader.scala
+++ b/notification/app/notification/NotificationApplicationLoader.scala
@@ -82,7 +82,7 @@ trait AzureHubComponents {
   lazy val apnsNotificationSender: APNSSender = wire[APNSSender]
 
   lazy val newsstandNotificationSender: NewsstandSender = {
-    val hubClient = new NotificationHubClient(appConfig.enterpriseHub, wsClient)
+    val hubClient = new NotificationHubClient(appConfig.newsstandHub, wsClient)
     new NewsstandSender(hubClient)
   }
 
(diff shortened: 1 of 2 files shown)


notification/app/notification/NotificationApplicationLoader.scala
```diff
@@ -80,7 +81,10 @@ trait AzureHubComponents {
 
   lazy val apnsNotificationSender: APNSSender = wire[APNSSender]
 
-  lazy val newsstandNotificationSender: NewsstandSender = wire[NewsstandSender]
+  lazy val newsstandNotificationSender: NewsstandSender = {
+    val hubClient = new NotificationHubClient(appConfig.enterpriseHub, wsClient)
```
alexduf: you probably want to use `appConfig.newsstandHub`

notification/app/notification/services/Configuration.scala
```diff
@@ -35,4 +30,13 @@ class Configuration extends NotificationConfiguration("notification") {
 
   lazy val disableElectionNotificationsAndroid = conf.getStringProperty("notifications.elections.android.disabled", "false").toBoolean
   lazy val disableElectionNotificationsIOS = conf.getStringProperty("notifications.elections.ios.disabled", "false").toBoolean
+
+  private def getConfigurableHubConnection(maybeEndpoint: Option[String]): NotificationHubConnection = {
```
alexduf: I think this function probably should receive a string as a parameter, either "enterprise" or "newstand".

That string would then be prepended to every key when getting every property.

That would make it easier to understand
```
getConfigurableHubConnection("enterprise")
getConfigurableHubConnection("newsstand")
```

PR #139, opened 2017-08-11
fix deep-link construction

alexduf:
This is to unblock @aodhol in his task to speedup the ios app when starting from a deeplink.

it should allow ios to use the uri instead of the link.

No client is currently reading the uri so updating it should be safe.


NathanielBennett:
:+1:

PR #140, opened 2017-08-14
send full web url to client side.

alexduf:
It's apprently simpler for the ios client side to resolve with the full url


alexduf:
cc @aodhol

PR #142, opened 2017-08-25
Add support for the new Football Status model

alexduf:
That model supports both the matchInfoUri and the articleUri


NathanielBennett:
:+1: Looks good

PR #147, opened 2017-10-05
Nb mss 109 use amigo

NathanielBennett:
This alters the notification, registration and report services to be deployed as debs to amigo. 

*Uses @alexduf's library to load application configuration from S3
*This meant updating the aws dependencies which in turn caused some little alterations


build.sbt
```diff
@@ -13,25 +16,29 @@ val standardSettings = Seq[Setting[_]](
...
     ),
     libraryDependencies ++= Seq(
       json,
       ws,
       "com.microsoft.azure" % "azure-servicebus" % "0.7.0",
       "org.typelevel" %% "cats" % "0.7.0",
       "joda-time" % "joda-time" % "2.8.2",
-      "com.amazonaws" % "aws-java-sdk" % "1.9.31",
+      "com.amazonaws" % "aws-java-sdk" % "1.11.205",
+      "com.amazonaws" % "aws-java-sdk-dynamodb" % "1.11.205",
```
alexduf: Isn't dynamoDB part of the SDK just above?

build.sbt
```diff
@@ -44,20 +51,22 @@ lazy val common = project
...
+  .enablePlugins(RiffRaffArtifact, RiffRaffArtifact, JDebPackaging)
   .settings(standardSettings: _*)
   .settings(
     libraryDependencies ++= Seq(
       "com.typesafe.play" %% "play-logback" % "2.5.3",
       "com.microsoft.azure" % "azure-storage" % "3.1.0"
     ),
-    riffRaffPackageType := (packageZipTarball in Universal).value,
-    version := "1.0-SNAPSHOT"
+    riffRaffPackageType := (packageBin in Debian).value,
```
alexduf: I saw you hadn't modified the backup cloudformation script in the corresponding PR. I don't think you need to touch this one.

notification/app/notification/NotificationApplicationLoader.scala
```diff
@@ -22,7 +25,16 @@ import scala.concurrent.ExecutionContext
 class NotificationApplicationLoader extends ApplicationLoader {
   override def load(context: Context): Application = {
     LoggerConfigurator(context.environment.classLoader) foreach { _.configure(context.environment) }
-    (new BuiltInComponentsFromContext(context) with AppComponents).application
+    val identity = AppIdentity.whoAmI(defaultAppName = "notification", defaultStackName = "mobile-notification")
+    val config = ConfigurationLoader.load(identity) {
+      case AppIdentity(app, stack, stage, _) => S3ConfigurationLocation (
```
alexduf: I just remembered the conversation we were having about it. I was wrong:
Configured like that, the loader will always go to s3, since the first case will always match, even if stage == "DEV"

Alternatively you could define it like that:
`case AppIdentity(app, stack, "CODE" | "PROD", _)`

So the DEV case falls back on the default. Maybe this is a sign I should formulate the library differently? Happy to hear your about on it.

NathanielBennett: No, that's my bad - I should have been clearer about what I wanted. The solution you propose doesn't quite work though. There needs to be some way if capturing the value for stage. Does 

`case AppIdentity(app, stack, stage) if List("CODE","PROD").contains(stage) =>` . 

Does that work?

alexduf: You're right that's better, maybe I'd go one step further and use `if (stage != "DEV")`

build.sbt
NathanielBennett: Yup you're right. That's what I need to do. Pay attention and take care . :(

notification/app/notification/NotificationApplicationLoader.scala
alexduf: Maybe I could enhance the configuration library by providing two different types depending on if it's a live environment or a test one.

NathanielBennett: I'll fix it as suggested and then if you've time to upgrade the library, I can make a change?

alexduf: oh yeah these are two separate problems. Just throwing ideas out there.

PR #149, opened 2017-10-12
Nb mss 159 convert backup to lambda

NathanielBennett:
Usual mess of commits which I'll squash when I merge. 

Converts the n10n backup application to run as a lamba


backup/src/main/scala/backup/Lambda.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  val config = Configuration.load()
+
+  def handler() : String = {
+
+    val wSClient = AhcWSClient()
+    val backup = new Backup(config, wSClient)
+
+    logger.info("Starting backup")
+    backup.execute().onComplete {
```
alexduf: simple, efficient 👍

PR #150, opened 2018-02-05
remove tag migration

alexduf:
This was once used to migrate some of our clients, but this isn't
necessary anymore.
It also is pushing us above 20 topics if we receive 20 topics,
triggering a server side error.


davidfurey:
🎉

PR #151, opened 2018-02-23
Nb mss 162 upgrade play 2.6

NathanielBennett:
Upgrades n10n to play 2.6. 
Had to flatten the applicationLoaders in order be able to inject ````ControllerComponents```` into the controllers


build.sbt
```diff
@@ -1,8 +1,16 @@
 import com.gu.riffraff.artifact.RiffRaffArtifact.autoImport._
-import com.typesafe.sbt.packager.archetypes.ServerLoader.Systemd
 
 val projectVersion = "1.0-latest"
 
+scalaVersion in ThisBuild := "2.12.4"
+
+scalacOptions in ThisBuild ++= Seq(
+  "-deprecation",
+  "-Xfatal-warnings",
```
alexduf: bold move :+1:

build.sbt
```diff
@@ -28,14 +36,15 @@ lazy val common = project
...
     ),
     libraryDependencies ++= Seq(
-      json,
       ws,
       "com.microsoft.azure" % "azure-servicebus" % "0.7.0",
       "org.typelevel" %% "cats" % "0.7.0",
       "joda-time" % "joda-time" % "2.8.2",
       "com.gu" %% "configuration" % "4.1",
       "io.spray" %% "spray-caching" % "1.3.3",
-      "com.typesafe.play" %% "play-logback" % "2.5.3",
```
alexduf: I wonder why we had this one?

common/src/main/scala/azure/APNSRawPush.scala
```diff
@@ -1,17 +1,14 @@
...
 import play.api.libs.ws.{WSRequest, WSResponse}
-import utils.WriteableImplicits._
 
 import scala.concurrent.Future
 
 case class APNSRawPush(body: Body, tags: Option[Tags]) extends RawPush {
   override def format: String = "apple"
 
-  private val writeable: Writeable[Body] = implicitly[Writeable[JsValue]].map(Json.toJson[Body]).withContentType("application/json;charset=utf-8")
+  implicit val writer = implicitly(bodyWritable[Body])
```
alexduf: Do you need to set the content type here?

common/src/main/scala/azure/APNSRawPush.scala
```diff
@@ -1,17 +1,14 @@
...
 
-  override def post(request: WSRequest): Future[WSResponse] = request.post(body)(writeable)
+  override def post(request: WSRequest): Future[WSResponse] = request.post[Body](body)//(writeable)
```
alexduf: you can drop (writeable) if you don't need it

common/src/main/scala/azure/WNSRawPush.scala
```diff
@@ -1,16 +1,14 @@
 package azure
 import play.api.http.Writeable
-import play.api.libs.ws.{WSRequest, WSResponse}
+import play.api.libs.ws.{DefaultBodyWritables, WSRequest, WSResponse}
 import utils.WriteableImplicits._
 
 import scala.concurrent.Future
 
-case class WNSRawPush(body: String, tags: Option[Tags]) extends RawPush {
+case class WNSRawPush(body: String, tags: Option[Tags]) extends RawPush with DefaultBodyWritables {
```
alexduf: I think I'd rather see it as a import
``` scala
import play.api.libs.ws.DefaultBodyWritables._
```

It should give you the same result, and that avoids tying all the members of `DefaultBodyWritables` to each instance of WNSRawPush

notification/app/notification/controllers/Main.scala
```diff
@@ -2,50 +2,40 @@ package notification.controllers
...
-  override def isPermittedTopic(apiKey: String): Topic => Boolean = {
-    if (configuration.electionRestrictedApiKeys.contains(apiKey)) {
-      topic => List(ElectionResults, LiveNotification).contains(topic.`type`)
-    } else {
-      _ => true
-    }
-  }
 
-  def healthCheck: Action[AnyContent] = Action {
+  def healthCheck = Action {
```
alexduf: I liked that type!

notification/conf/routes
```diff
@@ -1,5 +1,4 @@
 GET         /healthcheck                        notification.controllers.Main.healthCheck
 POST        /push/user/:userId                  notification.controllers.Main.pushUser(userId: java.util.UUID)
 POST        /push/topic                         notification.controllers.Main.pushTopics
-POST        /push/topic/*topic                  notification.controllers.Main.pushTopic(topic: Topic)
```
alexduf: I can't remember if this is still used. Have you checked?

PR #154, opened 2018-02-28
Cats 1 migration

alexduf:
Before migrating to the parameter store I upgraded our libs on mobile-notifications.
This is PR 1 out of 3

This on focuses only on migrating to the latest Cats version. That means using scala's 2.11 biased `Either` instead of the now deprecated `Xor`


NathanielBennett:
:+1: Just by itself this makes this code base *so* much nicer

PR #155, opened 2018-03-01
Mss 218 parameter store

alexduf:
There was a refactoring necessary in order to get rid of the deprecated guardian config library.
Then I switched it to the SSM parameter store.


common/src/main/scala/utils/CustomApplicationLoader.scala
```diff
@@ -0,0 +1,28 @@
...
+    LoggerConfigurator(context.environment.classLoader) foreach { _.configure(context.environment) }
+    val identity = AppIdentity.whoAmI(defaultAppName = appName)
+    val config = ConfigurationLoader.load(identity, credentials) {
+      case AwsIdentity(app, stack, stage, _) => SSMConfigurationLocation(s"/notifications/$stage/$stack")
+    }
+    val loadedConfig = Configuration(config)
+    val newContext = context.copy(initialConfiguration = context.initialConfiguration ++ loadedConfig )
+    buildComponents(newContext).application
+  }
+}
```
NathanielBennett: Nice. Why didn't I think of doing that four months ago. :+1:

registration/app/registration/services/Configuration.scala
```diff
@@ -2,38 +2,36 @@ package registration.services
...
   )
-  lazy val maxTopics = getConfigInt("notifications.max_topics")
-  lazy val dynamoTopicsTableName = getConfigString("db.dynamo.topics.table-name")
-  lazy val dynamoTopicsFlushInterval = getFiniteDuration("db.dynamo.topics.flush-interval").getOrElse(60.seconds)
+  lazy val maxTopics: Int = conf.get[Int]("notifications.max_topics")
+  lazy val dynamoTopicsTableName: String = conf.get[String]("db.dynamo.topics.table-name")
+  lazy val dynamoTopicsFlushInterval: FiniteDuration = conf.getOptional[FiniteDuration]("db.dynamo.topics.flush-interval").getOrElse(60.seconds)
 
-  lazy val defaultTimeout = getFiniteDuration("routes.defaultTimeout").getOrElse(30.seconds)
+  lazy val defaultTimeout: FiniteDuration = conf.getOptional[FiniteDuration]("routes.defaultTimeout").getOrElse(30.seconds)
```
NathanielBennett: Why are you de-lazifying(?) the Notification configuration values but not the ones here?

NathanielBennett:
:+1: Refactoring to a cleaner solution is always awesome. One question, but approving

alexduf:
@NathanielBennett you haven't posted your question, what's up?

NathanielBennett:
The one about lazy vals in one configuration but not in the other? See above

registration/app/registration/services/Configuration.scala
alexduf: I was de-lazifying to try to reduce the questions one might have when reading the conf, mainly why is it lazy?

Turns out the registration config needs to be lazy in order for the tests to pass.

I reverted to have it all lazy

backup/src/main/scala/backup/Lambda.scala
```diff
@@ -5,41 +5,47 @@ import akka.stream.ActorMaterializer
...
+  private def withWsClient[A](block: WSClient => Future[A]): Future[A] = {
+    implicit val actorSystem: ActorSystem = ActorSystem()
+    implicit val actorMaterializer: ActorMaterializer = ActorMaterializer()
+    val ws = AhcWSClient()
 
-  val config = Configuration.load(credentials)
+    block(ws)
+      .andThen { case _ => ws.close() }
+      .andThen { case _ => actorSystem.terminate() }
+  }
```
NathanielBennett: Nice. Does this fix the backup error thing or is that a separate issue?

alexduf: It makes running the lambda locally nicer as if an exception happens the whole jvm blocks and waits. I've also set `fork := true` just to be extra sure I don't have to close my sbt session

alexduf:
I'll let this run on CODE, then on Monday I'll insert the PROD config in the parameter store and deploy (this one isn't a Friday deployment)

PR #157, opened 2018-03-05
Bump

alexduf:
This is the last of the PRs I had lined up.

It bumps a whole lot of libraries, including scala to scala 2.12. (For some reason it was overriden [here](https://github.com/guardian/mobile-n10n/blob/master/project/common.scala#L10))

I think this PR is easier to read commit by commit


common/src/main/scala/utils/LruCache.scala
```diff
@@ -0,0 +1,80 @@
...
+}
+
+object LruCache {
+  def apply[V](
+    initialCapacity: Int = 16,
+    maxCapacity: Int = 256,
+    timeToLive: FiniteDuration = 1.hour
+  ): LruCache[V] =
+    new LruCache[V](initialCapacity, maxCapacity, timeToLive)
+}
```
NathanielBennett: At some point let's decide what to do about this - we're now doing this in 3 different places which makes me feel a little bit uneasy

PR #161, opened 2018-05-23
Partial Mobile Notifications Lambda - dynamodb integrated

markrichardsg:


notificationsschedulelambda/src/main/scala/com/gu/notificationschedule/dynamo/NotificationSchedulePersistenceImpl.scala
```diff
@@ -0,0 +1,61 @@
...
+  def write(notificationsScheduleEntry: NotificationsScheduleEntry, sent: Boolean, sent_epoch_s: Long): Unit
+}
+
+class NotificationSchedulePersistenceImpl(config: NotificationScheduleConfig, client: AmazonDynamoDBAsync) extends NotificationSchedulePersistence {
+  private val due_and_sent = "due_epoch_s_and_sent"
+
+  def query(): Seq[NotificationsScheduleEntry] = {
+    val scanRequest = new ScanRequest(config.notificationScheduleTable)
+      .withIndexName(due_and_sent)
+      .withFilterExpression("sent = :sent and due_epoch_s < :now")
```
alexduf: `and due_epoch_s > :now` right?

markrichardsg:
@NathanielBennett it was worth it, @alexduf had a good suggestion for a name change and then it evolved beyond just a stub, it has the logic in place for dynamodb.
Tests passed on my machine :( (race condition always hit on CI, fixed)

schedulelambda/src/main/scala/com/gu/notificationschedule/ProcessNotificationScheduleLambda.scala
```diff
@@ -3,39 +3,71 @@ package com.gu.notificationschedule
...
+  }
+
+  private def triggerNotifications(notificationsScheduleEntries: Seq[NotificationsScheduleEntry]) = {
+    cloudWatch.queueMetric("discovered", notificationsScheduleEntries.size, StandardUnit.Count)
     val nowEpoch = Instant.now().getEpochSecond
-    notificationsScheduleEntries.foreach(notificationSchedulePersistence.write(_, true, nowEpoch))
+    val triedUnits = notificationsScheduleEntries.map(notificationsScheduleEntry => triggerNotification(nowEpoch, notificationsScheduleEntry))
+    val failed = triedUnits.count(_.isFailure)
+    cloudWatch.queueMetric("trigger-failed", failed, StandardUnit.Count)
+    if(failed > 0) {
```
alexduf: failed < 0 => Success

schedulelambda/src/main/scala/com/gu/notificationschedule/ProcessNotificationScheduleLambda.scala
```diff
@@ -3,39 +3,71 @@ package com.gu.notificationschedule
...
+    else {
+      Failure(new Exception(s"$failed triggers failed out of ${triedUnits.size}"))
+    }
 
-    cloudWatch.sendMetricsSoFar()
+  }
+
+  private def triggerNotification(nowEpoch: Long, notificationsScheduleEntry: NotificationsScheduleEntry) = {
+    Try {
+      logger.warn("TODO: trigger notification {}", notificationsScheduleEntry)
```
alexduf: TODO

markrichardsg: Yup.
Any placeholder will do, I was trying to avoid doing the whole thing at once (just wanted a lambda wired in first... it's already evolved to include dynamodb, so could do the rest here too).

alexduf: Makes sense then

project/MergeFilesStrategy.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.apache.logging.log4j.core.config.plugins.processor.PluginCache
+import sbt.File
+import sbtassembly.MergeStrategy
+
+import scala.collection.JavaConverters.asJavaEnumerationConverter
+
+class MergeFilesStrategy extends MergeStrategy {
+  override def name: String = "mergeFilesStrategy"
+
+  override def apply(tempDir: File, path: String, files: Seq[File]): Either[String, Seq[(File, String)]] = {
```
alexduf: what's that for?

schedulelambda/src/main/scala/com/gu/notificationschedule/external/Jackson.scala
```diff
@@ -0,0 +1,15 @@
+package com.gu.notificationschedule.external
+
+import com.fasterxml.jackson.databind.{ DeserializationFeature, ObjectMapper }
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import com.fasterxml.jackson.module.scala.experimental.ScalaObjectMapper
+
+object Jackson {
```
alexduf: I think this isn't used, and unless I'm mistaken, won't be used in the future so you can drop the jackson library and this file.

project/MergeFilesStrategy.scala
markrichardsg: log4j2 plugin caches are needed for loading appenders. Amazon AWS Lambda log4j2 support provides it's own.
When sbt assembly is run this merges the files instead of attempting to guess at which to discard which can result in breaking logging.
Renaming to MergeLog4j2PluginCachesStrategy.

schedulelambda/src/main/scala/com/gu/notificationschedule/external/Jackson.scala
markrichardsg: Removed

PR #162, opened 2018-05-31
Adding notification sharding using dynamodb

markrichardsg:


common/src/main/scala/models/Notification.scala
```diff
@@ -66,6 +68,25 @@ object BreakingNewsNotification {
...
 
+case class NewsstandShardNotification(
+                                       id: UUID,
+                                       shard:Int
+                                     ) extends Notification {
+  override def title = ""
+  override def message: String = ""
+  override def sender: String = "newsstand-shard"
+  override def importance: Importance = Importance.Minor
+  override def withTopics(topics: Set[Topic]): Notification = this
```
alexduf: do you mind implementing that? It's short and would avoid spending time searching why it doesn't work if we suddenly decide to use it

notification/app/notification/services/azure/NewsstandSender.scala
```diff
@@ -1,19 +1,40 @@
...
+import scala.concurrent.{ExecutionContext, Future}
 
+
+class NewsstandSender(
+                       hubClient: NotificationHubClient,
+                       legacyNewsstandRegistrationConverterConfig:NewsstandShardConfig,
+                       notificationSchedulePersistence: NotificationSchedulePersistenceAsync,
+                       clock: Clock = Clock.systemUTC()
+                     ) {
+  implicit  val executionContext:ExecutionContext = ExecutionContext.Implicits.global
```
alexduf: Given that we're in the context of a play app, I'd rather you get the execution context from play through an implicit parameter at the class level here.

notification/app/notification/services/azure/NewsstandSender.scala
```diff
@@ -1,19 +1,40 @@
...
       tags = Some(Tags.fromTopics(Set(Topic(Newsstand, "newsstand"))))
     )
-    hubClient.sendNotification(push)
+    val shards = legacyNewsstandRegistrationConverterConfig.shards
+    val nowSeconds = Instant.now(clock).getEpochSecond
+
+    hubClient.sendNotification(push).flatMap(hubResult => {
+      Future.sequence{Range(1,shards.toInt + 1).map(shard => notificationSchedulePersistence.writeAsync(NotificationsScheduleEntry(
+        new UUID(id.getMostSignificantBits, id.getLeastSignificantBits+shard).toString,
+        shard.toString,
```
alexduf: I think you may be missing the json serialisation of the notification, or have I missed something myself?

notification/test/notification/services/azure/NewsstandSenderTest.scala
```diff
@@ -1,32 +1,87 @@
...
+          current = queue.poll()
+        }
+        val nowSeconds = instant.getEpochSecond
+        val sevenDaysFromNowSeconds = nowSeconds + 604800
+        val uuidMsb = randomId.getMostSignificantBits
+        val uuidLsb = randomId.getLeastSignificantBits
+        notificationsScheduleEntries must be equalTo Set(
+          NotificationsScheduleEntry(
+            new UUID(uuidMsb, uuidLsb + 1).toString,
+            1.toString,
```
alexduf: Shouldn't that be the serialised json of the notification?

common/src/main/scala/models/Notification.scala
alexduf: Nah after thinking about, don't bother

notification/app/notification/services/azure/NewsstandSender.scala
markrichardsg: Agreed

notification/app/notification/services/azure/NewsstandSender.scala
markrichardsg: Adding that too

notification/test/notification/services/azure/NewsstandSenderTest.scala
markrichardsg: Adding that too

schedulelambda/riff-raff.yaml
```diff
@@ -0,0 +1,10 @@
+stacks: [mobile-notifications]
+regions: [eu-west-1]
+deployments:
+  schedule:
+    type: aws-lambda
+    parameters:
+      bucket: mobile-dist
```
alexduf: `mobile-notifications-dist`

markrichardsg: changed

PR #171, opened 2018-06-18
Convert iOS payload to firebase Payload

alexduf:
This is branch 1/3

This PR only focuses on converting notifications to the ios firebase payload. It doesn't actually send the notification yet.


NathanielBennett:
Looks Good. There isn't going to be a world cup in 3012, though

alexduf:
Ahah thanks Nathaniel :)

PR #173, opened 2018-06-19
Simplify error handling

alexduf:
I can't find a good reason this was looking a Russian Doll so this is in the hope it will make it easier to read


NathanielBennett:
:+1: Thats nice. So nice

PR #180, opened 2018-07-03
(Re)create an unregistration endpoint

alexduf:
We use to have a unregistration enpoint, driven by the browser id. That's been removed [here](https://github.com/guardian/mobile-n10n/pull/177)

This PR re-instates a similar endpoint, relying on the device token instead.

I also renamed the `lastKnownChannelUri` we had in a few places. It used to make sense back when we supported w10 push notifications, but nowadays `pushToken` makes much more sense

### DIFF
--- a/registration/app/registration/services/azure/NotificationsHubRegistrar.scala
+++ b/registration/app/registration/services/azure/NotificationsHubRegistrar.scala
@@ -23,8 +23,8 @@ class NotificationHubRegistrar(
   override val providerIdentifier = "azure"
   val logger = Logger(classOf[NotificationHubRegistrar])
 
-  override def register(lastKnownChannelUri: String, registration: Registration): RegistrarResponse[RegistrationResponse] = {
-    findRegistrationResponses(lastKnownChannelUri).flatMap {
+  override def register(pushToken: String, registration: Registration): RegistrarResponse[RegistrationResponse] = {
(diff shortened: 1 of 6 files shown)


registration/app/registration/services/azure/NotificationsHubRegistrar.scala
```diff
@@ -23,29 +23,37 @@ class NotificationHubRegistrar(
   override val providerIdentifier = "azure"
   val logger = Logger(classOf[NotificationHubRegistrar])
 
-  override def register(lastKnownChannelUri: String, registration: Registration): RegistrarResponse[RegistrationResponse] = {
-    findRegistrationResponses(lastKnownChannelUri).flatMap {
+  override def register(pushToken: String, registration: Registration): RegistrarResponse[RegistrationResponse] = {
+    findRegistrationResponses(pushToken).flatMap {
       case Right(Nil) => createRegistration(registration)
       case Right(azureRegistration :: Nil) => updateRegistration(azureRegistration, registration)
```
NathanielBennett: Is it worth renaming this now?

NathanielBennett:
LG apart from little Q

registration/app/registration/services/azure/NotificationsHubRegistrar.scala
alexduf: Could be separate but it got really itchy :)

PR #181, opened 2018-07-06
Deleting our "Backup" module

alexduf:
Because:
- it wasn't really a backup
- it wasn't working
- we had no easy way to import the data again
- we won't be using Azure for long


NathanielBennett:
Doesnt seem unreasonable

PR #182, opened 2018-07-10
Add udid back for the purpose of querying only

alexduf:
This is cherry picked out of the git history.

We don't store the udid anymore, but if a device doesn't register again we may still have their udid.

This allows us to fetch a registration via the udid and potentially delete it using the token ID

cc @jorgeazevedo


jorgeazevedo:
@alexduf ace!

PR #186, opened 2018-08-06
MSS-369 Implement retry when azure rejects a notification

alexduf:
(and fix our unit test at the same time)

### DIFF
--- a/common/src/main/scala/azure/NotificationHubClient.scala
+++ b/common/src/main/scala/azure/NotificationHubClient.scala
@@ -77,17 +77,29 @@ class NotificationHubClient(val notificationHubConnection: NotificationHubConnec
   }
 
   def sendNotification(rawPush: RawPush): Future[HubResult[Option[String]]] = {
-    val serviceBusTags = rawPush.tagQuery.map(tagQuery => "ServiceBusNotification-Tags" -> tagQuery).toList
-    logger.debug(s"Sending Raw Notification: $rawPush")
-    rawPush.post(
-      request(Endpoints.Messages)
-        .addHttpHeaders(rawPush.extraHeaders: _*)
-        .addHttpHeaders("ServiceBusNotification-Format" -> rawPush.format)
-        .addHttpHeaders(serviceBusTags: _*)
-    ).map {
-      case r if r.isSuccess => Right(r.header("Location"))
-      case r => Left(XmlParser.parseError(r))
+    def retry(count: Int): Future[HubResult[Option[String]]] = {
+      val serviceBusTags = rawPush.tagQuery.map(tagQuery => "ServiceBusNotification-Tags" -> tagQuery).toList
+      logger.debug(s"Sending Raw Notification: $rawPush")
+      rawPush.post(
+        request(Endpoints.Messages)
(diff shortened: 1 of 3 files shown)


common/src/main/scala/azure/NotificationHubClient.scala
```diff
@@ -77,17 +77,29 @@ class NotificationHubClient(val notificationHubConnection: NotificationHubConnec
...
+    def retry(count: Int): Future[HubResult[Option[String]]] = {
+      val serviceBusTags = rawPush.tagQuery.map(tagQuery => "ServiceBusNotification-Tags" -> tagQuery).toList
+      logger.debug(s"Sending Raw Notification: $rawPush")
+      rawPush.post(
+        request(Endpoints.Messages)
+          .addHttpHeaders(rawPush.extraHeaders: _*)
+          .addHttpHeaders("ServiceBusNotification-Format" -> rawPush.format)
+          .addHttpHeaders(serviceBusTags: _*)
+      ).flatMap {
+        case r if r.isSuccess =>
```
markrichardsg: I find `r` hard to read as in request/response environment it can mean either. Is `resp` better?

common/src/test/scala/azure/NotificationHubClientSpec.scala
```diff
@@ -1,31 +1,95 @@
...
+      }} { client =>
+        client.sendNotification(rawPush)
+      }
+
+      result should beLeft
+    }
+
+    "try to send a notification twice if the first try fails" in {
+      val rawPush = GCMRawPush(GCMBody(data = Map.empty), None)
+      var tryCount = 0
```
markrichardsg: Is this threadsafe? AtomicInteger might be safer?

common/src/test/scala/azure/NotificationHubClientSpec.scala
```diff
@@ -1,31 +1,95 @@
 package azure
 
 import org.specs2.concurrent.ExecutionEnv
+import org.specs2.mock.Mockito
 import org.specs2.mutable.Specification
-import play.api.Play
+import play.api.BuiltInComponents
+import play.api.libs.ws.ahc.AhcWSResponse
```
markrichardsg: Not sure all these imports are used

common/src/main/scala/azure/NotificationHubClient.scala
```diff
@@ -77,17 +77,29 @@ class NotificationHubClient(val notificationHubConnection: NotificationHubConnec
...
+          if (count < 3 && error.reason.contains("Azure Notifications Hub Server Busy")) {
+            logger.error(s"Unable to push notification, try $count/3. Got response code: ${r.status} got the following error: $error for the notification $rawPush")
+            retry(count + 1)
+          } else {
+            logger.error(s"Unable to push notification, response code: ${r.status} got the following error: $error for the notification $rawPush")
+            Future.successful(Left(error))
+          }
+      }
     }
+    retry(1)
```
NathanielBennett: Just to be clear, we retry twice? Maybe its a little clearer if this starts as 0 and you do `count < 2` on L92

PR #187, opened 2018-08-06
Implement server side FCM registration

alexduf:
After countless rebases I wanted to split this PR in smaller chunks, alas I kept failing to make it more readable.

Apologies for the 700 lines. Here's what this PR does:

 - A bit of cleanup for the first few commits in order to prepare the work, such as removing dead code, introducing more sensible typing etc
 - Implement a Registrar via FCM
 - Implement a Registrar that takes two registrars, and migrated registration between one and the other automatically
 - Implement a RegistrarProvider that also handles the migration, such that depending on the registration request (containing only an azure token, only a firebase token, or both) will either delegate to the right registrar (only one token) or to the migrating registrar (two tokens)
 - Fixes a couple of issues with the android payload

Although the commits aren't in an ideal order I would still recommend to read this PR commit per commit


NathanielBennett:
There's a few more places you can change `deviceId` to `deviceToken`
https://github.com/guardian/mobile-n10n/blob/MSS-335-rebased-3/common/src/main/scala/azure/RawAPNSRegistration.scala
https://github.com/guardian/mobile-n10n/blob/MSS-335-rebased-3/common/src/main/scala/azure/Responses.scala
https://github.com/guardian/mobile-n10n/blob/MSS-335-rebased-3/registration/app/registration/services/azure/NotificationsHubRegistrar.scala
https://github.com/guardian/mobile-n10n/blob/MSS-335-rebased-3/registration/app/registration/services/NotificationRegistrar.scala

notification/app/notification/services/azure/GCMPushConverter.scala
```diff
@@ -157,7 +157,7 @@ class GCMPushConverter(conf: Configuration) extends AzurePushConverter {
       "awayTeamScore" -> Some(matchStatusAlert.awayTeamScore.toString),
       "awayTeamText" -> Some(matchStatusAlert.awayTeamMessage),
       "currentMinute" -> Some(""),
-      "important" -> Some(matchStatusAlert.importance.toString),
+      "importance" -> Some(matchStatusAlert.importance.toString),
```
markrichardsg: Are there any side effects to changing important to importance?

registration/app/registration/services/fcm/FcmRegistrar.scala
```diff
@@ -0,0 +1,160 @@
...
+
+    def listExistingTopics: RegistrarResponse[Set[Topic]] = {
+      fetchInstance(deviceToken).map {
+        case Right(instance) => Right(instance.topics.toSet)
+        case Left(InstanceNotFound) => Right(Set())
+        case Left(error) => Left(error)
+      }
+    }
+
+    val result = for {
```
markrichardsg: Is this safe if the server restarts?

registration/app/registration/services/MigratingRegistrar.scala
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  override def findRegistrations(topic: Topic, cursor: Option[String]): RegistrarResponse[Paginated[StoredRegistration]] =
+    legacyRegistrar.findRegistrations(topic, cursor)
+
+  override def findRegistrations(deviceToken: DeviceToken): RegistrarResponse[List[StoredRegistration]] = {
+    val registrations = for {
+      fcmRegistrations <- EitherT(fcmRegistrar.findRegistrations(deviceToken))
+      legacyRegistrations <- EitherT(legacyRegistrar.findRegistrations(deviceToken))
+    } yield fcmRegistrations ++ legacyRegistrations
```
markrichardsg: Could or should this react to a difference between the two?

notification/app/notification/services/azure/GCMPushConverter.scala
alexduf: Yes, it will fix the android side as it's currently not parsing it

common/src/main/scala/binders/querystringbinders/package.scala
```diff
@@ -104,24 +104,32 @@ package object querystringbinders {
...
-          for {
-            platform <- eitherPlatform.right
-            deviceToken <- eitherDeviceToken.right
-          } yield {
-            RegistrationsByDeviceToken(platform, deviceToken)
-          }
-
-        case (Some(_), None) => Left("Missing parameter: deviceToken")
-        case (None, Some(_)) => Left("Missing parameter: platform")
+      val optEitherDeviceToken = QueryStringBindable.bindableString.bind("azureToken", params)
```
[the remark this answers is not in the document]
alexduf: Good shout, fixing

common/src/main/scala/binders/querystringbinders/package.scala
```diff
@@ -104,24 +104,32 @@ package object querystringbinders {
...
+      val optEitherFirebaseToken = QueryStringBindable.bindableString.bind("firebaseToken", params)
+
+      condOpt((optEitherPlatform, optEitherDeviceToken, optEitherFirebaseToken)) {
+        case (Some(Right(platform)), Some(Right(azureToken)), Some(Right(fcmToken))) =>
```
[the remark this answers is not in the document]
alexduf: Fixing!

common/src/main/scala/binders/querystringbinders/package.scala
```diff
@@ -104,24 +104,32 @@ package object querystringbinders {
   implicit def qsbindableRegistrationsByDeviceToken: QueryStringBindable[RegistrationsByDeviceToken] = new QueryStringBindable[RegistrationsByDeviceToken] {
```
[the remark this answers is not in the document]
alexduf: I know it's not ideal but i think it's the best we have with Play: https://www.playframework.com/documentation/2.6.x/ScalaRequestBinders

registration/app/registration/services/fcm/FcmRegistrar.scala
alexduf: Connections are drained before the server is shut down so on that aspect it should be fine.

Even if the JVM explodes mid air, or if something else happen, the user might end up in an unstable position, which will be fix next time they try to register again (triggered every two weeks client side).

But assuming the server and the load balancer do behave correctly, it should return a 500, which will prompt the client to retry immediately up to three times.

registration/app/registration/services/MigratingRegistrar.scala
alexduf: Not really, this is more to check the data manually, that endpoint isn't called by anything automated

PR #188, opened 2018-08-08
MSS-313: Adding mobile-notifications-api-client project as a sub module

markrichardsg:
Moving mobile-notifications-api-client into mobile-n10n a sub project
Only slight concern is whether release still works.
` sbt "project api-client" release` does start the release process, but not completed it yet.


alexduf:
It looks good, I can try to release it as I have the whole thing already setup. If it works I'll approve :)

PR #189, opened 2018-08-09
Disable the frontend notification

alexduf:
This used to be used to push notification to the website. The website would then use that event to show a banner at the bottom of the page.
Since Garnett that feature has been unmaintained and broke at some point. We have been getting 503s in our logs for months and dotcom isn't quite sure if they will fix it or not.

So for now I'm disabling it until further notice


alexduf:
Tested on CODE :heavy_check_mark:

PR #195, opened 2018-08-30
Mss 426/event lambda

markrichardsg:


NathanielBennett:
:+1: Think this looks good apart from cfn comments

PR #196, opened 2018-08-31
Send token metrics

alexduf:
Send metrics about the nature of the registration token. Three cases:
 - The registration comes with an Azure token
 - The registration comes with a Firebase token
 - The registration comes with both

The cloudwatch logic is imported from Mapi. I was considering making a library but I'm actually not sure how relevant that would be for the rest of the department (or even ourselves)


NathanielBennett:
Looks Good. Lots of the code is identical or similar to MAPI right. How easy would that be to extract? (I didn't look, soz!)

PR #206, opened 2018-09-13
Improve logging and allow batching if wanted

markrichardsg:
General improvements.
Tidy up (remove old lambda)
If config updates to batch sqs messages; it should now support that.
Improved logging.
Moved code about to  make it easier to reason.


eventconsumer/src/main/scala/com/gu/notifications/events/SqsLambda.scala
```diff
@@ -56,4 +67,13 @@ class SqsLambda extends RequestStreamHandler {
     }
   }
 
+  private def processSqsEvent(event: SqsEvent) = event.Records.foldLeft(Future.successful[SqsAttemptCount](SqsAttemptCount(NotificationCount(0, 0), 0, 0))) {
```
alexduf: I'm having a hard time to understand what this function is doing.
 - If it's a function with side effect, make it clear by adding the return type `Unit` otherwise a return type would help me a lot here
 - Because the definition of `SqsAttemptCount`, it's hard to understand what the counts are. Either define `SqsAttemptCount.empty` or use named parameters
 - I'm wondering if there wouldn't be a way to split the processing from the counting as it seems like a lot is happening in the same place

eventconsumer/src/main/scala/com/gu/notifications/events/Router.scala
```diff
@@ -0,0 +1,44 @@
...
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.Try
+
+class Router(eventConsumer: S3EventProcessor, reportUpdater: DynamoReportUpdater)(implicit executionContext: ExecutionContext) {
+  private val logger: Logger = LogManager.getLogger(classOf[Router])
+
+  def sqsEventRoute(inputString: String): Unit = {
+    val s3Events: Seq[S3Event] = SqsEventReader.readSqsEventString(inputString)
+    val s3EventAggregateProviders = s3Events.map(s3Event => () => s3EventRoute(s3Event))
+    val eventualTriedNotificationCounts = callProvidersInSeries(s3EventAggregateProviders)
```
alexduf: reading `callProvidersInSeries` I don't see why there's any Future involved here as the processing is called in series, and the futures are all Future.successful.

If you want to trigger the computation in parallel I'd recommend having something like that
``` scala
val aggregates = s3Events.map(s3Event => () => s3EventRoute(s3Event))
Future.sequence(aggregates)
```

eventconsumer/src/main/scala/com/gu/notifications/events/model/AggregationCounts.scala
```diff
@@ -0,0 +1,42 @@
...
+case class AggregationCounts(success: Int, failure: Int) {
+  def combine(otherAttemptedCount: AggregationCounts): AggregationCounts = AggregationCounts(success + otherAttemptedCount.success, failure + otherAttemptedCount.failure)
+}
+
+object AggregationCounts {
+  private val logger = LogManager.getLogger(classOf[AggregationCounts])
+  implicit val jf = Json.format[AggregationCounts]
+
+  def aggregate(triedAggregationResults: Seq[Try[AggregationCounts]]): S3ResultCounts = {
+    triedAggregationResults.foldLeft(S3ResultCounts(0, 0, AggregationCounts(0, 0))) {
```
alexduf: You could flatten that pattern match:

```scala
triedAggregationResults.foldLeft(S3ResultCounts(0, 0, AggregationCounts(0, 0))) {
  case (s3Result, Success(aggregationResult @ AggregationCounts(_, 0)) => 
    s3Result.copy(success = s3Result.success + 1, aggregationResults = aggregationResult.combine(s3Result.aggregationResults))
  case (s3Result, Success(aggregationResult) => 
    s3Result.copy(failure = s3Result.failure + 1, aggregationResults = aggregationResult.combine(s3Result.aggregationResults))
  case (s3Result, Failure(NonFatal(exception))) =>
    logger.warn("Error processing an s3 event", exception)
    s3Result.copy(failure = s3Result.failure + 1)
}
```

eventconsumer/src/main/scala/com/gu/notifications/events/Router.scala
alexduf: As discussed, not a priority, doesn't require to be fixed right now

PR #209, opened 2018-09-19
Better notification reports

alexduf:
The report.notifications.guardianapis.com/notifications? endpoint now lists counts without timmings, but the detailed endpoint lists all the metrics

### DIFF
--- a/report/app/report/models/ExtendedNotificationReport.scala
+++ b/report/app/report/models/ExtendedNotificationReport.scala
@@ -1,13 +1,13 @@
-package models
+package report.models
 
 import java.time.Instant.ofEpochMilli
 import java.time.LocalDateTime.ofInstant
 import java.time.ZoneOffset.UTC
-import java.time.{Instant, LocalDateTime, ZoneOffset}
 import java.util.UUID
 
 import azure.NotificationDetails
-import com.gu.notifications.events.model.{DynamoEventAggregation, EventAggregation}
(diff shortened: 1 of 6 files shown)


report/app/report/models/ExtendedNotificationReport.scala
```diff
@@ -0,0 +1,105 @@
...
+object ExtendedEventAggregation {
+  implicit val extendedEventAggregationJF: Format[ExtendedEventAggregation] = Json.format[ExtendedEventAggregation]
+
+  def full(dynamoEventAggregation: DynamoEventAggregation, originalSentTime: LocalDateTime): ExtendedEventAggregation = {
+    def toOrderedTiming(timing: List[List[Int]]): Option[List[(LocalDateTime, Int)]] = {
+      val sentTime = originalSentTime.truncatedTo(TenSecondUnit)
+      val (_, orderedTiming) = dynamoEventAggregation.timing.map(timed => (timed.head, timed(1))).foldLeft((sentTime, List[(LocalDateTime, Int)]())) {
+        case ((lastTime, newList), (offset, count)) =>
+          val nextTime = lastTime.plusSeconds(10 * offset)
+          (nextTime, newList :+ (nextTime, count))
```
markrichardsg: This should probably use prepend like:
```((nextTime, count) :: nextTime, newList)```
finally reverse if needed.
There's a big performance penalty when using `:+`

PR #211, opened 2018-09-19
MSS-487: Efficient timing conversion

markrichardsg:
Converting the timings was taking upto 7 seconds!
It seems this is because of the repeated rebuilding of the list, so amended to use a mutabe list (ListBuffer) which has a much smaller overhead to build as it doens't have to be rebuilt.


markrichardsg:
@davidfurey swapped to prepend, forgot that was constant time

PR #214, opened 2018-10-02
Stabilise android notifications

alexduf:
Add logic to rollback the android migration to a hybrid solution: Azure + Firebase


registration/app/registration/models/LegacyRegistration.scala
```diff
@@ -3,7 +3,7 @@ package registration.models
 import play.api.libs.json._
 import play.api.libs.json.JodaReads._
 import LegacyJodaFormat._
-import models.Provider
+import models.{Provider, RegistrationProvider}
```
NathanielBennett: Do you actually still use ```Provider```?

PR #216, opened 2018-10-15
Report: Add timing percentiles

TBonnin:
Notification report endpoins now returns the timing percentiles rather than a map of all timings 10secs slots

screen shot 2018-10-15 at 11 11 01

### DIFF
--- a/commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
+++ b/commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
@@ -8,6 +8,7 @@ import com.gu.notifications.events.utils.Percentiles
 import play.api.libs.json.{JsValue, Json, OFormat, Writes}
 import com.gu.notifications.events.utils.OWriteOps._
 
+// Percentiles are durations since the first received timing
 case class TimingPercentiles(
   `10th`: LocalDateTime,
   `20th`: LocalDateTime,
@@ -23,7 +24,7 @@ case class TimingPercentiles(
 )
 
 object TimingPercentiles {
-  implicit val jf = Json.format[TimingPercentiles]
+  implicit val jf = Json.writes[TimingPercentiles]
 }
 
 case class EventAggregation(
@@ -33,16 +34,11 @@ case class EventAggregation(
 ) {
 
   def timingPercentiles: Option[TimingPercentiles] = {
-    val allTimings: Seq[LocalDateTime] = timing
-      .toSeq
-      .flatMap { case (t, count) =>
-        Seq.fill(count)(t)
-      }
-      .view
+    implicit val dateOrdering: Ordering[LocalDateTime] = _ compareTo _
     Seq(10, 20, 30, 40, 50, 60, 70, 80, 90, 95, 99)
       .flatMap { p =>
         Percentiles
-          .percentile(p)(allTimings)(_ compareTo _)
+          .percentileBuckets(p)(timing)
           .toOption
       } match {
       case Seq(d10, d20, d30, d40, d50, d60, d70, d80, d90, d95, d99) =>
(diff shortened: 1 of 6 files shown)


commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
```diff
@@ -4,13 +4,54 @@ import java.time.temporal.{ChronoUnit, Temporal, TemporalUnit}
...
   providerCounts: ProviderCount,
   timing: Map[LocalDateTime, Int]
-)
+) {
+
+  def timingPercentiles: Option[TimingPercentiles] = {
+    val allTimings: Seq[LocalDateTime] = timing
+      .toSeq
+      .flatMap { case (t, count) =>
+        Seq.fill(count)(t)
```
TBonnin: This is quite heavy memory-wise. @alexduf Can you think of a better way to calculate percentiles?

commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
```diff
@@ -79,6 +120,12 @@ object EventAggregation {
...
     (dateTimeMap: Map[LocalDateTime, Int]) => Json.toJsObject(dateTimeMap.map { case (k, v) => (k.toString, v) })
   )
-  implicit val jf = Json.format[EventAggregation]
+
+  implicit val jreads = Json.reads[EventAggregation]
+  implicit val jwrites: Writes[EventAggregation] =
+    Json
+      .writes[EventAggregation]
+      .addField("timingPercentiles", _.timingPercentiles)
+      .removeField("timing")
```
TBonnin: I made the decision to hide the timing map. It is super long and doesn't provide much information as is.

commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
alexduf: Yes, I think we have count of the total notification received.
If we take the timings and sort it, then iterate through it, we can stop when we have the total number of notification so far > than the percentile you're calculating. Something like that (definitely not working scala, just to get the logic)

```scala
val mySortedTimings: Seq[(LocalDateTime, Int)] = timing.toSeq.sort
val `80thPercentileCount` = total * 0.8
val (`80thPercentileTime`, _) = mySortedTimings.foldLeft((notificationSentTime, 0)) {
  case ((_, totalSoFar), (time, count)) if totalSoFar < `80thPercentileCount` => (time, count + totalSoFar)
  case ((time, totalSoFar), _) if totalSoFar >= `80thPercentileCount` => (time, totalSoFar)
}
```

commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
alexduf: No I agree! we just need to double check if that has an impact on the event consumer that stores the data in dynamo

alexduf:
Also, should the percentile be relative to the sent time?

commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
alexduf: I double-checked, it's okay to hide it like that :+1:

TBonnin:
@alexduf I have refactored the percentiles calculations. See https://github.com/guardian/mobile-n10n/pull/216/files#diff-9d2d9b7cfea3578513d3888e49c5feb6R24
It requires to iterate through the map more than once but it doesn't create a gigantic list of all values. (tl;dr: less efficient computatin-wise but more efficient memory-wise)

As discussed on the chat, relative duration could be good but will require some more refactoring to store the notification sending time. Let's do that in another PR if we realise we really need it

PR #217, opened 2018-10-16
Add SubscriptionService to query database

TBonnin:
First pass at creating a database layer to access the registration database.

@Alex can you send me the exact schema, I will change the class/fields/queries accordingly

How to use it:
```
val prog: IO[Int] = for {
    xa <- DatabaseConfig.transactor[IO](jdbcConfig)
    service = SubscriptionService(xa)
    rowsAffected <- service.save(Subscription(...))
  } yield rowsAffected

  prog.unsafeRunSync()
```


TBonnin:
Hi @alexduf 
I have added a few commits since last night (LocalDateTime for registration date type, save does an upsert, using existing Platform model). 
I think this is ready to merge. I might have missed something but we can always add it when we need it

alexduf:
closed in favour of https://github.com/guardian/mobile-n10n/pull/219

PR #218, opened 2018-10-17
Copying registrar

alexduf:
_This doesn't affect the production flow yet, it's just additional code that isn't wired in._

I'm adding another layer of abstraction such that we can keep all the existing logic and wrap it in a `CopyingRegistrar`, that will replicate every operation on two data stores (one being our database)

### DIFF
--- a/registration/app/registration/services/CopyingRegistrar.scala
+++ b/registration/app/registration/services/CopyingRegistrar.scala
@@ -11,8 +11,8 @@ import scala.util.{Failure, Success}
 
 class CopyingRegistrar(
   val providerIdentifier: String,
-  mainRegistrar: NotificationRegistrar,
-  copyRegistrar: NotificationRegistrar
+  val mainRegistrar: NotificationRegistrar,
+  val copyRegistrar: NotificationRegistrar
 )(
   implicit ec: ExecutionContext
 ) extends NotificationRegistrar {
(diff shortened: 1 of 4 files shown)


registration/app/registration/services/CopyingRegistrar.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  override def findRegistrations(deviceToken: DeviceToken): RegistrarResponse[List[StoredRegistration]] = {
+    val mainResponseF = mainRegistrar.findRegistrations(deviceToken)
+    val copyResponseF = copyRegistrar.findRegistrations(deviceToken)
+
+    for {
+      mainResponse <- EitherT(mainResponseF).getOrElse(Nil)
+      copyResponse <- EitherT(copyResponseF).getOrElse(Nil)
+    } yield Right(mainResponse ++ copyResponse)
```
alexduf: This may look odd but wrapping it in EitherT allows me to swallows errors and return Nil instead, which is what we want for this method. This is only called by us (dev) to check where a registration is

TBonnin: You could use `Future.sequence` and get the rights but I am not sure this is clearer
```suggestion
    Future
      .sequence(List(mainResponseF, copyResponseF))
      .map(eithers => Right(eithers.partition(_.isRight)._1.flatMap(_.right.get)))
```

alexduf: The problem with that is in case of an exception both requests will fail. EitherT wraps both cases and only maps on the happy case which is handy

alexduf:
superseded by https://github.com/guardian/mobile-n10n/pull/219

PR #223, opened 2018-10-23
Add APNS worker 

TBonnin:
How to use it:
```
val repo: RegistrationRepository = ???
val registrationService: RegistrationService[IO, Stream] = new RegistrationService(repo)
val apns = new Apns(registrationService, config)

val notification: Notification = ???
val shards: Seq[Shard] = ???
def report: Pipe[IO, Either[Throwable, String], Unit] = ???

apns
    .send(notification, shards)
    .through(report)
    .compile
    .drain
    .unsafeRunSync()
```

TODO:
- [x] Set all notification payload parameters
- [x] Add dryrun mode


apnsworker/src/main/scala/apnsworker/ApnsClient.scala
```diff
@@ -0,0 +1,104 @@
...
+    //      payloadBuilder.setLocalizedAlertMessage()
+    //      payloadBuilder.setLocalizedAlertSubtitle()
+    //      payloadBuilder.setLocalizedAlertTitle()
+    //      payloadBuilder.setMutableContent()
+    //      payloadBuilder.setPreferStringRepresentationForAlerts()
+    //      payloadBuilder.setShowActionButton()
+    //      payloadBuilder.setSound()
+    //      payloadBuilder.setThreadId()
+    //      payloadBuilder.setUrlArguments()
+    //      payloadBuilder.addCustomProperty()
```
TBonnin: @alexduf That's the part I need help to make sure I don't forget to set any parameters and properly translate our notification model to the Pushy payload

apnsworker/src/main/scala/apnsworker/Apns.scala
```diff
@@ -0,0 +1,42 @@
...
+  def send(notification: Notification, shardRange: ShardRange): Stream[F, Either[Throwable, String]] = {
+
+    def sendAsync(token: String)(client: PushyApnsClient): F[String] =
+      Async[F].async[String] { (cb: Either[Throwable, String] => Unit) =>
+        ApnsClient.sendNotification(token, notification)(cb)(client, config)
+      }
+
+    def tokens: Stream[F, String] = notification
+        .topic
+        .map(topic => registrationService.find(Topic(topic.name), iOS, shardRange))
```
alexduf: You're probably aware of it as this is a WIP, but we'll need to support either a sql request that de-duplicates when querying multiple topics, or an in-memory de-duplication.

Content notification and football notification have generally more than one topic at a time and if I understand the code correctly a single token might come up multiple times.

apnsworker/src/main/scala/apnsworker/Apns.scala
```diff
@@ -0,0 +1,42 @@
...
+        .reduce(_ merge _)
+        .map(_.device.token)
+
+    for {
+      client <- Stream.eval(apnsClient)
+      res <- tokens
+        .map(token => Stream.eval(sendAsync(token)(client)).attempt)
+        .parJoinUnbounded
```
alexduf: what's `parJoinUnbounded` doing?

apnsworker/src/main/scala/apnsworker/Apns.scala
TBonnin: Yes I have a `TODO` to let the db handle that. Merging stream is not adequate here

apnsworker/src/main/scala/apnsworker/Apns.scala
TBonnin: We have a `Stream` of `Stream` (`Stream[Stream[ResponseEither]]`). The outer stream emitting `tokens` and the inner stream emiting the `send response` for a given token.
`parJoinUnbounded` merge the `Stream` of `Stream` into a single `Stream` (`Stream[ResponseEither]`) running all inner `Streams` at the same time.
I am using `parJoinUnbounded` rather than `parJoin(maxConcurrenty: Int)` because `Pushy` handles for us the limitation of in-flight requests.

alexduf: got it :+1:

apnsworker/src/test/scala/apnsworker/payload/ApnsPayloadSpec.scala
```diff
@@ -0,0 +1,270 @@
...
+    val expected = Some(
+      """
+        |{
+        |   "aps":{
+        |      "content-available":1
+        |   }
+        |}""".stripMargin
+    )
+  }
+}
```
TBonnin: Those tests are a bit meaningless for now because I had no real source of truth for the expected json. However it would be very useful if/when changing the `APNSPayload` logic

apnsworker/src/main/scala/apnsworker/Apns.scala
```diff
@@ -0,0 +1,75 @@
...
+    } yield res
+
+    def sending(token: Token, payload: ApnsPayload, apnsClient: PushyApnsClient): Stream[F, ApnsResponse] = {
+
+      def toApnsResponse(token: Token): Pipe[F, Token, ApnsResponse] =
+        _.attempt.map {
+          _.leftMap {
+            _ match {
+              case ae: ApnsException => ae
+              case e: Throwable => ApnsGenericFailure(notification.id, token, e)
```
alexduf: You probably want to use `case NonFatal(e) =>` so that you don't swallow fatal exceptions such as OutOfMemoryError and ClassNotFound

apnsworker/src/main/scala/apnsworker/models/ApnsException.scala
```diff
@@ -0,0 +1,58 @@
+package apnsworker.models
+
+import java.time.LocalDateTime
+import java.util.UUID
+
+
+sealed trait ApnsException extends RuntimeException {
```
alexduf: I didn't actually find a place where you'd throw the `ApnsException`. Assuming it's not thrown, should that be renamed `ApnsError` and should it stop inheriting from `RuntimeException`?

apnsworker/src/main/scala/apnsworker/models/payload/ApnsPayload.scala
```diff
@@ -0,0 +1,157 @@
...
+      categoryName = Option(n.link match {
+        case _: Link.External => ""
+        case _: Link.Internal => "ITEM_CATEGORY"
+      }),
+      contentAvailable = true,
+      mutableContent = imageUrl.isDefined,
+      sound = Some("default"),
+      customProperties = Seq(
+        CustomProperty(Keys.UniqueIdentifier -> n.id.toString),
+        CustomProperty(Keys.UniqueIdentifier -> n.id.toString),
```
alexduf: duplicate line

apnsworker/src/main/scala/apnsworker/models/ApnsException.scala
TBonnin: Yes you are right 👍 . I change it and keep it inheriting from RuntimException

PR #224, opened 2018-10-23
Write notifications to sqs

alexduf:
This code isn't wired so it won't have any impact in production when merged.

I created a class `GuardianNotificationSender` that will write units of work (notification + range) to one SQS queue for one platform.
When we'll wire that class we'll have to create three instances of it: iOS, Android and Newsstand. Both the newsstand and the iOS will share the same SQS queue as the same worker will pick up the work.

I've also ensured the generated ranges are inclusive on both ends [1, 250], [251, 500] such that we can directly use these values in a `between` sql statement.

The payload is of the following format:

```json
{
  "notification": {...},
  "range": {"start": 1, "end": 250}
}
```

where notification is a serialised `models.Notification`

### DIFF
--- a/notification/app/notification/services/guardian/GuardianNotificationSender.scala
+++ b/notification/app/notification/services/guardian/GuardianNotificationSender.scala
@@ -26,17 +26,18 @@ class GuardianNotificationSender(
   sqsArn: String,
 )(implicit ec: ExecutionContext) extends NotificationSender {
 
-  val BATCH_SIZE: Int = 10000
+  val WORKER_BATCH_SIZE: Int = 10000
+  val SQS_BATCH_SIZE: Int = 200 // max 256kb per sqs call, 1 notification ~< 1kb
 
   private val logger: Logger = Logger.apply(classOf[GuardianNotificationSender])
 
   override def sendNotification(push: Push): Future[SenderResult] = for {
       topicStats <- registrationCounter.count(push.notification.topic)
       registrationCount = topicStats.counts.getOrElse(platform, 10)
-      batch = prepareBatch(platform, push.notification, registrationCount)
-      batchResult <- sendBatch(batch)
+      workerBatches = prepareBatch(platform, push.notification, registrationCount)
+      sqsBatchResults <- sendBatch(workerBatches)
   } yield {
-    val failed = Option(batchResult.getFailed).map(_.asScala.toList).getOrElse(Nil)
+    val failed = sqsBatchResults.flatMap(response => Option(response.getFailed).map(_.asScala.toList).getOrElse(Nil))
     if (failed.isEmpty) {
       Right(SenderReport(
         Guardian.value,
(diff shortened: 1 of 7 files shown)


notification/app/notification/services/guardian/ShardedNotification.scala
```diff
@@ -0,0 +1,10 @@
+package notification.services.guardian
+
+import models.Notification
+
+case class ShardRange(start: Short, end: Short)
```
TBonnin: We would need to move this `ShardRange` model into `common` so it can be used by the `db` logic and the `workers`

notification/app/notification/services/guardian/GuardianNotificationSender.scala
```diff
@@ -0,0 +1,92 @@
...
+)(implicit ec: ExecutionContext) extends NotificationSender {
+
+  val WORKER_BATCH_SIZE: Int = 10000
+  val SQS_BATCH_SIZE: Int = 200 // max 256kb per sqs call, 1 notification ~< 1kb
+
+  private val logger: Logger = Logger.apply(classOf[GuardianNotificationSender])
+
+  override def sendNotification(push: Push): Future[SenderResult] = for {
+      topicStats <- registrationCounter.count(push.notification.topic)
+      registrationCount = topicStats.counts.getOrElse(platform, 10)
```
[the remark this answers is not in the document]
TBonnin: Also let's add a comment to say that we want to proceed anyway (even with a single range) in case `topicStats` fails

notification/app/notification/services/guardian/GuardianNotificationSender.scala
```diff
@@ -60,6 +60,18 @@ class GuardianNotificationSender(
...
+    registrationCounter.count(topics).map(t => t.counts(platform)).recover {
+      case NonFatal(e) if topics.exists(_.`type` == TopicTypes.Breaking) && platform != Newsstand =>
+        logger.error("Unable to count registration for a list of topics during a breaking news", e)
+        1500000 // fallback on 1.5M: the worse case
+      case NonFatal(e) =>
+        logger.error("Unable to count registration for a list of topics", e)
+        1
+    }
+  }
+
```
TBonnin: +1 for the logic but it is related to the number of shard we want to create and therefore rather than having it decoupled from the shard count I would return a `Future[Option[Int]]` here and have the logic in `prepareBatch`:
```
def prepareBatch(platform: Platform, notification: Notification, registrationCount: Option[Int]) = {
     ...
     val count = registrationCount match {
           case None if notification.topics.exists(_.`type` == TopicTypes.Breaking) && platform != Newsstand =>
                logger.error("Unable to count registration for a list of topics during a breaking news", e)
                1500000 // if breaking news we assume everybody is subscrided to it
           case None =>
                logger.error("Unable to count registration for a list of topics", e)
                1
      case Some(c) => c
   }
   shard(count).map { shard =>
   ...
```

TBonnin:
:+1:

PR #225, opened 2018-10-25
Count registration per topic(s) and platform

alexduf:
The endpoint returns a json:
```json
{
  "total": 20,
  "ios": 15,
  "android": 5,
  "newsstand": 0
}
```

The idea of counting per platform is to be able to cache the request on the client side such that for one breaking news we only count once, and use the result twice.

### DIFF
--- a/common/src/main/scala/db/RegistrationService.scala
+++ b/common/src/main/scala/db/RegistrationService.scala
@@ -1,5 +1,6 @@
 package db
 
+import cats.data.NonEmptyList
 import cats.effect.internals.IOContextShift
 import cats.effect.{Async, ContextShift, IO}
 import doobie.util.transactor.Transactor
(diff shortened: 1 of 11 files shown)


report/app/report/controllers/RegistrationCount.scala
```diff
@@ -0,0 +1,31 @@
...
+  controllerComponents: ControllerComponents,
+  authAction: AuthAction
+) extends AbstractController(controllerComponents) {
+
+  def forTopics(topics: List[Topic]): Action[AnyContent] = authAction.async { request =>
+    topics match {
+      case Nil => Future.successful(BadRequest("Couldn't find any valid topic"))
+      case firstTopic :: moreTopics =>
+        val nonEmptyTopics = NonEmptyList(firstTopic, moreTopics)
+        registrationService.countPerPlatformForTopics(nonEmptyTopics.map(topic => db.Topic(topic.toString)))
```
TBonnin: would it be safer to use `topic.name` rather than `topic.toString`?

common/src/main/scala/db/RegistrationService.scala
```diff
@@ -1,20 +1,27 @@
...
 import doobie.util.transactor.Transactor
 import fs2.Stream
+import models.PlatformCount
 import play.api.Configuration
 import play.api.inject.ApplicationLifecycle
 
 import scala.concurrent.ExecutionContext
 
-class RegistrationService[F[_], S[_[_], _]](repository: RegistrationRepository[F, S]) {
+class RegistrationService[F[_]: Async, S[_[_], _]](repository: RegistrationRepository[F, S]) {
```
TBonnin: why do you need the `Async`?

report/app/report/controllers/RegistrationCount.scala
alexduf: Yes! Changed it from models.Topic to db.Topic and that is definitely safer. I'll change

common/src/main/scala/db/RegistrationService.scala
alexduf: That's a mistake, I'll remove that

common/src/main/scala/db/SqlRegistrationRepository.scala
```diff
@@ -73,4 +76,46 @@ class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
...
+        coalesce(sum(case platform when 'newsstand' then 1 else 0 end), 0) as newsstand
+      FROM
+        registrations
+      WHERE
+        topic = $topic
+      """
+      .query[PlatformCount]
+      .unique
+      .transact(xa)
+  }
```
TBonnin: As discussed face to face, this is a nice optimization but it also means maintaining to methods/queries. 
Let's start with a single generic one `countPerPlatformForTopics` and see if the cost worth adding more code later

PR #230, opened 2018-10-30
apns payload: add missing 'uri' custom key for football status payload

TBonnin:
This great document by @jorgeazevedo highlights a missing custom property in the football status payload.


TBonnin:
:oups: forgot

PR #233, opened 2018-11-01
notificationworker: Adding DLQ and set msg retention period

TBonnin:


notificationworkerlambda/cfn.yaml
```diff
@@ -37,10 +37,18 @@ Parameters:
...
+    Type: AWS::SQS::Queue
+
   Sqs:
     Type: AWS::SQS::Queue
     Properties:
       VisibilityTimeout: 300
+      MessageRetentionPeriod: 3600 # 1 hour
+      RedrivePolicy:
+        deadLetterTargetArn: !GetAtt Dlq.Arn
+        maxReceiveCount: !Ref 1 # ie: no retry
```
alexduf: Why `!Ref 1` instead of 1 ?

TBonnin: because I am stupid ;)

PR #234, opened 2018-11-01
Wire notification to sqs

alexduf:
**This should not be merged until** `create-workers-resources` **is merged**

This wires the code that writes to the SQS queues, and also implements a flip/flop filter that decides where to send the notification

### DIFF
--- a/notification/app/notification/NotificationApplicationLoader.scala
+++ b/notification/app/notification/NotificationApplicationLoader.scala
@@ -11,16 +11,19 @@ import azure.NotificationHubClient
 import com.softwaremill.macwire._
 import controllers.Main
 import _root_.models.NewsstandShardConfig
+import com.amazonaws.services.sqs.{AmazonSQSAsync, AmazonSQSAsyncClientBuilder}
 import com.google.auth.oauth2.GoogleCredentials
 import com.google.firebase.messaging.FirebaseMessaging
 import com.google.firebase.{FirebaseApp, FirebaseOptions}
 import com.gu.AppIdentity
 import com.gu.notificationschedule.dynamo.NotificationSchedulePersistenceImpl
+import _root_.models.{Android, Newsstand, iOS}
 import notification.authentication.NotificationAuthAction
 import notification.services.frontend.{FrontendAlerts, FrontendAlertsConfig}
 import notification.services._
 import notification.services.azure._
 import notification.services.fcm.{APNSConfigConverter, AndroidConfigConverter, FCMNotificationSender}
+import notification.services.guardian.{GuardianNotificationSender, ReportTopicRegistrationCounter, TopicRegistrationCounter}
 import play.api.libs.ws.ahc.AhcWSComponents
 import play.api.routing.Router
 import play.api.{BuiltInComponents, BuiltInComponentsFromContext}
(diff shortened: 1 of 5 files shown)


notification/app/notification/NotificationApplicationLoader.scala
```diff
@@ -114,6 +108,55 @@ class NotificationApplicationComponents(context: Context) extends BuiltInCompone
...
+  def withFilter(notificationSender: NotificationSender, invertCondition: Boolean): NotificationSender =
+    new FilteredNotificationSender(notificationSender, topicRegistrationCounter, invertCondition)
+
+  lazy val notificationSenders = List(
+    withFilter(guardianIosNotificationSender, invertCondition = false),
+    withFilter(guardianAndroidNotificationSender, invertCondition = false),
+    withFilter(guardianNewsstandNotificationSender, invertCondition = false),
+    gcmNotificationSender,
+    apnsNotificationSender,
+    newsstandShardNotificationSender,
```
TBonnin: ok so if I understand this means notification will be pass to all senders? right

notification/app/notification/services/FilteredNotificationSender.scala
```diff
@@ -0,0 +1,51 @@
...
+
+class FilteredNotificationSender(
+  notificationSender: NotificationSender,
+  topicRegistrationCounter: TopicRegistrationCounter,
+  invertCondition: Boolean
+)(implicit ec: ExecutionContext) extends NotificationSender {
+
+  private val logger: Logger = Logger.apply(classOf[FilteredNotificationSender])
+
+  val ALLOWED_TOPICS: Set[TopicType] = Set(TopicTypes.Content, TopicTypes.TagContributor, TopicTypes.TagSeries)
```
TBonnin: uppercase?

notification/app/notification/services/FilteredNotificationSender.scala
```diff
@@ -0,0 +1,51 @@
...
+      }
+    }
+
+    def defaultCountRecovery: PartialFunction[Throwable, PlatformCount] = {
+      case NonFatal(e) if push.notification.topic.exists(_.`type` == TopicTypes.Breaking) =>
+        logger.error(s"Unable to count registration for topics ${push.notification.topic}, falling back to default value", e)
+        PlatformCount(1500000, 750000, 750000, 0)
+      case NonFatal(e) =>
+        logger.error(s"Unable to count registration for topics ${push.notification.topic}, falling back to default value", e)
+        PlatformCount(1, 1, 1, 0)
```
TBonnin: There is a risk a notification count fails here, that will make the total count <= 100 and shouldSend will return true when it shouldn't

notification/app/notification/services/guardian/ReportTopicRegistrationCounter.scala
```diff
@@ -8,16 +8,18 @@ import scala.concurrent.duration.DurationLong
...
 )(implicit ec: ExecutionContext) extends TopicRegistrationCounter {
 
   val lruCache: LruCache[PlatformCount] = new LruCache[PlatformCount](200, 1000, 3.days)
 
   override def count(topics: List[Topic]): Future[PlatformCount] = {
     lruCache(topics.toSet) {
       val topicParameters = topics.map(topic => "topics" -> topic.toString)
-      ws.url(registrationCounterUrl)
-        .withQueryStringParameters(topicParameters: _*)
+      val allParams = topicParameters ++ List("api-key" -> apiKey)
```
TBonnin: can you not append a single tuple: `val allParams = topicParameters :+ ("api-key" -> apiKey)`

notification/app/notification/NotificationApplicationLoader.scala
alexduf: Yes, you're correct, but we filter notifications sent on the SQS for now

PR #237, opened 2018-11-05
Remove filter, and go full dry run mode

alexduf:
The initial dry run was successful, now I want to see how the database handles big queries.


alexduf:
You're right, probably safer. Let's see bigger notifications then, I've put the threshold at 50k

TBonnin:
👍

PR #239, opened 2018-11-06
Add registration cleaning worker

alexduf:
Receives events via SQS.

The events aren't sent yet, I'll add that in a subsequent PR

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/RegistrationCleaningWorker.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/RegistrationCleaningWorker.scala
@@ -1,0 +1,67 @@
+package com.gu.notifications.worker
+
+import cats.effect.{ContextShift, IO}
+import cats.effect.IO._
+import com.amazonaws.services.lambda.runtime.{Context, RequestHandler}
+import com.amazonaws.services.lambda.runtime.events.SQSEvent
+import com.gu.notifications.worker.models.InvalidTokens
+import db.{DatabaseConfig, RegistrationService}
+import doobie.util.transactor.Transactor
+import org.slf4j.{Logger, LoggerFactory}
+import play.api.libs.json.Json
+
+import scala.collection.JavaConverters._
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor}
+import scala.util.control.NonFatal
+
+class RegistrationCleaningWorker extends RequestHandler[SQSEvent, Unit] {
+
+  implicit val logger: Logger = LoggerFactory.getLogger(this.getClass)
+  implicit val ec: ExecutionContextExecutor = ExecutionContext.global
+  implicit val ioContextShift: ContextShift[IO] = IO.contextShift(ec)
+  val config: FcmWorkerConfiguration = Configuration.fetchFirebase()
+  val transactor: Transactor[IO] = DatabaseConfig.transactor[IO](config.jdbcConfig)
+  val registrationService = RegistrationService(transactor)
+
+  case class CleaningResult(
+    deletedRegistrations: Int,
+    deletedRows: Int,
+    failures: Int) {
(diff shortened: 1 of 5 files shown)


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/RegistrationCleaningWorker.scala
```diff
@@ -0,0 +1,67 @@
...
+      .getRecords.asScala
+      .map(_.getBody)
+      .map(Json.parse)
+      .flatMap(_.validate[InvalidTokens].asOpt)
+      .foldLeft(List.empty[String]){ case (agg, value) => agg ++ value.tokens }
+
+    val processReport = traverseAndDelete(tokens, CleaningResult(0, 0, 0)).unsafeRunSync()
+
+    logger.info(s"Deleted ${processReport.deletedRows} rows, deleted ${processReport.deletedRegistrations} registration and failed ${processReport.failures} time(s)")
+  }
```
TBonnin: I will keep using streams to handle the logic. It make it quite easy:

```
    def printResult(result: CleaningResult): IO[Unit] = IO(
      logger.info(s"Deleted ${result.deletedRows} rows, deleted ${result.deletedRegistrations} registration and failed ${result.failures} time(s)")
    )

    val tokens: List[String] = input
      .getRecords.asScala
      .map(_.getBody)
      .map(Json.parse)
      .flatMap(_.validate[InvalidTokens].asOpt)
      .flatMap(_.tokens)
      .toList

    Stream
      .emits(tokens).covary[IO]
      .evalMap(deleteAndSwallowError) // or .parEvalMapUnordered(awsApiConcurrencyLimit)(deleteAndSwallowError) for parallel execution
      .reduce(_ combined _)
      .evalTap(printResult)
      .compile
      .drain
      .unsafeRunSync()
```

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/InvalidTokens.scala
```diff
@@ -0,0 +1,11 @@
...
+
+import play.api.libs.json.{Format, Json}
+
+case class InvalidTokens(
+  tokens: List[String]
+)
+
+object InvalidTokens {
+  implicit val invalidTokensJF: Format[InvalidTokens] = Json.format[InvalidTokens]
+}
```
TBonnin: APNS provides a token invalidation timestamp that could be use to avoid discarding a token that would have been refreshed since last time it was fetched. I am not sure we need to handle this edge case though

alexduf: Yes I'm hesitating to implement that, I'm worried it might introduce more edge cases

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/RegistrationCleaningWorker.scala
```diff
@@ -0,0 +1,67 @@
...
+      }
+      case Nil => IO.pure(total)
+    }
+
+    val tokens = input
+      .getRecords.asScala
+      .map(_.getBody)
+      .map(Json.parse)
+      .flatMap(_.validate[InvalidTokens].asOpt)
+      .foldLeft(List.empty[String]){ case (agg, value) => agg ++ value.tokens }
```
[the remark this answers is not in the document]
alexduf: Now that you mention id, yeah why? :smile:

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/RegistrationCleaningWorker.scala
alexduf: Got it, Streams still looks like alien technology to me but I'll get used to it.

PR #241, opened 2018-11-07
Count dry run directly, and send the metric. Don't trace each dry run token

alexduf:


TBonnin:
Fair. Is there anything in the log that would say the lambda is running in dry-mode.

TBonnin:
Answering my own question: the total at the end will should a dry run count.

PR #243, opened 2018-11-08
Ignore empty notifications

alexduf:


notification/app/notification/services/guardian/GuardianNotificationSender.scala
```diff
@@ -89,15 +89,19 @@ class GuardianNotificationSender(
   }
 
   def shard(registrationCount: Int): List[ShardRange] = {
-    val shardSpace = -Short.MinValue.toInt + Short.MaxValue.toInt
-    val shardCount = Math.ceil(Math.max(1, registrationCount) / WORKER_BATCH_SIZE.toDouble)
-    val step = Math.ceil(shardSpace / shardCount).toInt
+    if (registrationCount == 0) {
```
TBonnin: why would this be 0?

alexduf: if we send a topic (say morning briefing) to a combination of topic / platform that has 0 registrations (newsstand won't have any registration for it)

It could also be that no-one follows that topic.

alexduf: Right now a newsstand notification is triggered for each notification, which isn't ideal.

PR #245, opened 2018-11-09
Optimise workers

alexduf:
- reduce the number of initial connections to 2 per workers, so that a big notification doesn't saturate the database
- bump the retry as the EventSourceMapping needs it to handle messages properly
- treat the delay of the retry as an initial delay applied no matter what

### DIFF
--- a/notificationworkerlambda/cfn.yaml
+++ b/notificationworkerlambda/cfn.yaml
@@ -130,6 +130,7 @@ Resources:
       Role: !GetAtt ExecutionRole.Arn
       Runtime: java8
       Timeout: 300
+      ReservedConcurrentExecutions: 100
       VpcConfig:
         SecurityGroupIds:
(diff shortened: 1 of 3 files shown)


notificationworkerlambda/cfn.yaml
```diff
@@ -134,7 +130,7 @@ Resources:
       Role: !GetAtt ExecutionRole.Arn
       Runtime: java8
       Timeout: 300
-      ReservedConcurrentExecutions: !Ref Concurrency
+      ReservedConcurrentExecutions: 100
```
markrichardsg: I might be getting the math wrong on this but:
maxConnections = LEAST({DBInstanceClassMemory/9531392},5000)
db.m4.large has 8gb ram
8000000000/9531392 = ~839

100 lambda executions * 20 fixedThreadPool = 2000 connections.
Will we have threadpool starvation: potentially lambdas starting that are unable to get any threads?

PR #253, opened 2018-11-14
Send invalid tokens to cleaner worker

alexduf:
The key parts of this PR are:

 - split the cloudformation template in two, one for the ios and android worker, one for the cleaner as the platform workers needs an extra permission in order to write to SQS
 - the FcmClient now unwraps ExecutionException which I believe are exceptions happening on a different thread. Doing so allows us to keep pattern matching. The error code has been changed to match what we observe in production (the firebase documentation isn't quite right)
 - the pipeline now has two sinks at the end, one to log successes, one to process failures and send them to the cleaner


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/WorkerRequestHandler.scala
```diff
@@ -45,16 +48,32 @@ trait WorkerRequestHandler[C <: DeliveryClient] extends RequestHandler[SQSEvent,
...
       n <- sharedNotification
       notificationLog = s"(notification: ${n.notification.id} ${n.range})"
       _ = logger.info(s"Sending notification $notificationLog...")
       resp <- deliveryService.send(n.notification, n.range)
-        .through(Reporting.report(s"Sending failure: "))
-        .fold(SendingResults.empty){ case (acc, resp) => SendingResults.inc(acc, resp) }
-        .through(logInfo(prefix = s"Results $notificationLog: "))
-        .through(Cloudwatch.sendMetrics(env.stage, platform))
+        .evalTap(Reporting.log(s"Sending failure: "))
+        .broadcastTo(reportSuccesses(n), cleanupFailures)
```
TBonnin: 👍

PR #257, opened 2018-11-14
Bump the threshold high enough to deliver the morning briefing

alexduf:
The last increase (10K) didn't bring that many runs (1 @ 6k) as notifications that big are less frequent.

I suggest we now increase it up to tomorrow's morning briefing. As a reminder the breaking news are still being filtered and won't be delivered through our system yet.


alexduf:
Tested on CODE :heavy_check_mark:

PR #260, opened 2018-11-16
Concurrency and disable briefing

alexduf:
- Raise the concurrency limit to what it should be (was like that before merging #253 which brought the bug we observe)
- lower the threshold of what the guardian system handles. This will force the morning briefing to go through Azure this weekend.


alexduf:
Closed in favour of
https://github.com/guardian/mobile-n10n/pull/261 and https://github.com/guardian/mobile-n10n/pull/262

PR #261, opened 2018-11-16
Filter morning briefing by lowering the filter to a 100 000 registrations

alexduf:


alexduf:
Yes, it's small enough so I'm not worried about it

PR #264, opened 2018-11-16
Only mark invalid tokens as warning

alexduf:
We get quite a lot of these, marking them as error leaves a lot of noise in the logs.
We'll eventually disable them completely as this is just normal behaviour, but in the meantime I'm downgrading these "Errors" to warnings so that searching for an Error in the logs is more meaningful


markrichardsg:
Nice

PR #267, opened 2018-11-23
Activate in-house notifications for breaking news

alexduf:


alexduf:
UK is bellow 1M but growing, and a global alert is processed as 4 separate notifications

NathanielBennett:
:+1:

PR #271, opened 2018-11-26
Direct correlation between delivery speed and memory (CPU)

alexduf:
By doubling the memory last week, we've almost divided the processing time by 2, leading me to believe CPU is actually the current bottleneck.

So I'm increasing the memory size (and by extension the CPU) to the current maximum aws lambdas can take, and I'm looking at the code to see if there's anything that can be done performance wise.

### DIFF
--- a/notificationworkerlambda/platform-worker-cfn.yaml
+++ b/notificationworkerlambda/platform-worker-cfn.yaml
@@ -138,7 +138,7 @@ Resources:
           App: !Ref App
       Description: Consumes queue events and sends notifications
       Handler: !Ref FullyQualifiedHandler
-      MemorySize: 2048
+      MemorySize: 3008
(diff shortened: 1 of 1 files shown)


alexduf:
It is indeed a lot of RAM, but what matters most is how much CPU we get out of it.

I think we're CPU bound, and I haven't found yet what's to optimise in the lambda (off the top of my head I'd say json serialisation). In the meantime I want to see if that makes us linearly faster like last time I've bumped that value, or if we've reached another bottleneck.

TBonnin:
Both apns and firebase clients are firing hundreds of requests in parrallel. I am not surprised this is maxing out the lambda cpu cores. Worth looking into `pushy` and `firebase` config to check if they are configured properly to balance cpu an network throughput

PR #278, opened 2018-12-10
Mss 593/improving lambda cloudwatch alarms

markrichardsg:
Not having errors should not be treated as an alarm, so it seems appropriate to set `TreatMissingData` to `notBreaching`

However, not having data is a worry, so adding `TooFewInvocationsAlarm` which will also be set up on CODE, but because we expect this to alarm on CODE I've set ActionsEnabled to false if not prod.


alexduf:
very nice

PR #281, opened 2018-12-11
add DDLs for the athena tables

alexduf:


athena-tables.sql
```diff
@@ -0,0 +1,85 @@
+CREATE EXTERNAL TABLE `raw_events_code`(
```
markrichardsg: Is it worth putting these in cloudformation?
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-athena-namedquery.html

Cloudformation may do this, but also would it be worth templating these? There seem to be some common variables like stage or even the timeframe?

PR #286, opened 2018-12-14
Removing timings

markrichardsg:


markrichardsg:
@alexduf import error causing that now fixed :)

PR #287, opened 2018-12-14
Mss 599 delete dead code

alexduf:
:chainsaw:


alexduf:
Actually I'll rebase and extract the newsstand change in a separate PR

PR #292, opened 2018-12-17
Testing athena

markrichardsg:
This should only put the athena event statistics into a new column (athena) and if we're happy with it, then we can swap it to events and delete the sqs driven lambda.
There is a next step of adding a trigger for this, but I'll do this after it looks like it runs okay when triggered manually.

### DIFF
--- a/eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
+++ b/eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
@@ -34,7 +34,6 @@ case class Query(database: String, queryString: String, outputLocation: String)
 class AthenaLambda {
 
   import ExecutionContext.Implicits.global
-
   private val envDependencies = new EnvDependencies
   private val logger: Logger = LogManager.getLogger(classOf[AthenaLambda])
   private val dynamoReportUpdater = new DynamoReportUpdater(envDependencies.stage)
--- a/eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
+++ b/eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
@@ -122,7 +122,7 @@ class AthenaLambda {
         FROM notification_received_code
 WHERE partition_date = '${startOfReportingWindow.getYear}-${startOfReportingWindow.getMonthValue}-${startOfReportingWindow.getDayOfMonth}'
         AND partition_hour >= ${startOfReportingWindow.getHour}
-GROUP BY  notificationid, platform, provider""".stripMargin, envDependencies.athenaOutputLocation)
+GROUP BY  notificationid, platform""".stripMargin, envDependencies.athenaOutputLocation)
     startQuery(loadParitionsQuery).thenComposeAsync(_ => route(fetchEventsQuery, startOfReportingWindow)).join()
   }
 
--- a/eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
+++ b/eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
@@ -1,0 +1,127 @@
+package com.gu.notifications.events
+
+import java.time.{Duration, ZonedDateTime}
+import java.util.concurrent.{CompletableFuture, CompletionStage, ConcurrentLinkedQueue}
+
+import com.gu.notifications.events.AthenaLambda.athenaAsyncClient
+import com.gu.notifications.events.dynamo.DynamoReportUpdater
+import com.gu.notifications.events.model.{AggregationCounts, EventAggregation, NotificationReportEvent, PlatformCount}
+import org.apache.logging.log4j.{LogManager, Logger}
+import software.amazon.awssdk.auth.credentials.{AwsCredentialsProviderChain, DefaultCredentialsProvider, ProfileCredentialsProvider}
+import software.amazon.awssdk.regions.Region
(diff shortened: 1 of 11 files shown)


eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
```diff
@@ -0,0 +1,129 @@
...
+
+
+  def route(query: Query, startOfReportingWindow: ZonedDateTime): CompletableFuture[Void] = {
+    startQuery(query)
+      .thenComposeAsync(fetchQueryResponse)
+      .thenComposeAsync(updateDynamoIfRecent(_, startOfReportingWindow))
+      .thenAcceptAsync((aggregationCounts: AggregationCounts) => {
+        logger.info(s"Aggregation counts $aggregationCounts")
+        if (aggregationCounts.failure > 0) {
+          throw new RuntimeException("Failures")
```
alexduf: Should we trace which failure before throwing an error?

eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
```diff
@@ -0,0 +1,129 @@
...
+  def startQuery(query: Query): CompletableFuture[String] = {
+    val startQueryAndGetId: CompletableFuture[String] = executeQuery(query).thenApplyAsync((startQueryExecutionResponse: StartQueryExecutionResponse) => startQueryExecutionResponse.queryExecutionId())
+    startQueryAndGetId.thenComposeAsync((executionId: String) =>
+      waitUntilQueryCompletes(executionId).thenApplyAsync((_: Void) => executionId)
+    )
+  }
+
+
+  private def waitUntilQueryCompletes(id: String): CompletableFuture[Void] = {
+    Thread.sleep(1000)
```
alexduf: Would this be of any use?
https://aws.amazon.com/blogs/developer/waiters-in-the-aws-sdk-for-java/

eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
```diff
@@ -0,0 +1,129 @@
...
+
+
+  private def updateDynamoIfRecent(notificationIdCounts: Map[String, PlatformCount], startOfReportingWindow: ZonedDateTime): CompletionStage[AggregationCounts] = {
+    val notificationReportEvents = notificationIdCounts.toList.map { case (id, count) => NotificationReportEvent(id, EventAggregation(count)) }
+    val resultCounts = dynamoReportUpdater.updateSetEventsReceivedAfter(notificationReportEvents, startOfReportingWindow)
+    val aggregatedCounts = AggregationCounts.aggregateResultCounts(resultCounts)
+    FutureConverters.toJava(aggregatedCounts)
+  }
+
+  private def fetchQueryResponse(id: String)(implicit executionContext: ExecutionContext): CompletableFuture[Map[String, PlatformCount]] = {
```
alexduf: If you want to avoid this complex logic you could read the results line by line.
That would require to aggregate the platform totals at the query levels rather than reading multiple lines and aggregating it here.

for instance

```sql
select 
  count(*),
  count_if(platform = 'android'),
  count_if(platform = 'ios')
from notification_received_prod
where partition_date = '2018-12-17'
```
https://prestodb.io/docs/current/functions/aggregate.html#count_if

eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
markrichardsg: It looks like Waiters have been relegated to the test-utils ☹️

alexduf: It might be a different one, are we using the sdk v2?
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/waiters/package-summary.html

TBonnin:
I was wondering if it would make sense for you to use the Athena JDBC driver (and your favorite DB library in scala) rather than using the awkward Athena API and the waiting dance to fetch results.
I have not used it from a lambda though so it might be a bad idea.

markrichardsg:
@TBonnin maybe for a bigger use case, but I doubt we run enough queries in this Lambda to benefit or lose much either way.

eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
markrichardsg: Nice

eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
```diff
@@ -0,0 +1,125 @@
...
+    val queue = new ConcurrentLinkedQueue[List[(String, PlatformCount)]]()
+    athenaAsyncClient.getQueryResultsPaginator(GetQueryResultsRequest.builder()
+      .queryExecutionId(id)
+      .build()).subscribe((getQueryResultsResponse: GetQueryResultsResponse) => {
+      queue.add(getQueryResultsResponse.resultSet.rows().asScala.toList.tail.map { row =>
```
alexduf: `toList.tail` aren't you skipping a row here?

markrichardsg: Yes, the head is the headers; added a commit to print this as it makes the code more self-explanatory for future readers.

PR #293, opened 2018-12-17
Add athena tables DDL

alexduf:


athena-tables.sql
```diff
@@ -0,0 +1,85 @@
...
+    LOCATION
+    's3://<PROD_BUCKET>/fastly';
+
+
+CREATE OR REPLACE VIEW notification_received_prod AS
+  SELECT
+    "dateTime" "received_timestamp",
+    "regexp_extract"("querystring", 'notificationId=([\w-]+)', 1) "notificationId",
+    "regexp_extract"("querystring", 'platform=(\w+)', 1) "platform",
+    "regexp_extract"("querystring", 'provider=(\w+)', 1) "provider",
```
markrichardsg: We don't need the provider any more.

athena-tables.sql
```diff
@@ -0,0 +1,85 @@
...
+    OUTPUTFORMAT
+    'org.apache.hadoop.hive.ql.io.HiveIgnoreKeyTextOutputFormat'
+    LOCATION
+    's3://<PROD_BUCKET>/fastly';
+
+
+CREATE OR REPLACE VIEW notification_received_prod AS
+  SELECT
+    "dateTime" "received_timestamp",
+    "regexp_extract"("querystring", 'notificationId=([\w-]+)', 1) "notificationId",
```
markrichardsg: We might want to add a WHERE != unknown clause due to the bug that's been fixed, but still populating s3 with invalid events.

athena-tables.sql
alexduf: It's a field sent by the device, so I want to keep it

athena-tables.sql
alexduf: I'm not sure that's a concern the table should have

PR #299, opened 2018-12-20
MSS-595: Delete sqs lambda

markrichardsg:
Deleting the sqs consumer lambda and s3 events to populate it.
Risk of conflict when updating counts (with the update still done in notifications) improved as a conditional.


markrichardsg:
Do not merge for a bit to see how the metrics compare.

PR #304, opened 2018-12-24
Log the request when there is a failure

markrichardsg:
When Play logs the exception the stack trace should hopefully include both the request and error

### DIFF
--- a/common/src/main/scala/aws/AWSAsync.scala
+++ b/common/src/main/scala/aws/AWSAsync.scala
@@ -6,9 +6,9 @@ import java.util.concurrent.{Future => JFuture}
 import scala.concurrent.{Future, Promise}
 
 object AWSAsync {
-  private def promiseToAsyncHandler[Request <: AmazonWebServiceRequest, Result](p: Promise[Result]) =
+  private def promiseToAsyncHandler[Request <: AmazonWebServiceRequest, Result](p: Promise[Result], request: AmazonWebServiceRequest) =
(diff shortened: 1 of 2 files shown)


common/src/main/scala/aws/AWSAsync.scala
```diff
@@ -6,9 +6,9 @@ import java.util.concurrent.{Future => JFuture}
 import scala.concurrent.{Future, Promise}
 
 object AWSAsync {
-  private def promiseToAsyncHandler[Request <: AmazonWebServiceRequest, Result](p: Promise[Result]) =
+  private def promiseToAsyncHandler[Request <: AmazonWebServiceRequest, Result](p: Promise[Result], request: AmazonWebServiceRequest) =
     new AsyncHandler[Request, Result] {
-      override def onError(exception: Exception): Unit = { p.failure(exception); () }
+      override def onError(exception: Exception): Unit = {p.failure(new RuntimeException(s"Dynamo request failed $request", exception)); ()}
```
alexduf: This part is meant to be used by any AWS service, so I'm not too keen on tracing the Dynamo specific request here.

Could that be simply done at the time the AWSAsync is used? The future is failed with an exception, so it should be possible to have myFuture.recover { case NonFatal(e) => blah blah} and find a way to pass the request at that point?

markrichardsg: Changing this for notification only for now.

PR #305, opened 2018-12-27
Add ttl for notifications that aren't breaking news

markrichardsg:


common/src/main/scala/models/DynamoNotificationReport.scala
```diff
@@ -50,18 +54,47 @@ object SenderReport {
...
+    reports,
+    version,
+    events,
+    ttlFromSentTime(sentTime, `type`)
+  )
+
+  private def ttlFromSentTime(sentTime: DateTime, notificationType: NotificationType): Option[Long] = {
+    notificationType match {
+      case BreakingNews => None
+      case _ => Some(sentTime.plusMonths(3).getMillis / 1000)
```
alexduf: :+1:

PR #306, opened 2019-01-03
MSS-595: Log athena polling states

markrichardsg:
When the lambda errors there might be a bad state indicated here so worth logging.
Also, there's a risk the query is in remaining in QUEUED or RUNNING state. This should make that clear.


eventconsumer/src/main/scala/com/gu/notifications/events/AthenaLambda.scala
```diff
@@ -66,7 +66,9 @@ class AthenaLambda {
     athenaAsyncClient.getQueryExecution(GetQueryExecutionRequest.builder()
       .queryExecutionId(id)
       .build()).thenComposeAsync((response: GetQueryExecutionResponse) => {
-      response.queryExecution().status().state() match {
+      val state = response.queryExecution().status().state()
+      logger.info(state.toString)
```
NathanielBennett: Is it worth-adding something that you can search for in all cases here?

markrichardsg: Added id too as the id should be visible in the console later

PR #309, opened 2019-01-04
Adding ok http client with random DNS resolution

markrichardsg:


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
```diff
@@ -71,15 +73,28 @@ class FcmClient private (firebaseMessaging: FirebaseMessaging, firebaseApp: Fire
 
 object FcmClient {
 
+  def randomlySetOkHttpTransport(builder: FirebaseOptions.Builder): FirebaseOptions.Builder = {
+    val logger = LoggerFactory.getLogger(classOf[FcmClient])
+    val randomNumber = Random.nextInt(100)
+    val percentageOkHttpClient = 50
```
markrichardsg: This percentage is probably too high. What percentage should we use?

alexduf: let's start with 5 so we can see it live in the logs, then we'll ramp it up and measure if there's any impact on the NAT metrics

markrichardsg: Sounds good. Changed to 5

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/oktransport/OkGoogleHttpRequest.scala
```diff
@@ -0,0 +1,100 @@
...
+import okhttp3.{Headers, MediaType, OkHttpClient, Request, RequestBody, Response}
+import okio.BufferedSink
+
+import scala.annotation.tailrec
+import scala.collection.mutable.ArrayBuffer
+
+
+class OkGoogleHttpRequest(okHttpClient: OkHttpClient, url: String, method: String) extends LowLevelHttpRequest {
+  //No idea whether FCM now or in the future will call addHeaders concurrently
+  private val headers = new ConcurrentLinkedQueue[(String, String)]()
```
alexduf: I think this is a premature optimisation. An array, list or Map would do just fine

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/oktransport/OkGoogleHttpRequest.scala
```diff
@@ -0,0 +1,100 @@
...
+    val headerTuples = ArrayBuffer[(String, String)]()
+    val okHeaders: Headers = response.headers()
+    val headerCount = okHeaders.size()
+    for (headerIndex <- 0 until headerCount) {
+      headerTuples += ((okHeaders.name(headerIndex), okHeaders.value(headerIndex)))
+    }
+    headerTuples.toList
+  }
+
+  private def readBody(response: Response): (Option[String], Option[Long], Option[ByteArrayInputStream]) = {
```
alexduf: It looks like the reason you return these three attribute together is because you want to close the body after.

What I suggest would be to have a function defined like so:

```scala
private def withResponseBody[A](response: response)(f: Body => A): A {
  val body = Option(response.body())
  val result = body.map(f)
  body.foreach(_.close())
  result
}
```

then you can pick whatever attributes you want from the body and that allows you to keep the truple definition local to the function that uses it

```scala
val (maybeContentType: Option[String], maybeContentLength: Option[Long], maybeContent: Option[ByteArrayInputStream]) = withResponseBody(response) { body =>
  (Option(body.getContentType), Option(...))
}

```

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/oktransport/OkGoogleHttpTransport.scala
```diff
@@ -0,0 +1,25 @@
...
+import okhttp3.{Dns, OkHttpClient}
+
+import scala.collection.JavaConverters._
+import scala.util.Random
+
+
+class OkGoogleHttpTransport extends HttpTransport {
+  private val okHttpClient = new OkHttpClient.Builder()
+    .dns((hostname: String) => Random.shuffle(Dns.SYSTEM.lookup(hostname).asScala).asJava)
+    .followRedirects(false)
```
alexduf: Shouldn't that be true by default?

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/oktransport/OkGoogleHttpRequest.scala
alexduf: After our chat, probably fine to keep like that

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/oktransport/OkGoogleHttpTransport.scala
markrichardsg: I thought so, but when I tested it against https://www.theguardian.com (which includes a redirect) the client FCM already uses does not redirect; so I disabled it to match the behaviour of the existing client.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/oktransport/OkGoogleHttpRequest.scala
markrichardsg: Done

PR #311, opened 2019-01-07
Add worker unit tests

alexduf:
This isn't my proudest unit test, I find it very hard to test side effects with fs2. @TBonnin  do you know if I've missed a trick?

I toyed with the idea of using Mockito here, but the test was just becoming unreadable (also, it didn't work).

### DIFF
--- a/notificationworkerlambda/src/test/scala/com/gu/notifications/worker/WorkerRequestHandlerSpec.scala
+++ b/notificationworkerlambda/src/test/scala/com/gu/notifications/worker/WorkerRequestHandlerSpec.scala
@@ -23,7 +23,7 @@ class WorkerRequestHandlerSpec extends Specification with Matchers {
 
   "the WorkerRequestHandler" should {
     "Send one notification" in new WRHSScope {
-      processShardedNotification(Stream(shardedNotification))
+      workerRequestHandler.processShardedNotification(Stream(shardedNotification))
 
       deliveryCallsCount shouldEqual 1
       cloudwatchCallsCount shouldEqual 1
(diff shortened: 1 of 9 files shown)


notificationworkerlambda/src/test/scala/com/gu/notifications/worker/WorkerRequestHandlerSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+    var deliveryCallsCount = 0
+    var cleaningCallsCount = 0
+    var tokensToCleanCount = 0
+    var cloudwatchCallsCount = 0
+    var sendingResults: Option[SendingResults] = None
+
+    val workerRequestHandler = new WorkerRequestHandler[ApnsClient] {
+      override def platform: Platform = iOS
+
+      override def deliveryService: IO[DeliveryService[IO, ApnsClient]] = IO.pure(new DeliveryService[IO, ApnsClient] {
```
TBonnin: Do you need override?

markrichardsg:
It seems like the aim of the Handler is application flow.

The specifics like how to report successes, cleanup failures, determine platform type, could be extracted into other module(s) reducing the amount going on inside `WorkerRequestHandler`

notificationworkerlambda/src/test/scala/com/gu/notifications/worker/WorkerRequestHandlerSpec.scala
alexduf: Technically I don't think the keyword override is necessary, it just happen it's included automatically by intelliJ. It also has the nice effect that if the signature changes, or the attribute is dropped, I'll get a compilation error, whereas dropping the keyword won't catch it.

PR #316, opened 2019-01-15
handle old android devices without firebase token

NathanielBennett:
Deals with special case to handle (android) devices that haven't updated their app in a while and don't have an FCM token. 

@alexduf have I totally misunderstood this?

### DIFF
--- a/common/src/main/scala/error/Errors.scala
+++ b/common/src/main/scala/error/Errors.scala
@@ -5,5 +5,5 @@ case class UnsupportedPlatform(platform: String) extends RequestError {
 }
 
 case class MalformattedRegistration(description: String) extends RequestError {
-  override def reason: String = s"Malformatred request: $reason"
+  override def reason: String = s"Malformatred request: $description"
 }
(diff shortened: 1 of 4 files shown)


alexduf:
This is a much better implementation than what I had in mind. I think you may be missing something at the controller level such that in case of a malformed registration the server returns an HTTP 4xx rather than a 500. (Or am I wrong and is it already the case?)

NathanielBennett:
I was assuming that `MalformattedRegistration` subclasses `RequestError` that this line was cause a 400 to be returned https://github.com/guardian/mobile-n10n/blob/MSS-609/add-special-case-for-registrations-from-old-android-apps/registration/app/registration/controllers/Main.scala#L147.

alexduf:
I think you've mangled Mark's branch with yours, do you want to rebase from master?

NathanielBennett:
Oops hang on

common/src/main/scala/error/Errors.scala
```diff
@@ -5,5 +5,5 @@ case class UnsupportedPlatform(platform: String) extends RequestError {
 }
 
 case class MalformattedRegistration(description: String) extends RequestError {
-  override def reason: String = s"Malformatred request: $reason"
+  override def reason: String = s"Malformatred request: $description"
```
NathanielBennett: @alexduf good thing you suggested the curl test .. 
(presumably the default case here: https://github.com/guardian/mobile-n10n/blob/master/registration/app/registration/services/LegacyRegistrationConverter.scala#L15 has never been reached ..

NathanielBennett:
OK all good now

PR #320, opened 2019-01-17
Mss 610/add api key header

NathanielBennett:
Retrieves the api-key (passed to the notifications-client) as a Authorisation header rather than a url parameter: https://theguardian.atlassian.net/browse/MSS-610


common/src/main/scala/authentication/AuthAction.scala
```diff
@@ -18,11 +18,12 @@ abstract class AuthAction(controllerComponents: ControllerComponents) extends Ac
...
 
   override def invokeBlock[A](request: Request[A], block: RequestWithAuthentication[A] => Future[Result]): Future[Result] = {
-    request.getQueryString("api-key") match {
+    getApiKey(request) match {
       case Some(apiKey) if validApiKey(apiKey) => block(RequestWithAuthentication(isPermittedTopic(apiKey), request))
       case _ => Future.successful(Unauthorized("A valid api key is required"))
     }
   }
 
+  private def getApiKey[A](request: Request[A]) = request.headers.get("Authorization").orElse(request.getQueryString("api-key"))
```
markrichardsg: To be HTTP compliant, I believe we should prefix the Authorization header with something like `Bearer`

https://security.stackexchange.com/questions/108662/why-is-bearer-required-before-the-token-in-authorization-header-in-a-http-re

Interested in thoughts as I'm not sure we have to be compliant if we don't want to be.

alexduf: yes I think that would be nice to ensure the key is prefixed by "Bearer ", just to make it industry standard

NathanielBennett: Done

PR #321, opened 2019-01-17
Mss 610/update api client with api key header

NathanielBennett:
Alters the notifications-api-client to take the api-key as a http header rather than url parameter: See https://theguardian.atlassian.net/browse/MSS-610

### DIFF
--- a/api-client/src/test/scala/com/gu/mobile/notifications/client/NextGenApiClientSpec.scala
+++ b/api-client/src/test/scala/com/gu/mobile/notifications/client/NextGenApiClientSpec.scala
@@ -24,8 +24,9 @@ class NextGenApiClientSpec(implicit ee: ExecutionEnv) extends ApiClientSpec[Next
     topic = List(Topic(Breaking, "n1")),
     debug = true
   )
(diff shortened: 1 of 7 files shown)


api-client/version.sbt
```diff
@@ -1 +1 @@
-version := "1.3-SNAPSHOT"
\ No newline at end of file
+version := "1.4"
```
alexduf: not yet :) that's automatically updated when you release so I'd recommend you leave it the way it was

api-client/src/test/scala/com/gu/mobile/notifications/client/NextGenApiClientSpec.scala
```diff
@@ -24,8 +24,9 @@ class NextGenApiClientSpec(implicit ee: ExecutionEnv) extends ApiClientSpec[Next
     topic = List(Topic(Breaking, "n1")),
     debug = true
   )
+  //  val expectedPostUrl = s"$host/push/topic?api-key=$apiKey"
```
alexduf: can be removed

PR #322, opened 2019-01-17
OK HTTP to 50%

markrichardsg:


alexduf:
I'm not sure we're going to push that client to 50% yet, I want more datapoints first

alexduf:
I think we should merge this one now, we probably have enough data points to compare. It's also early in the week so we can catch anything wrong

PR #328, opened 2019-01-25
publish to bintray

NathanielBennett:
Publishes apiClient to bintray rather than (not actually publishing it) to sonatype


alexduf:
isn't there a plugin or something to modify in the plugins?

NathanielBennett:
@alexduf Thanks! This is what happens when you try to go vegan for four days a week, you have friday afternoon off and you're anticipating the all day breakfast you're going to have for lunch ...

PR #335, opened 2019-01-30
Fix report authentication

alexduf:
@NathanielBennett I forgot one. This isn't the end of the world as if this HTTP call fails the service assumes the worse case and considers the notification to be a big notification.

I'm testing it on CODE real quick and we should merge that soon after


NathanielBennett:
@alexduf Should be OK, after we reverted yesterday. I've raised this PR: https://github.com/guardian/mobile-n10n/pull/334 to re implement the change but aren't going to merge until after I've changed the newsstand script with Central Prod and checked it works OK on Thursday night

alexduf:
Oh yes of course!
I've tested it on CODE so I'll merge it anyway. Thanks!

PR #336, opened 2019-02-05
MSS-613: Use last modified as an epoch

markrichardsg:


NathanielBennett:
Er, fix the tests, obviously. But it doesn't seem unreasoable

alexduf:
Why?

NathanielBennett:
@alexduf we can't migrate a datetime column. Life sucks. You and @markrichardsg  should probably have a chat later. Feel free to dial me in

markrichardsg:
https://docs.aws.amazon.com/dms/latest/userguide/CHAP_Source.PostgreSQL.html
In section _Limitations on Using a PostgreSQL Database as a Source for AWS DMS_
_Timestamp with a time zone type column is not supported._

PR #340, opened 2019-02-14
Remove unused configuration

alexduf:
Before deleting the keys from SSM, let's ensure they aren't loaded


NathanielBennett:
:+1: Plan

PR #342, opened 2019-02-15
Mss 158/monitor newsstand notifications

NathanielBennett:
Adds alert to test that newsstand has been published

### DIFF
--- a/common/cfn/notification.yaml
+++ b/common/cfn/notification.yaml
@@ -431,12 +431,12 @@ Resources:
           - triggers if less that ${DailyNewsstandPushCount} daily edition notification was sent for ${Stage} in the last 24 hours..
           - DailyNewsstandPushCount: !FindInMap [ StageVariables, !Ref Stage, DailyNewsstandPushCount ]
       ComparisonOperator: LessThanThreshold
-      EvaluationPeriods: 28
+      EvaluationPeriods: 1
       Namespace:
         !FindInMap [ StageVariables, !Ref Stage, MetricNamespace ]
       MetricName:
         !FindInMap [ StageVariables, !Ref Stage, NewsstandNotificationMetricName ] 
-      Period: 3600
+      Period: 86400
       Statistic: Sum
       Threshold: !FindInMap [ StageVariables, !Ref Stage, NewsstandNotificationAlarmThreshold ]
       TreatMissingData: notBreaching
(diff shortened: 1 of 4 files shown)


common/cfn/notification.yaml
```diff
@@ -417,4 +421,25 @@ Resources:
       Threshold: 0
       TreatMissingData: notBreaching
 
+  NewsstandSentAlarm:
+    Type: AWS::CloudWatch::Alarm
+    Properties:
+      AlarmActions: [!Ref AlarmTopic]
+      AlarmDescription: !Sub
+          - triggers if less that ${DailyNewsstandPushCount} sent for ${Stage} in the last 24 hours.
```
alexduf: - triggers if less that ${DailyNewsstandPushCount} daily edition notification was sent for ${Stage} in the last 24 hours.

NathanielBennett: Yes so less than 1 for PROD, less than 0 for CODE

alexduf: No I mean if you expand the sentence without the variable name, you lose the context:
>triggers if less that 1 sent for PROD in the last 24 hours.

I'm suggesting you add context such that the description includes what we're monitoring

>triggers if less that 1 daily edition notification was sent for PROD in the last 24 hours.

markrichardsg: I might be confused, but is less than 24 hours too strict?
If they're early one day and late the next  (maybe editorial is held up) then it'd be more than 24 hours?
Daylight savings will also push this to 25 hours once a year.
Perhaps this could be 26hours .

alexduf: That's a good point, also the daily edition might be published late, so 26 to 28 hours sounds about right

NathanielBennett: Changed to 28 and updated description

NathanielBennett: @alexduf @markrichardsg 28 hours isn't a goer. It breaks in cfn : "

  | Metrics  cannot be checked across more than a day (EvaluationPeriods * Period  must be <= 86400) (Service: AmazonCloudWatch; Status Code: 400; Error  Code: ValidationError; Request ID:  0bbdd85a-3372-11e9-a0f6-7dcbb9aa765a)"

I'm having a think about what else could work ..

alexduf: Maybe you could put the `Period` to 1h, and the `EvaluationPeriods` to 28? I have no idea if that would work

PR #347, opened 2019-02-20
Missing a newsstand success metric is an alarm condition

markrichardsg:


NathanielBennett:
Do we want to fix this function to send zero to the same metric if the push fails?

https://github.com/guardian/mobile-n10n/blob/master/notification/app/notification/controllers/Main.scala#L42

PR #355, opened 2019-02-22
Mss 751/segregate token fetching from delivery

markrichardsg:
I've not worried about perfectly clean code as some of this should be temporary.

Also, what chunk size should we use?

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/WorkerRequestHandler.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/WorkerRequestHandler.scala
@@ -8,10 +8,11 @@ import com.gu.notifications.worker.delivery._
 import models.SendingResults
 import com.gu.notifications.worker.utils.{Cloudwatch, Logging, NotificationParser, Reporting}
 import org.slf4j.{Logger, LoggerFactory}
-import fs2.{Sink, Stream}
+import fs2.{Chunk, Sink, Stream}
 import _root_.models.{Platform, ShardedNotification}
 import com.gu.notifications.worker.cleaning.CleaningClient
 import com.gu.notifications.worker.delivery.DeliveryException.InvalidToken
+import com.gu.notifications.worker.tokens.TokenService
 
 import collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutor}
@@ -32,8 +33,10 @@ trait WorkerRequestHandler[C <: DeliveryClient] extends RequestHandler[SQSEvent,
 
   def platform: Platform
   def deliveryService: IO[DeliveryService[IO, C]]
+  def tokenService: IO[TokenService[IO]]
   val cleaningClient: CleaningClient
   val cloudwatch: Cloudwatch
+  def maxConcurrency: Int
 
   def env = Env()
   implicit val ec: ExecutionContextExecutor = ExecutionContext.global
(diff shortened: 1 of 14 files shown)


alexduf:
Is this ready for review?

markrichardsg:
@alexduf yes, there are three paths routing can take: deliver straight from the db, queue for delivery and dequeue to deliver.
I've not wired in the queue for delivery logic - yet: need to choose a low risk candidate first which can be done after this is  merged.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/DeliveryService.scala
```diff
@@ -92,11 +75,8 @@ class DeliveryServiceImpl[F[_], C <: DeliveryClient] (
...
     for {
       payload <- Stream.eval(payloadF)
-      res <- tokens
-        .map(token => sending(client)(token, payload))
-        .parJoin(maxConcurrency)
+      res <- sending(client)(token, payload)
     } yield res
   }
 
-  def close(): F[Unit] = F.delay(client.close())
```
alexduf: Was that unused?

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/tokens/SqsDeliveryService.scala
```diff
@@ -0,0 +1,66 @@
...
+    .withCredentials(Aws.credentialsProvider)
+    .withRegion(Regions.EU_WEST_1)
+    .build()
+
+
+  private def send(chunkedTokens: ChunkedTokens)(oncomplete: Either[Exception, Unit] => Unit): Unit = {
+    amazonSQSAsyncClient.sendMessageAsync(new SendMessageRequest()
+      .withMessageBody(Json.stringify(Json.toJson(chunkedTokens)))
+        .withQueueUrl(queueUrl)
+      , new AsyncHandler[SendMessageRequest, SendMessageResult] {
```
alexduf: Can you use AWSAsync.wrapAsyncMethod here?

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/NotificationParser.scala
```diff
@@ -1,18 +1,44 @@
...
 
-  def parseShardedNotification(input: String): IO[ShardedNotification] = {
-    IO(
-      Json.parse(input).validate[ShardedNotification] match {
-        case JsSuccess(shard, _) => shard
+object NotificationParser {
+  def parseEventNotification(input: String): IO[Either[ShardedNotification, ChunkedTokens]] = {
+    IO {
+      val json: JsValue = Json.parse(input)
+      json.validate[Event] match {
```
alexduf: I think it's fine for now, but we should add a `type` attribute the event upstream to branch on the value of the `type` attribute as a next step.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/WorkerRequestHandler.scala
```diff
@@ -75,16 +81,51 @@ trait WorkerRequestHandler[C <: DeliveryClient] extends RequestHandler[SQSEvent,
...
+        .broadcastTo(reportSuccesses(n.notification.id, n.range), cleanupFailures)
     } yield resp
 
+    def queueShardedNotification(n: ShardedNotification): Stream[IO, Either[Throwable, Unit]] = {
+      for {
+        sqsDeliveryService <- Stream.eval(sqsDeliveryService)
+        tokenService <- Stream.eval(tokenService)
+        notificationLog = s"(notification: ${n.notification.id} ${n.range})"
+        _ = logger.info(s"Queuing notification $notificationLog...")
+        resp <- tokenService.tokens(n.notification, n.range, platform).chunkN(10000)
```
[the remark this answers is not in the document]
markrichardsg: Looks like 1000 messages in 10 batches might be best given the size of tokens can be >150 characters.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/tokens/SqsDeliveryService.scala
markrichardsg: That returns a Future and I wasn't sure whether Future's were best to mix in with fs2. Let's chat about this.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/DeliveryService.scala
markrichardsg: I belive so, I expected a compiler error when I deleted that line and it didn't happen.
 It might be valuable to re-use connections as much as possible (especially over a nat) so this may be a good thing (for now) and given the Lambda lifecycle that might mean not closing between invocations.
Let's chat about this too as either way the code is a bit confusing with close's referenced in the service but not used.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/tokens/SqsDeliveryService.scala
alexduf: Oh yeah I just realised it was handled with a callback. Fine :+1:

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/NotificationParser.scala
markrichardsg: i've swapped to two code paths:
handleShardedNotification and handleChunkedNotification

Currently, handleRequest is supported (for transition purposes) so PROD doesn't need to change any time soon, but will be using the refactored flow for sharded notifications.

This means we don't need `type`

handleEither is there termporarily, but will be removed when second queue and lambda are created

PR #357, opened 2019-02-26
Recursive fetching of SSM parameters

alexduf:
This fetches recursively all the parameters under a namespace. Tested on CODE :heavy_check_mark:

### DIFF
--- a/db-cleaner/src/cleaner.rs
+++ b/db-cleaner/src/cleaner.rs
@@ -36,7 +36,43 @@ fn env_var(key: String, default: String) -> String {
 }
 
 fn fetch_config(credentials: ChainProvider) -> Result<HashMap<String, String>, String> {
-    let mut config = HashMap::new();
+
+    fn recursive_fetch(
+        ssm_client: SsmClient,
+        path: String,
+        mut current_config: HashMap<String, String>,
+        next_token: Option<String>
+    ) -> Result<HashMap<String, String>, String> {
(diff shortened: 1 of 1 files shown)


db-cleaner/src/cleaner.rs
```diff
@@ -36,7 +36,43 @@ fn env_var(key: String, default: String) -> String {
...
+        };
+
+        ssm_client.get_parameters_by_path(req).sync()
+            .map_err(|error| {
+                error!("Error while fetching parameter: {}", error);
+                "Unable to fetch configuration".to_string()
+            })
+            .and_then(|result| {
+                for parameter in result.parameters.unwrap_or(Vec::new()) {
+                    if parameter.name.is_some() && parameter.value.is_some() {
```
markrichardsg: What do you think of?
```
                     match (parameter.name, parameter.value) {
                        (Some(long_name), Some(value)) => {
                            let prefix_length = path.len() + 1; // +1 for the extra slash
                            let name = long_name[prefix_length..].to_string();
                            current_config.insert(name, value.to_string());
                        }
                        _ => ()
                    }
```

alexduf: I think it would work too, but if you look at the diff without the whitespace it might make my changes more obvious as I haven't touched this part:
https://github.com/guardian/mobile-n10n/pull/357/files?utf8=%E2%9C%93&diff=split&w=1

PR #368, opened 2019-03-05
MSS-735: Prepare for restricted newsstand api keys

markrichardsg:
Need to add Newsstand keys in config before deployment.
After that, logs should indicate if there is a misconfiguration, corrected if necessary and then the next step of applying the restriction can be applied.


markrichardsg:
Do not merge until config is created

PR #372, opened 2019-03-07
Mss 751/add harvester

markrichardsg:
The harvester will replace the workers and doesn't have to care about direct delivery, only queueing based on the platform.

A later improvements might be to remove platform from the ShardNotification and db query parameters, but for now it is needed to maintain compatibility with the workers.

Because it handles the same event types as the workers, the deployment process should probably be:

1. Deploy harvester stack
2. Disable workers doing direct delivery
3. Convert workers to harvesters (3 harversers)
4. Direction notifications to the harvester queue
5. Delete workers


notificationworkerlambda/harvester-cfn.yaml
```diff
@@ -0,0 +1,242 @@
...
+        PolicyDocument:
+          Statement:
+            Effect: Allow
+            Action: cloudwatch:PutMetricData
+            Resource: "*"
+
+  LambdaSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: !Sub Lambda to db in ${Stage}
```
alexduf: I don't think the description is right

notificationworkerlambda/riff-raff.yaml
```diff
@@ -7,10 +7,15 @@ deployments:
     app: notificationworkerlambda
     parameters:
       bucket: mobile-notifications-dist
-      functionNames: [ios-notification-worker-, ios-notification-worker-sender-, android-notification-worker-sender-, android-notification-worker-, registration-cleaning-worker-]
+      functionNames: [ios-notification-worker-, ios-notification-worker-sender-, android-notification-worker-sender-, android-notification-worker-, registration-cleaning-worker-, mobile-notifications-harvester-]
       fileName: notificationworkerlambda.jar
       prefixStack: false
     dependencies: [ios-notification-worker-cfn, android-notification-worker-cfn, registration-cleaning-worker-cfn]
```
alexduf: ` , registration-cleaning-worker-cfn]`

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def queueShardedNotification(shardNotifications: Stream[IO, ShardedNotification]): Stream[IO, Unit] = {
+    val chunkedTokenStream = for {
+      n <- shardNotifications
+      platform = n.platform
+      notificationLog = s"(notification: ${n.notification.id} ${n.range})"
+      _ = logger.info(s"Queuing notification $notificationLog...")
+      tokens <- tokenService.tokens(n.notification, n.range, platform).chunkN(1000)
+    } yield ChunkedTokens(n.notification, tokens.toList, platform, n.range)
+    chunkedTokenStream.broadcastTo(firebaseSink, apnsSink)
```
alexduf: This is cool 👍

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+
+  def queueShardedNotification(shardNotifications: Stream[IO, ShardedNotification]): Stream[IO, Unit] = {
+    val chunkedTokenStream = for {
+      n <- shardNotifications
+      platform = n.platform
+      notificationLog = s"(notification: ${n.notification.id} ${n.range})"
+      _ = logger.info(s"Queuing notification $notificationLog...")
+      tokens <- tokenService.tokens(n.notification, n.range, platform).chunkN(1000)
```
alexduf: I think the chunking should happen in the firebase and the apns sink, otherwise we're going to have batches of about 500 tokens each

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/Cloudwatch.scala
```diff
@@ -44,4 +45,21 @@ class CloudwatchImpl extends Cloudwatch {
...
+  }.evalMap { count =>
+    IO.delay {
+      cloudwatchClient.putMetricData(
+        new PutMetricDataRequest()
+          .withNamespace(s"Notifications/$stage/harvester")
+          .withMetricData(Seq(countDatum(
+            "failure",
+            count,
+            new Dimension()
+              .withName("platform")
```
alexduf: 👍 good idea to have the platform as the dimension

notificationworkerlambda/src/test/scala/com/gu/notifications/worker/HarvesterRequestHandlerSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+      apnsSqsDeliveriesCount shouldEqual 0
+    }
+    "Send one iOS breaking news notification" in new WRHSScope {
+      workerRequestHandler.handleHarvesting(sqsEventShardNotification(breakingNewsNotification, iOS), null)
+      tokenStreamCount shouldEqual 1
+      firebaseSqsDeliveriesCount shouldEqual 0
+      apnsSqsDeliveriesCount shouldEqual 1
+
+    }
+    "Queue one iOS content notification" in new WRHSScope {
```
alexduf: Maybe add a test to check the chunking? In it's current form, and with a mixed input (iOS and Android tokens) the chunks will be smaller than 1K, when we expect it to be exactly 1k.

notificationworkerlambda/harvester-cfn.yaml
```diff
@@ -0,0 +1,230 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: Consumes events produced by the notification service and batches notifications for senders to send
```
alexduf: Consumes events produced by the notification service and fetches batches of tokens for senders to send

notificationworkerlambda/harvester-cfn.yaml
```diff
@@ -0,0 +1,230 @@
...
+          Stage: !Ref Stage
+          Stack: !Ref Stack
+          App: !Ref App
+      Description: Consumes shard events and may send notifications or queue them for sending
+      Handler: com.gu.notifications.worker.Harvester::handleHarvesting
+      MemorySize: 3008
+      Role: !GetAtt ExecutionRole.Arn
+      Runtime: java8
+      Timeout: 300
+      ReservedConcurrentExecutions: 100
```
alexduf: 100 => 200 as one lambda handles twice the load

PR #373, opened 2019-03-08
MSS-735: Segregate key responsibilities for Newsstand

markrichardsg:
Newsstand keys only for Newsstand. Other keys, not for Newsstand


NathanielBennett:
Yeah, sensible and clean!

PR #396, opened 2019-03-20
Mss 753/use pre counted topics take 2

NathanielBennett:
OK, this time have deployed on CODE and triple checked parameter store value. Have also rechecked posting notifications to stored and non-stored topics


report/app/report/services/TopicCountService.scala
```diff
@@ -0,0 +1,5 @@
+package report.services
+
+class TopicCountService {
```
alexduf: I think you can remove this file

NathanielBennett: Thanks!

PR #399, opened 2019-03-21
send one request for breaking news instead of four

NathanielBennett:
We want to ensure that breaking news alerts can be targeted to our largest arguments. This will done in the notification microservice, so we need to send the breaking news as one request here


NathanielBennett:
Yeah I was assuming I'd have a facia-branch with an updated dependency which I can test with a notifications branch on CODE. Once that's all good and the notifications microservice is released, I'll create a Facia PR.

PR #400, opened 2019-03-21
Mss 715/fix bintray package name

NathanielBennett:
How did bintray publishing ever work??


NathanielBennett:
Hang on. I don't wanna change the version

PR #403, opened 2019-03-22
MSS-805: Allow harvester to support shards for all platforms

markrichardsg:
Harvester will need to support both types of shard until notification migrates successfully to send for all platforms.

### DIFF
--- a/common/src/main/scala/db/SqlRegistrationRepository.scala
+++ b/common/src/main/scala/db/SqlRegistrationRepository.scala
@@ -12,10 +12,10 @@ import doobie.postgres.sqlstate
 import doobie.Fragments
 import models.{Platform, PlatformCount, TopicCount}
 import play.api.Logger
-import tracking.DynamoNotificationReportRepository
 
 class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
   extends RegistrationRepository[F, Stream] {
+  val logger = Logger(classOf[SqlRegistrationRepository[F]])
   override def findTokens(
     topics: NonEmptyList[String],
     platform: Option[String],
--- a/common/src/main/scala/db/SqlRegistrationRepository.scala
+++ b/common/src/main/scala/db/SqlRegistrationRepository.scala
@@ -16,7 +16,6 @@ import tracking.DynamoNotificationReportRepository
 
 class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
   extends RegistrationRepository[F, Stream] {
-  val logger = Logger(classOf[DynamoNotificationReportRepository])
   override def findTokens(
     topics: NonEmptyList[String],
     platform: Option[String],
--- a/common/src/main/scala/db/SqlRegistrationRepository.scala
+++ b/common/src/main/scala/db/SqlRegistrationRepository.scala
@@ -10,11 +10,13 @@ import cats.data.NonEmptyList
 import doobie.free.connection.ConnectionIO
 import doobie.postgres.sqlstate
 import doobie.Fragments
-import models.{PlatformCount, TopicCount}
+import models.{Platform, PlatformCount, TopicCount}
+import play.api.Logger
+import tracking.DynamoNotificationReportRepository
 
 class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
   extends RegistrationRepository[F, Stream] {
-
+  val logger = Logger(classOf[DynamoNotificationReportRepository])
   override def findTokens(
     topics: NonEmptyList[String],
     platform: Option[String],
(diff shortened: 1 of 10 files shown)


common/src/main/scala/db/SqlRegistrationRepository.scala
```diff
@@ -10,11 +10,13 @@ import cats.data.NonEmptyList
...
 import doobie.Fragments
-import models.{PlatformCount, TopicCount}
+import models.{Platform, PlatformCount, TopicCount}
+import play.api.Logger
+import tracking.DynamoNotificationReportRepository
 
 class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
   extends RegistrationRepository[F, Stream] {
-
+  val logger = Logger(classOf[DynamoNotificationReportRepository])
```
NathanielBennett: You don't appear to be actually using this logger ..

notification/app/notification/services/guardian/GuardianNotificationSender.scala
```diff
@@ -117,8 +117,8 @@ class GuardianNotificationSender(
     }
 
     shard(countWithDefault).map { shard =>
-      ShardedNotification(notification, shard, platform)
-      val shardedNotification = ShardedNotification(notification, shard, platform)
+      ShardedNotification(notification, shard, Some(platform))
```
NathanielBennett: I hadnn't noticed  this before. I don't really get the point of this line. What am I missing? We''re calling apply but doing nothing with the returned object

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -34,9 +34,14 @@ trait HarvesterRequestHandler extends Logging {
     throwables.broadcastTo(logErrors, cloudwatch.sendFailures(env.stage, platform))
   }
 
-  def firebaseSink(notification: Notification, range: ShardRange): Sink[IO, String] = {
+  def firebaseSinkBuilder(notification: Notification, range: ShardRange): Sink[IO, (String, Platform)] = {
```
NathanielBennett: why rename this  method and not the aprns one on L54

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -62,17 +72,16 @@ trait HarvesterRequestHandler extends Logging {
...
       n <- shardNotifications
-      platform = n.platform
-      sink: Sink[IO, String] = platform match {
-        case Android => firebaseSink(n.notification, n.range)
-        case _root_.models.iOS => apnsSink(n.notification, n.range, platform)
-        case Newsstand => apnsSink(n.notification, n.range, platform)
-        case _ => throw new Exception("Unknown platform")
-      }
+      firebaseSink = firebaseSinkBuilder(n.notification, n.range)
+      newsstandSink = apnsSink(n.notification, n.range, Newsstand)
```
NathanielBennett: Don't get why this is hardcoded to Newsstand

NathanielBennett:
Just questions reallly.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
markrichardsg: There was a name conflict with L75

notification/app/notification/services/guardian/GuardianNotificationSender.scala
markrichardsg: Good spot

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
markrichardsg: Apple notifications are either Newsstand or iOS, so there are two sinks that might be used.
For some reason, Newsstand is its own platform (can't remember history).
Maybe this could be revisited to see what value it adds, I'm not sure it's exposed in metrics past this point. Can I raise a ticket for this?

NathanielBennett: Yes  that is a good plan.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
NathanielBennett: Fair enogh

notification/app/notification/services/guardian/GuardianNotificationSender.scala
NathanielBennett: :+1:

NathanielBennett:
:+1: Fix  the tests and  I'll approve  ..

PR #410, opened 2019-04-01
Mss 715/prioritise uk breaking news alerts

NathanielBennett:
Have the api-client send breaking news requests sequentially. 
Need to add la lot of Reads to get the test to work

### DIFF
--- a/api-client/src/main/scala/com/gu/mobile.notifications.client/NextGenApiClient.scala
+++ b/api-client/src/main/scala/com/gu/mobile.notifications.client/NextGenApiClient.scala
@@ -23,13 +23,44 @@ protected class NextGenApiClient(
   private val url = s"$host/push/topic"
 
   override def send(notificationPayload: NotificationPayload)(implicit ec: ExecutionContext): Future[Either[ApiClientError, Unit]] = {
-    val json = Json.stringify(Json.toJson(notificationPayload))
-    postJson(url, json) map {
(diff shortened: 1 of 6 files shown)


api-client/src/main/scala/com/gu/mobile.notifications.client/NextGenApiClient.scala
```diff
@@ -23,13 +24,46 @@ protected class NextGenApiClient(
...
+        responses.collect { case Left(error) => error } match {
+          case Nil => Right(())
+          case onlyOneError :: Nil => Left(onlyOneError)
+          case errors: List[ApiClientError] => Left(UnexpectedApiResponseError(errors.map(_.description).mkString(", ")))
+        }
+      }
+    }
+
+    notificationPayload match {
+      case breakingNews: BreakingNewsPayload if breakingNews.topic.size > 1  => doSendOncePerTopic(breakingNews)
```
alexduf: Any reason to revert that bit to sending the notification multiple time?

NathanielBennett: Yeah I couldn't figure out a way of ensuring they were done sequentially in the notification service. This way, we guarentee the notifications are sent in the correct order ..

NathanielBennett: It's not really a revert. Each future is executed after the other. Which is slightly differen

PR #411, opened 2019-04-10
Simplify logs

alexduf:
This PR simplifies the setup of logs of our lambdas.

I haven't bothered extracting the version numbers as the following PR will factorise the build into one function.


build.sbt
```diff
@@ -326,14 +326,15 @@ lazy val notificationworkerlambda = project
   .enablePlugins(RiffRaffArtifact)
   .settings(
     libraryDependencies ++= Seq(
-      "com.typesafe.play" %% "play-logback" % "2.6.16",
+      "com.amazonaws" % "aws-lambda-java-log4j2" % "1.1.0",
```
NathanielBennett: How come you need the aws log4j dependency for just the worker lambda and not the others?

alexduf: I didn't add it if the dependency was already there :)

NathanielBennett: :+1:

PR #412, opened 2019-04-11
Simplify lambda projects

alexduf:
Rather than copy-pasting our lambda config from one project to another, this extracts the shared config into a function such that the creation of a new lambda shouldn't take us too long.

This is also a first step towards renaming and moving our modules into a more sane hierarchy.

After the refactoring I ran into a deadlock while running the Unit Tests. It seems entirely unrelated and as far as I understand is linked to the way Mockito works. I've user the work around from this issue: https://github.com/sbt/sbt/issues/3022

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,10 @@ val projectVersion = "1.0-latest"
 organization := "com.gu"
 scalaVersion in ThisBuild := "2.12.6"
 
+// this prevents deadlocks when running the tests from the root project.
+// See https://github.com/sbt/sbt/issues/3022
+Global / concurrentRestrictions += Tags.limit(Tags.Test, 1)
+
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -10,6 +10,10 @@ val projectVersion = "1.0-latest"
 organization := "com.gu"
 scalaVersion in ThisBuild := "2.12.6"
 
+// this prevents deadlocks when running the tests from the root project.
+// See https://github.com/sbt/sbt/issues/3022
+Global / concurrentRestrictions += Tags.limit(Tags.Test, 1)
+
```
NathanielBennett: :+1:

NathanielBennett:
:+1: 
This is great!

PR #417, opened 2019-05-01
Simplify our code by removing the platform

alexduf:
The harvester knows how to batch per platform, there's no need to send and batch once per platform anymore


alexduf:
Tested on CODE

PR #419, opened 2019-05-13
Add collapse ID based on article ID or match ID

alexduf:
This theoretically allows Editorial to re-send a notification that would fix a typo or a headline


alexduf:
Tested on CODE.

This also allows the ios team to remove the football notification hack. cc @ddramowicz

PR #428, opened 2019-06-20
Simplify Notification Report.

alexduf:
- doesn't need a version anymore => it was used to manage concurrent writes when counting how many devices received a push notification. This is now re-counted every time by a lambda so no concurrency issue.
 - Because there's no concurrency issue anymore, we don't need to be retry to write the report in the DB
 - the model was duplicated yet serialising and deserialising the same thing (following a previous simplification)


alexduf:
Tested on CODE with a global breaking news. Checked the report endpoint looks good.
IDs:
- d2d4b5a4-39f8-4969-815e-cca23fed1ea7
- 33b6fa24-3895-4865-ad0d-520a04c6fd9f
- becd1835-f257-404d-a740-871a7ccafb12
- c674d06b-d552-4617-849b-34fdcd93ba1c

PR #429, opened 2019-06-21
Remove topic limit

alexduf:
This could have been a one line change but I think I need to be more aggressive on removing tech debt.

I recommend reading commit by commit


alexduf:
tested on CODE with a global alert

PR #432, opened 2019-07-04
Mutable content is always true to trigger the right background process

alexduf:
This is in preparation of the new ios version of the app.

It won't rely on having `content-available: 1` anymore, but will rely on `mutable-content: 1` as a more reliable way to trigger the pre-caching and tracking of push notifications. 🤞


alexduf:
To be merged on Monday the 8th of July

PR #433, opened 2019-07-04
Simplify notification report

alexduf:
This is take two on https://github.com/guardian/mobile-n10n/pull/428

I didn't realise one attribute would have two possible names after that merge. This allows backward compatibility such that existing records in the database can still be read, and new attributes will bear the "platformCounts" name (as exposed by the Rest API).

I also removed the recursive retry that was present in the event consumer lambda as there's no concurrency risk anymore


commoneventconsumer/src/main/scala/com/gu/notifications/events/model/EventAggregation.scala
```diff
@@ -1,11 +1,21 @@
...
 case class EventAggregation(platformCounts: PlatformCount)
 
 object EventAggregation {
-  implicit val jreads = Json.reads[EventAggregation]
+  private val defaultJsonReads = Json.reads[EventAggregation]
+
+  implicit val jreads = new Reads[EventAggregation] {
+    override def reads(json: JsValue): JsResult[EventAggregation] = json match {
+      case JsObject(fields) =>
+        val platformCount = fields.get("platform").orElse(fields.get("platformCounts"))
```
jacobwinch: Do we need to support this indefinitely or just for the next few days / weeks?

alexduf: Indefinitely as the records in the database are already written like that. I just found it strange that the database model and the api model were the same except for that attribute name, adding these 250 lines of code for no good reason.

PR #435, opened 2019-07-08
Remove the content-available flag

alexduf:
From breaking news and content notifications. This should hopefully increase our delivery rate.

we need to wait long enough for the right version of the iOS app to be deployed in the wild.

@ddramowicz : with what version is this going live?


ddramowicz:
https://github.com/guardian/ios-live/pull/4479 is still in review but I expect to get it merged soon which means it should be version 7.20 - i'll let you know when i know for sure

alexduf:
Version currently in beta 7.20

PR #437, opened 2019-07-09
Instantiate two new lambdas with the same code

alexduf:
- Add an environment variable to the lambda, telling it which platform it's handling
- Use this platform to decide which configuration to load. This will need to be refactored later such that each lambda has its own SSM prefix based on the platform. This will keep the number of loaded properties to the minimum
- Remove `platform` from the model that gets serialised on the SQS queue

TODO

- [x] Wire the harvester such that newsstand notifications are sent to the new SQS queue. In its current form this PR would break the daily edition


alexduf:
Deployment:

Apply the cloudformation manually four times: one for each platform
Then create the two missing config keys for the two new workers

PR #439, opened 2019-07-11
Purge breaking news article before sending the breaking news

alexduf:
This is to ensure the article that is about to be read is as fresh as possible.

Editorial noticed in some cases the articles had been edited shortly before the notification was sent, but the version visible on the device was an older cached version. This should fix the issue

### DIFF
--- a/notification/app/notification/services/FastlyPurge.scala
+++ b/notification/app/notification/services/FastlyPurge.scala
@@ -1,0 +1,31 @@
+package notification.services
+
+import play.api.Logger
+import play.api.libs.ws.WSClient
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait FastlyPurge {
+  def softPurge(url: String): Future[Boolean]
+}
+
+class FastlyPurgeImpl(wsClient: WSClient)(implicit ec: ExecutionContext) extends FastlyPurge {
+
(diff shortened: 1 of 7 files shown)


alexduf:
I'm using soft purges, which means if by lack of luck the article returns a 500, Fastly will keep serving the cached version, even after purging.

https://docs.fastly.com/guides/purging/soft-purges

notification/app/notification/services/FastlyPurge.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  private val logger: Logger = Logger(this.getClass)
+
+  def softPurge(url: String): Future[Boolean] = {
+
+    wsClient.url(url)
+      .addHttpHeaders("Fastly-Soft-Purge" -> "1")
+      .execute("PURGE")
+      .map { resp =>
+        println(s"Soft purged $url got HTTP ${resp.status} back")
```
NathanielBennett: Get rid of this?

NathanielBennett:
1 minor thing

notification/app/notification/services/FastlyPurge.scala
alexduf: Woups!

PR #440, opened 2019-07-12
Publish models only rather than a client

alexduf:
The client is not bringing any value anymore, so rather after having a chat with Michael Barton we decided to only publish the models and their serialisers.

This changes the library name, so I put the version back to 1.0.0.


build.sbt
```diff
@@ -339,18 +339,3 @@ lazy val reportExtractor = lambda("reportextractor", "reportextractor", Some("co
   .settings(
     riffRaffArtifactResources += (baseDirectory.value / "cfn.yaml", "reportextractor-cfn/cfn.yaml")
   )
-
-lazy val root = (project in file(".")).
```
jacobwinch: Sorry if I'm being silly, but can you explain why this PR allows us to remove this?

alexduf: It turns out you can completely remove the top level project and sbt will generate it for you so I figured I'd do that. But you're right it's unrelated. I'll revert that part and raise that separately.

I'll start by fixing the build and I'll revert that

jacobwinch: Happy to include that in this PR if you like. I genuinely didn't understand how it was working!

PR #441, opened 2019-07-12
Papercuts

alexduf:
This is a collection of small things, not enough to justify a ticket, but enough to grind my gears :)

Each commit fixes one thing, no new code was written, I only renamed or moved things.


jacobwinch:
> Each commit fixes one thing, no new code was written, I only renamed or moved things.

I really like this approach - it made this super easy to review 👍

PR #443, opened 2019-07-25
Mss 945 editions notification type

NathanielBennett:
Create new notification type for editions

PR #445, opened 2019-08-02
Add editions notification Payload

alexduf:
This is an extension of https://github.com/guardian/mobile-n10n/pull/443 to add the extra fields that the editions app expects.


alexduf:
cc @jimhunty this is the ironing of our solution. It will require to add the fields to the payload when using the curl.

PR #447, opened 2019-08-05
Fix edition topic type

alexduf:
I forgot something 🙄


NathanielBennett:
I didn't notice that :(

PR #450, opened 2019-08-07
Support transactions while registering

alexduf:
This is to cover the very unlikely case where a breaking news is sent while a device re-registers.

There is a short window (~60 ms) where a device is deleted from the database before being re-inserted. A selection during that window would be without the token of that device.

To solve that case I'm introducing transactions here, although I haven't been able to measure the impact on performance yet, and it's going to be hard to measure that impact without being in production.

### DIFF
--- a/common/src/main/scala/db/RegistrationService.scala
+++ b/common/src/main/scala/db/RegistrationService.scala
@@ -13,8 +13,6 @@ import scala.concurrent.ExecutionContext
 
 class RegistrationService[F[_], S[_[_], _]](repository: RegistrationRepository[F, S]) {
   def findByToken(token: String): S[F, Registration] = repository.findByToken(token)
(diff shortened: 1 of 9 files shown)


common/src/main/scala/db/RegistrationService.scala
```diff
@@ -3,23 +3,43 @@ package db
...
-  def save(sub: Registration): F[Int] = repository.save(sub)
-  def remove(sub: Registration): F[Int] = repository.remove(sub)
-  def removeAllByToken(token: String): F[Int] = repository.removeByToken(token)
+
+  def registerDevice(token: String, registrations: List[Registration]): F[Int] = {
+    val prog = for {
+      _ <- repository.deleteByToken(token)
+      inserted <- registrations.map(repository.insert).sequence
+    } yield inserted.sum
+    prog.transact(xa)
```
alexduf: This is the useful transaction. All the other `transact` are pretty much the same as running in autocommit mode

PR #459, opened 2019-09-16
MSS-1015: Trigger purge of an article on breaking news with surrogate key

frankie297:
This allows a soft purge of a breaking news notification using the surrogate key `Item`. It also adds a timeout of 2 seconds.


frankie297:
@alexduf can you see any glaring/obvious reasons why this wouldn't work? am I missing something in my code?

PR #460, opened 2019-09-18
MSS-1139: Add a push type to notifications

frankie297:


frankie297:
@alexduf it's pretty impossible to test this, my original thinking along with @ddramowicz was to set `breakingNewsPayload` for instance to be a background alert and to see if we still received the notification (hoping that we wouldn't). But we think there might be other factors that will play a part in deciding if a notification is sent or not (such as content available)? Do you agree or can you think of another way to test?

frankie297:
@alexduf haha yes sorry, we was just testing some things out. But can confirm iOS 13 has still been receiving notifications so I think it's good.

PR #466, opened 2019-10-03
Force socket timeout to 30 seconds. 

alexduf:
This should dramatically improve RDS incident recovery.

### DIFF
--- a/common/src/main/scala/db/DatabaseConfig.scala
+++ b/common/src/main/scala/db/DatabaseConfig.scala
@@ -1,5 +1,6 @@
 package db
 
+import java.util.Properties
 import java.util.concurrent.Executors
 
 import cats.effect.internals.IOContextShift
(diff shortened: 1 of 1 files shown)


common/src/main/scala/db/DatabaseConfig.scala
```diff
@@ -38,6 +39,11 @@ object DatabaseConfig {
     hikariConfig.setJdbcUrl(config.url)
     hikariConfig.setUsername(config.user)
     hikariConfig.setPassword(config.password)
+
+    val dsProperties: Properties = new Properties()
+    dsProperties.setProperty("socketTimeout", "30")
```
dominickendrick: Could this be achieved with `hikariConfig.addDataSourceProperties("socketTimeout", "30")` instead ? (note the `add` rather than `set`

alexduf: Ha, you're probably right!

PR #472, opened 2019-10-30
Investigate timeouts

alexduf:
This won't solve our issue, but it should inform us better onto why we get timeouts.

It also increases the timeout from 45s to 90s


alexduf:
cc @jacobwinch

PR #477, opened 2019-11-08
Add a new election topic type

frankie297:


jonathonherbert:
@frankie297 just double checking – is this PR still needed?

PR #478, opened 2019-11-12
Add 'election' breaking news topic type to mobile notifications client

jonathonherbert:
Adds a new election topic type to the mobile notifications client, for the benefit of notification emitters (e.g. https://github.com/guardian/facia-tool).


api-models/src/main/scala/com/gu/mobile.notifications.client/models/Topic.scala
```diff
@@ -35,6 +35,7 @@ object Topic {
   val BreakingNewsAu = Topic(Breaking, AU.toString)
   val BreakingNewsInternational = Topic(Breaking, International.toString)
   val BreakingNewsSport = Topic(Breaking, "sport")
+  val BreakingNewsElection = Topic(Breaking, "election")
```
alexduf:
```suggestion
  val BreakingNewsElection = Topic(Breaking, "uk-general-election")
```

jonathonherbert:
Great point @alexduf 👍

PR #484, opened 2019-11-20
Set a default title for Android payload and pass a title to iOS payload

frankie297:
- Reverts guardian/mobile-n10n#483
- Sets a default title for Android to be "The Guardian"
- Pass a title to iOS payload when a topic is available
- Adds "General Election 2019" title to the notification when there is an election topic type


alexduf:
We were too eager, but we need the change from facia tool to be merged at the same time.
Build was intercepted and no deploy happen

PR #489, opened 2020-02-06
Add the appVersion to the registration model

alexduf:
~I think this ticket is a one-liner.~ No it's not 😛 
I might have been overboard when creating the tickets


alexduf:
Tested on CODE ✅

PR #490, opened 2020-02-07
MSS-1384: Store the buildTier into the registration database

jacobwinch:
https://theguardian.atlassian.net/browse/MSS-1384

### DIFF
--- a/common/src/main/scala/db/Registration.scala
+++ b/common/src/main/scala/db/Registration.scala
@@ -6,7 +6,7 @@ import java.sql.Timestamp
 import java.time.LocalDateTime
 
 import doobie.util.Meta
-import models.{DeviceToken, Platform}
+import models.{Android, DeviceToken, Platform}
 
 object Registration {
   implicit val PlatformMeta: Meta[Platform] =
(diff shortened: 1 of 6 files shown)


common/src/main/scala/db/Registration.scala
```diff
@@ -18,10 +18,11 @@ object Registration {
     Meta[Timestamp].timap(ts => ts.toLocalDateTime)(dt => Timestamp.valueOf(dt))
 }
 
-case class Registration(device: Device, topic: Topic, shard: Shard, lastModified: Option[LocalDateTime] = None)
+case class Registration(device: Device, topic: Topic, shard: Shard, lastModified: Option[LocalDateTime] = None, buildTier: Option[BuildTier])
```
alexduf: I'm confused, I can't quite see where the BuildTier model was defined?

common/src/main/scala/db/Registration.scala
```diff
@@ -18,10 +18,11 @@ object Registration {
...
 case class Device(token: String, platform: Platform)
 case class Topic(name: String)
 case class Shard(id: Short)
+case class BuildTier(tier: String)
```
alexduf: Should this be defined as an Enum using `enumeratum`?

PR #494, opened 2020-02-13
Deploy a new android beta worker

alexduf:
To help the firebase migration we need to deploy a new set of SQS/Lambda specifically for the android beta population

- Add a new platform for the android beta population
- Fix the `VisibilityTimeout` to be higher than the lambda `Timeout`
- I created the stack in CODE and PROD


alexduf:
I've also provisioned the reserved concurrency in a more reasonable way.
I set it to almost double the concurrency observed in production.

PR #495, opened 2020-02-20
Add buildTier to the harvester

frankie297:


frankie297:
@alexduf Not tested yet, but wanted to know if I was on the right lines

common/src/main/scala/db/SqlRegistrationRepository.scala
```diff
@@ -87,18 +87,19 @@ class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
       )
       ++ fr"GROUP BY token, platform"
       )
-      .query[(String, String)]
+      .query[(String, String, String)]
```
alexduf: Here it would have to be an Option[String] as there's no guarantee the buildTier will be present in the database. In fact we know already some records won't have it in the database

common/src/main/scala/db/SqlRegistrationRepository.scala
```diff
@@ -87,18 +87,19 @@ class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
...
       .stream
       .transact(xa)
-      .map{ case(token, platformString) => {
+      .map{ case (token, platformString, buildTierString) => {
         val maybePlatform = Platform.fromString(platformString)
         if(maybePlatform.isEmpty) {
           logger.error(s"Unknown platform in db $platformString")
         }
-        (token, maybePlatform)
+        val maybeBuildTier = BuildTier.fromString(buildTierString)
```
alexduf: So here you'll have `buildTierString` of type `Option[String]`, and a function `BuildTier.fromString` which takes a `String` and returns an Option[BuildTier].

How can you reconcile the two such that `maybeBuildTier` is of type `Option[BuildTier]` ?

If that's unclear I suggest you start a worksheet to work on that part

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -40,12 +42,12 @@ trait HarvesterRequestHandler extends Logging {
     throwables.broadcastTo(logErrors, cloudwatch.sendFailures(env.stage, platform))
   }
 
-  def platformSink(shardedNotification: ShardedNotification, platform: Platform, deliveryService: SqsDeliveryService[IO]): Sink[IO, (String, Platform)] = {
-    val iosSinkErrors = sinkErrors(platform)
+  def platformSink(shardedNotification: ShardedNotification, platform: Platform, deliveryService: SqsDeliveryService[IO], buildTier: Option[BuildTier] = None): Sink[IO, HarvestedToken] = {
```
alexduf: It seems you aren't using `buildTier` here, which I believe is a good thing! You probably want to drop the parameter 👍

common/src/main/scala/db/SqlRegistrationRepository.scala
alexduf: So here's an example of setup where you could try to guess the missing part. Don't pay attention to what the code actually does (it's rubbish), but only focus on the types!


```scala

case class BuildTier(s: String)

def buildTierFromString(buildTierString: String): Option[BuildTier] = Some(buildTierString)

val myOptionalString: Option[String] = Some("Hello")

val maybeBuildTier: Option[BuildTier] = <insertSomethingHere>
```

Without touching anything else than what's in the angle brakets `<insertSomethingHere>` can you get `maybeBuildTier` to be the result of calling `buildTierFromString` with the content of `myOptionalString` such that the code compiles?

frankie297:
@alexduf I think I need some help with some failing tests, I don't think my logic is right but can't see where it's going wrong

common/src/main/scala/db/SqlRegistrationRepository.scala
```diff
@@ -87,18 +88,20 @@ class SqlRegistrationRepository[F[_]: Async](xa: Transactor[F])
...
       .stream
       .transact(xa)
-      .map{ case(token, platformString) => {
+      .map{ case (token, platformString, buildTierString) => {
         val maybePlatform = Platform.fromString(platformString)
         if(maybePlatform.isEmpty) {
           logger.error(s"Unknown platform in db $platformString")
         }
-        (token, maybePlatform)
+        val maybeBuildTier: Option[BuildTier] = BuildTier.fromString(buildTierString.toString)
```
alexduf: I'm rooting for you, but that's not quite what we want!
Have a look at what the [scala doc](https://www.scala-lang.org/api/current/scala/Option.html) define as "useful methods" here, and see if there's one that appear to have the type signature you need

frankie297:
Took me a while to get there, thank you for your help @alexduf  😓

notificationworkerlambda/src/test/scala/com/gu/notifications/worker/HarvesterRequestHandlerSpec.scala
```diff
@@ -113,10 +113,10 @@ class HarvesterRequestHandlerSpec extends Specification with Matchers {
         sqsDeliveries
       }
 
-      override val androidBetaDeliveryService: SqsDeliveryService[IO] = (chunkedTokens: ChunkedTokens) => {
-        firebaseSqsDeliveriesCount.incrementAndGet()
-        firebaseSqsDeliveriesTotal.addAndGet(chunkedTokens.tokens.size)
-        sqsDeliveries
+       override val androidBetaDeliveryService: SqsDeliveryService[IO] = (chunkedTokens: ChunkedTokens) => {
```
alexduf: It seems like this part of the code has been reformatted, but it doesn't seem to align with the code above :/

PR #501, opened 2020-03-30
Fix harverster error part 2

alexduf:
While fixing this issue I realised our definition of what is `platform` has reached a confusing state. In fact I believe that confusion is what caused this bug in the first place.

I've created a ticket to address that MSS-1539.

Regarding this PR, I introduce a new concept `WorkerSqs` which points to each of our sender workers.
Then I ensured all the routing logic was encapsulated in a single pattern match (see `routeToSqs`) such that we guarantee that a single message won't match two workers.

✅  Tested on CODE

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
@@ -1,6 +1,6 @@
 package com.gu.notifications.worker
 
-import _root_.models.{Android, AndroidEdition, Ios, IosEdition, Newsstand, Platform, ShardedNotification}
+import _root_.models.{Android, AndroidBeta, AndroidEdition, Ios, IosEdition, Newsstand, Platform, ShardedNotification}
 import cats.effect.{ContextShift, IO, Timer}
 import com.amazonaws.services.lambda.runtime.Context
 import com.amazonaws.services.lambda.runtime.events.SQSEvent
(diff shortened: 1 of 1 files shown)


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -60,18 +68,28 @@ trait HarvesterRequestHandler extends Logging {
 
   }
 
+  def routeToSqs: PartialFunction[HarvestedToken, (WorkerSqs, HarvestedToken)] = {
```
jacobwinch, line 71: This is nice and clear 💯

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -60,18 +68,28 @@ trait HarvesterRequestHandler extends Logging {
...
   def queueShardedNotification(shardedNotifications: Stream[IO, ShardedNotification]): Stream[IO, Unit] = {
     for {
       shardedNotification <- shardedNotifications
-      androidSink = platformSink(shardedNotification, Android, androidLiveDeliveryService)
-      iosSink = platformSink(shardedNotification, Ios, iosLiveDeliveryService)
-      newsstandSink = platformSink(shardedNotification, Newsstand, iosEditionDeliveryService)
-      androidEditionSink = platformSink(shardedNotification, AndroidEdition, androidEditionDeliveryService)
-      iosEditionSink = platformSink(shardedNotification, IosEdition, iosEditionDeliveryService)
+      androidSink = platformSink(shardedNotification, Android, WorkerSqs.AndroidWorkerSqs, androidLiveDeliveryService)
+      androidBetaSink = platformSink(shardedNotification, AndroidBeta, WorkerSqs.AndroidBetaWorkerSqs, androidLiveDeliveryService)
```
jacobwinch, line 83: Should this be `androidBetaDeliveryService`?

alexduf: I'm glad you spotted that one!

alexduf:
Awaiting client side investigation before merging this change. cc @mohammad-haque

PR #502, opened 2020-04-07
Adding a version check to the registration service

alexduf:
Context: this is related to our ongoing problem with our beta users on Android. This should mitigate the issue in the immediate future, and provide a solution once the new beta is released.

All app version bellow build 10000 will be put into the PROD bucket. 

cc @mohammad-haque for your information.

We'll merge that first to put everyone in the PROD bucket, then we'll raise another PR with the correct build number once you've released or are about to release


mohammad-haque:
Thanks, Alex 👍

PR #505, opened 2020-04-30
Play 2.7

alexduf:
Tested in CODE

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,9 +6,8 @@ import sbtassembly.MergeStrategy
 
 val projectVersion = "1.0-latest"
 
-
 organization := "com.gu"
-scalaVersion in ThisBuild := "2.12.6"
+scalaVersion in ThisBuild := "2.12.11"
 
 val compilerOptions = Seq(
   "-deprecation",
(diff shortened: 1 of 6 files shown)


build.sbt
```diff
@@ -22,22 +21,13 @@ val compilerOptions = Seq(
...
 scalacOptions in ThisBuild ++= compilerOptions
 
-val minJacksonVersion: String = "2.10.1"
-val minJacksonLibs = Seq(
-  "com.fasterxml.jackson.core" % "jackson-core" % minJacksonVersion,
-  "com.fasterxml.jackson.core" % "jackson-annotations" % minJacksonVersion,
-  "com.fasterxml.jackson.dataformat" % "jackson-dataformat-cbor" % minJacksonVersion,
-  "com.fasterxml.jackson.core" % "jackson-databind" % minJacksonVersion,
-  "com.fasterxml.jackson.datatype" % "jackson-datatype-jsr310" % minJacksonVersion
-)
```
JamieB-gu, line 32: I assume this was added to force a more recent version of Jackson by making it a direct dependency? This no longer needed because a more recent version of Jackson is now being requested as a transitive dependency?

alexduf: >I assume this was originally added to force a more recent version of Jackson by making it a direct dependency?

Yes

>This is now no longer needed because a more recent version of Jackson is now being requested as a transitive dependency?

It should! I haven't checked. But next  up is Play 2.8, so I recommend we look again then.

alexduf:
To be merged on Monday!

PR #506, opened 2020-05-04
Upgrade doobie

alexduf:
Doobie is our database access library. It relies on Cats and FS2 which are two functional programming libraries.

This Pull Requests focuses on upgrading doobie to the latest version available in order to prepare the project to scala 2.13. This implies upgrading Cats, and indirectly upgrading FS2.

Most of the change are cosmetic (version bump, import change, slight api change)


alexduf:
Tested on CODE ✅

PR #507, opened 2020-05-05
Add OKActions to alarms

frankie297:
Show when alarms are resolved

### DIFF
--- a/fakebreakingnewslambda/fakebreakingnewslambda-cfn.yaml
+++ b/fakebreakingnewslambda/fakebreakingnewslambda-cfn.yaml
@@ -113,6 +113,7 @@ Resources:
     Type: AWS::CloudWatch::Alarm
     Properties:
       AlarmActions: [!Ref AlarmTopic]
+      OKActions: [!Ref AlarmTopic]
       AlarmDescription: !Sub Triggers if not enough Ios dry run notifications are happening on ${Stage}
       ComparisonOperator: LessThanThreshold
       Dimensions:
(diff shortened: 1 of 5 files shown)


frankie297:
@jacobwinch @alexduf would it be safe to say this is fine to action, but don't include the OKActions on the throttle alarms as they appear to be too noisy in letting us know everything is ok?

jacobwinch:
As @alexduf pointed out, the problem with the throttle alarms is that they appear to be regularly transitioning between `INSUFFICIENT_DATA` and `OK` states.

The [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-okactions) note that: 

> The actions to execute when this alarm transitions to the OK state from *any other state*. 

So we should avoid adding `OKActions` to any alarm where `INSUFFICIENT_DATA` is an expected state. Throttling is a good example of this - most of the time there should be 0 throttle events sent to AWS, so in some sense the 'good' state for an alert like this is `INSUFFICIENT_DATA`. 

I'm not 100% sure, but I wonder if explicitly [treating missing data](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/AlarmThatSendsEmail.html#alarms-and-missing-data) as `notBreaching` would help us here.

jacobwinch:
Looking at this more closely, I can see that we are treating missing data as `notBreaching` for the [error-based lambda alerts](https://github.com/guardian/mobile-n10n/blob/master/notificationworkerlambda/sender-worker-cfn.yaml#L188-L203) and these appear to be behaving as we'd like, so I think this might be the fix.

frankie297:
> Looking at this more closely, I can see that we are treating missing data as `notBreaching` for the [error-based lambda alerts](https://github.com/guardian/mobile-n10n/blob/master/notificationworkerlambda/sender-worker-cfn.yaml#L188-L203) and these appear to be behaving as we'd like, so I think this might be the fix.

Yes, I noticed the same thing. I think this is the right approach

PR #512, opened 2020-06-01
Bump Play To 2.8.2

JamieB-gu:
## What does this change?

Upgrades from Play 2.7.4 to Play 2.8.2. Also upgrades `play-json` to 2.8.1, and `play-json-extensions` to 0.42.0.

Two changes have been made to handle deprecated methods:

1. Switched from `schedule` method to `scheduleWithFixedDelay`. This is in the `CloudWatchMetrics` class, and reflects a change in Akka (which was bumped in 2.8). A description of the new method can be found [here](https://doc.akka.io/docs/akka/current/scheduler.html#schedule-periodically). There's an alternative called `scheduleAtFixedRate` which achieves a slightly different outcome; I'm not sure which more accurately reflects our previous behaviour, but the docs recommend the first if unsure.

2. Used `withFallback` over `++` to merge `ApplicationLoader` configurations. By my understanding, our previous behaviour had the `loadedConfig` taking precedence over the `initialConfiguration`, as specified in [the docs](https://www.playframework.com/documentation/2.8.x/api/scala/play/api/Configuration.html#++(other:play.api.Configuration):play.api.Configuration) (if you click on the entry for `++` it gives you a longer description that explains this). Therefore I've flipped the order for `withFallback`, as that produces the same precedence order by my reading of [the docs](https://www.playframework.com/documentation/2.8.x/api/scala/play/api/Configuration.html#withFallback(other:play.api.Configuration):play.api.Configuration) (again, click on the method to get a longer description).

A further change has been made to import a missing implicit from `play-json-extensions`.

**Note:** I've verified that this compiles, and that the tests pass, but nothing more than that, as I'm not very familiar with this project. What other checks ought we to perform @alexduf @jacobwinch?

We're on a more recent Play version, functionality doesn't change, and no bugs are introduced.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ val compilerOptions = Seq(
 
 scalacOptions in ThisBuild ++= compilerOptions
 
-val playJsonVersion = "2.7.4"
+val playJsonVersion = "2.8.1"
 val specsVersion: String = "4.5.1"
 val awsSdkVersion: String = "1.11.772"
 val doobieVersion: String = "0.9.0"
@@ -80,7 +80,7 @@ lazy val common = project
       "com.amazonaws" % "aws-java-sdk-dynamodb" % awsSdkVersion,
       "com.amazonaws" % "aws-java-sdk-cloudwatch" % awsSdkVersion,
       "com.googlecode.concurrentlinkedhashmap" % "concurrentlinkedhashmap-lru" % "1.4.2",
-      "ai.x" %% "play-json-extensions" % "0.40.2",
+      "ai.x" %% "play-json-extensions" % "0.42.0",
       "org.tpolecat" %% "doobie-core"      % doobieVersion,
       "org.tpolecat" %% "doobie-hikari"    % doobieVersion,
       "org.tpolecat" %% "doobie-postgres"  % doobieVersion,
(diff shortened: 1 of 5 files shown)


alexduf:
very nice!

You also need to update play Json to 2.8.x here: https://github.com/guardian/mobile-n10n/blob/master/build.sbt#L22

JamieB-gu:
> You also need to update play Json to 2.8.x here

Hmm, I looked into this, but my understanding was that [as of Play 2.6](https://www.playframework.com/documentation/2.8.x/Migration26#Play-JSON-moved-to-separate-project) `play-json` updates independently/out-of-sync with the core Play library? For example, there's a Play `2.8.2`, but `play-json` only goes up to `2.8.1` currently 🤔

alexduf:
As far as I understand, play-json is indeed its own project with its own release cycle, but play 2.8 will still use play-json 2.8.x, so it's a good idea to keep them lined up.

Here's the play-json version that play 2.8 uses:
https://github.com/playframework/playframework/blob/bc1badd22f36bcfe09517be37a9ef66559b53b31/project/Dependencies.scala#L15

JamieB-gu:
Ok, I've bumped that and updated the description. Again, it compiles and the tests pass, but I don't have any proof it works beyond that 🤔

PR #519, opened 2020-10-13
Add logging for bad requests (to registration)

jacobwinch:
Adds logging for bad requests to help @mohammad-haque debug an Editions problem.

### DIFF
--- a/registration/app/registration/CustomErrorHandler.scala
+++ b/registration/app/registration/CustomErrorHandler.scala
@@ -14,8 +14,8 @@ import scala.concurrent._
 class CustomErrorHandler(env: Environment, config: Configuration, sourceMapper: Option[SourceMapper], router: => Option[Router]) extends DefaultHttpErrorHandler(env, config, sourceMapper, router) {
   private val logger: Logger = LoggerFactory.getLogger(this.getClass)
   override def onBadRequest(request: RequestHeader, message: String): Future[Result] = {
-    val debugInfo = request.headers.toSimpleMap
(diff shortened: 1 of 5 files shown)


registration/app/registration/CustomErrorHandler.scala
```diff
@@ -0,0 +1,20 @@
...
+import play.api.mvc._
+import play.api.routing.Router
+import play.core.SourceMapper
+
+import scala.concurrent._
+
+class CustomErrorHandler(env: Environment, config: Configuration, sourceMapper: Option[SourceMapper], router: => Option[Router]) extends DefaultHttpErrorHandler(env, config, sourceMapper, router) {
+  private val logger: Logger = LoggerFactory.getLogger(this.getClass)
+  override def onBadRequest(request: RequestHeader, message: String): Future[Result] = {
+    val debugInfo = request.headers.get("User-Agent")
```
mohammad-haque, line 16: Can we also log the request payload? It will help find the root cause quickly (maybe we can turn that off later if this seems like excessive logging)

registration/app/registration/CustomErrorHandler.scala
```diff
@@ -0,0 +1,20 @@
...
+    logger.error(s"Bad request due to $message. User agent = $debugInfo")
```
mohammad-haque, line 17: Can we also log the request payload? It will help find the root cause quickly (maybe we can turn that off later if this seems like excessive logging)

PR #522, opened 2020-11-03
Send silent updates for US Election notification

jacobwinch:
Updates to this notification will happen frequently throughout the night, so we want these notifications to be silent.


davidfurey:
This seems like a reasonable solution once the tests pass

jacobwinch:
Now that we have stopped sending such frequent updates this PR may no longer be required - I'll leave it open in case somebody asks for this during the night.

PR #535, opened 2021-03-05
Improve debugging I/O errors on the database

frankie297:
## What does this change?
Add logs to SQL statement and harvester to help us debug errors we see with database connections.


frankie297:
In favour of #536

PR #556, opened 2021-05-12
LIVE-2383: Move away from bintray, and rely on AWS SDK V2 for configuration

frankie297:
## What does this change?
This PR is doing a number of things:

1.  We remove bintray from our resolvers and migrate to sonatype
2. We update dependency versions of PA Client and Logstash Encoder
3. We update a dependency on Simple Configuration. The updated version included a need to depend on AWS SDK V2, so we are pulling this only for what we need to use (what simple configuration relies on)
4. In now depending on AWS SDK V2 it has meant we need to increase our lambda memory, which this PR does
5. When deploying in riff-raff there were issues around the size of the jars that are now created (too big for lambdas), so we excluding Play as a dependency from our lambdas, except the ones that need to rely on Play Json which I have included explicitly. 

## How to test
I have tested every service and lambda in CODE and everything deploys and runs successfully. I have also tested breaking news notifications and they were sent successfully to both iOS and Android.

We can deploy the project again, and we get no alarms. After merging there should be a period of monitoring logs and cloudwatch to check everything is ok.

Paired with @marjisound and @JamieB-gu 

With special thanks to @philmcmahon @sihil @adamnfish @johnduffell


johnduffell (approved):
well done working all that out, it was not a simple problem to deal with.  There's not a lot of precedent and experience of jar trimming and analysis in the department (now that mario G has left), I'm not sure if there's anything worth knowledge sharing?
👍

frankie297:
> well done working all that out, it was not a simple problem to deal with. There's not a lot of precedent and experience of jar trimming and analysis in the department (now that mario G has left), I'm not sure if there's anything worth knowledge sharing?
> 👍

Thanks! It was tough 😓 I think there is still room for improvement so inclined to share knowledge when we have thought about it a bit further, and in further PRs. But for the sake of deadlines and getting our project back into a deployable state again this is probably fine for now 😄

PR #560, opened 2021-07-13
LIVE-2059 - notification - Set associatePublicIpAddress to false

marjisound:
## What does this change?
This PR removes the public IPv4 address from the EC2 instances. This change would resolve the relevant security issue in AWS security Hub.

Find more info in https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#ec2-9-remediation

Deployed and tested in code

### DIFF
--- a/notification/conf/notification.yaml
+++ b/notification/conf/notification.yaml
@@ -325,7 +325,7 @@ Resources:
   NotificationLaunchConfig:
     Type: AWS::AutoScaling::LaunchConfiguration
     Properties:
-      AssociatePublicIpAddress: True
+      AssociatePublicIpAddress: false
       IamInstanceProfile: !Ref NotificationInstanceProfile
       ImageId: !Ref AMI
       InstanceType: !FindInMap [StageVariables, !Ref Stage, InstanceType]


JamieB-gu (approved):

notification/conf/notification.yaml
```diff
@@ -325,7 +325,7 @@ Resources:
   NotificationLaunchConfig:
     Type: AWS::AutoScaling::LaunchConfiguration
     Properties:
-      AssociatePublicIpAddress: True
+      AssociatePublicIpAddress: false
```
frankie297, line 328: I think @JamieB-gu mentioned in other PR reviews but wanted to flag here too, the titlecase on True and lowercase false

marjisound: Good point. The reason I made it lower case is that throughout all our cloudformations for boolean we are using lowercase `true` & `false` so to make it consistent I made this lowercase. 

Also, in AWS cloud formation examples I only found lowercase. Although, both lowercase and uppercase are accepted when you run update the stack, I think lowercase is the lowercase is most common :)

PR #561, opened 2021-07-13
LIVE-2061 - registration - Set AssociatePublicIpAddress to false

marjisound:
## What does this change?
This PR removes the public IPv4 address from the EC2 instances. This change would resolve the relevant security issue in AWS security Hub.

Find more info in https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#ec2-9-remediation

Deployed and tested in code

### DIFF
--- a/registration/conf/registration.yaml
+++ b/registration/conf/registration.yaml
@@ -277,7 +277,7 @@ Resources:
   RegistrationLaunchConfig:
     Type: AWS::AutoScaling::LaunchConfiguration
     Properties:
-      AssociatePublicIpAddress: True
+      AssociatePublicIpAddress: false
       IamInstanceProfile: !Ref DistributionInstanceProfile
       ImageId: !Ref AMI
       InstanceType: !FindInMap [StageVariables, !Ref Stage, InstanceType]


registration/conf/registration.yaml
```diff
@@ -277,7 +277,7 @@ Resources:
   RegistrationLaunchConfig:
     Type: AWS::AutoScaling::LaunchConfiguration
     Properties:
-      AssociatePublicIpAddress: True
+      AssociatePublicIpAddress: false
```
JamieB-gu, line 280: Why is `True` titlecase and `false` lowercase?

JamieB-gu (approved):

registration/conf/registration.yaml
marjisound: The reason I made it lower case is that throughout all our cloudformations for boolean we are using lowercase `true` & `false` so to make it consistent I made this lowercase as well

Also, in AWS cloud formation examples I found, for boolean lowercase `true` & `false` is used. Although, both lowercase and uppercase are accepted when you update the stack, I think lowercase is the most common

PR #565, opened 2021-09-14
Convert our Notification Lambdas to use Container Images

paulmr:
Currently our notification Lambdas are skirting very close to the limit of 250MB available for an S3 based Lambda (in fact strictly speaking we are over the limit at somewhere around 300MB but there must be some buffer). As a result, a change in our dependencies risks pushing us over the edge and making it so we can no longer deploy our Lambdas.

An alternative would be to use container images for our Lambdas instead:

> When you create a Lambda function, you use a deployment package to deploy your function code. Lambda supports two types of deployment packages: .zip file archives and container images.
>
> You can use the Lambda console and the Lambda API to create a function defined as a container image, update and test the image code, and configure other function settings.

(from https://docs.aws.amazon.com/lambda/latest/dg/configuration-images.html#configuration-images-optimization)

According to the docs this should give us a size limit of 10GB, which is plenty of room to grow.

I think it will offer us two other advantages in terms of future work:

* This is a pre-requisite for using ECS instead of Lambdas for this or other services, and so this brings us one step closer to that.
* It may help us to streamline our Lambdas going forward by being able to combine multiple code packages into one docker image with potentially different endpoints.

For the time being however, everything is still done using Lambdas, and we are just converting over to storing the code as a docker image. This means that we don’t need to make any significant modifications to our code or to our cloudformation setup.

After this PR is merged, the deployment process should be the same as before from our point of view: TeamCity will build the code using SBT as before, and then if the build is successful it will push the resulting docker image to our Elastic Container Repository (also created as part of this PR). These images are tagged with the teamcity build number, so we are then able to deploy the build as before using RiffRaff, and it will know to pick up the image that is associated with the build number.

There's a few moving parts to this PR, but essentially it consists of:

* Changing the SBT build (within the file `build.sbt`) so that it creates a Docker image. This is all done within the build.sbt file using the [sbt-native-packager](https://www.scala-sbt.org/sbt-native-packager/) plugin and its Docker image support.

  * The resulting image contains the same deliverables as before, as they are copied over from the file system into the docker image as part of the build

  * The base image for the docker build is the AWS Lambda image, which means that our code sees exactly the same runtime environment as it would do if it was running as a JVM lambda (which is what it is currently doing).

* Within the `Dockerfile` (which is generated within the SBT build) the `CMD` parameter tells the lambda runtime environment which handler function should be called for this instance of the lambda. This is the part that is overwritten in the cloud formation for each instance of the lambda, to implement different behaviour for, e.g. Android vs iOS. This is analagous to the existing handler function configuration variable that is associated with a traditional Lambda.

* We create an Elastic Container Registry (ECR) repository which stores the docker images that are produced by the build step above. This ECR repository is defined in a new cloudformation file, ecr.yaml and the deploy tools root account is given write access to this repository, which allows Team City to upload the images after a successful build.

* The cloudformation which defines the lambdas (described in `sender-worker-cfn.yaml`) sets the `PackageType` for the Lambdas to Image, which is how we tell AWS that these lambdas are defined in a docker image. We point to the above ECR repository so that it knows where to find these images. The actual image to use is defined by the `BuildId` parameter for the CloudFormation stack, and this is set by RiffRaff during the deployment.

* The TeamCity build configuration is modified so that it runs the newly added `deploy.sh` script after succesfully building the image. This script will login to the ECR repository using `docker login` and the credentials that are available in team city, and it will then upload the image, tagging it with the build number.

  * Note that this change to the TeamCity settings really needs to be done after this PR is merged, otherwise it would cause the build to fail because it couldn't find the `deploy.sh` script. Currently it is wrapped in a condition so that it is only applied to this dev branch:

image

There is at least one gotcha with deploying this: annoyingly you cannot change the `ImageType` of an existing Lambda, you have to destroy it and recreate it. This isn’t a problem as they are stateless, but it is annoying, as there are quite a few lambdas involved. In CODE I dealt this this by temporarily changing the name of the lambdas, deploying the cloudformation, then changing the name back again and redeploying it. Assuming there isn't a notification in progress then this shouldn't be observable by any users.

Another disadvantage is that it takes slightly longer to start up a lambda that uses a container image. However, I think this is reasonable, my testing suggests we are still up and running within a minute or so of triggering the notification alert in the Fronts tool, so I think this is still within the area of expected behaviour from this tool. But I am open to hearing otherwise!

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -344,16 +344,26 @@ lazy val lambdaDockerCommands = dockerCommands := Seq(
 
 lazy val buildNumber = sys.env.get("BUILD_NUMBER").orElse(Some("DEV"))
 
+lazy val ecrRepositorySettings =
+  sys.env.get("NOTIFICATION_LAMBDA_REPOSITORY_URL") match {
+    case Some(url) =>
+      val Array(repo, name) = url.split("/", 2)
+      Seq(
+        dockerRepository := Some(repo),
+        packageName in Docker := name
+      )
+    case None => Nil
+  }
+
 lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notificationworkerlambda", Some("com.gu.notifications.worker.TopicCounterLocalRun"))
   .dependsOn(common)
   .enablePlugins(DockerPlugin)
   .enablePlugins(JavaAppPackaging)
+  .settings(ecrRepositorySettings: _*)
   .settings(
     lambdaDockerCommands,
-    dockerRepository := Some("201359054765.dkr.ecr.eu-west-1.amazonaws.com"),
     dockerExposedPorts := Seq(9000), // exposed by the lambda runtime api inside the image
-    packageName in Docker := "notificationworkerlambda-image",
-    dockerAlias := DockerAlias(registryHost = dockerRepository.value, username = None, name = "notificationworkerlambda-image", tag = buildNumber),
+    dockerAlias := DockerAlias(registryHost = dockerRepository.value, username = None, name = (packageName in Docker).value, tag = buildNumber),
     libraryDependencies ++= Seq(
       "com.turo" % "pushy" % "0.13.10",
       "com.google.firebase" % "firebase-admin" % "6.3.0",
--- a/build.sbt
+++ b/build.sbt
@@ -3,6 +3,7 @@ import play.sbt.PlayImport.specs2
 import sbt.Keys.{libraryDependencies, mainClass}
 import sbtassembly.AssemblyPlugin.autoImport.{assemblyJarName, assemblyMergeStrategy}
 import sbtassembly.MergeStrategy
+import com.typesafe.sbt.packager.docker.{Cmd, ExecCmd}
 
 val projectVersion = "1.0-latest"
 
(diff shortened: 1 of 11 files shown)


marjisound (approved):
Well done on your great work 🎉👏

frankie297 (approved):

frankie297:
🎉 great work

paulmr:
Having deployed this a couple of times now on `CODE` now, I think it is probably worth document the steps here before deploying to `PROD` to make sure we don't miss anything &mdash; there are a few gotchas.

* [ ] Merge into master
  * this should hopefully be straightforward
* [ ] Deploy the ECR repository couldformation (from `ecr.yaml`)
  * this isn't currently deployed by riffraff, mostly because it will probably only be deployed once, and so it didn't seem worth it, but perhaps we should fix that.
* [ ] modify the teamcity build configuration so that it builds the docker image for all branches
  - we may choose to make this conditional on the presence of the `deploy.sh` script perhaps, although once this is merged, it should always be present anyway
* [ ] retrigger the build, this should then result in an image being delivered to the ECR for the build

At this point, we are then ready to deploy the Lambdas. However, there is a limitation for Lambdas, which is that we are not allowed to change the `PackageType` of an existing Lambda Function, we have to create a new one. So if we try to deploy the CloudFormation now, as is, the deployment will fail saying that we cannot make this change.

I really struggled to come up with a good solution to this. During the `CODE` deployments, the best I could come up with was to either delete all of the relavent Lambdas prior to the deploy, or to rename all the Lambdas in the cloudformation, deploy (which cause the old Lambdas to be deleted) and undo the name change and redeploy.

PR #566, opened 2021-09-30
fixing cloudwatch logs permissions

paulmr:
A sneaky use of explicit Lambda names was hiding in the cloudformation, meaning that the policy which gives the notification lambdas permission to write to the cloudwatch logs was not being applied after PR#565.

This PR fixes this and also makes the policy less specific which I think is a reasonable comporomise, and it means that if the names change again for some reason we won't have this problem again.


frankie297 (approved):
👍

marjisound (approved):

PR #567, opened 2021-09-30
alarms need to be updated for new lambda names

paulmr:
As discussed on the call, this PR updates the alarms to use the new lambda names. In fact, I have updated them to use `!Ref`s to the actual lambda so they shouldn't be dependent on the name any more.

This is currently deploy on `CODE` and all seems well:

image


marjisound (approved):
LGTM

frankie297 (approved):

PR #569, opened 2021-11-02
Fix eventconsumer athena logging

davidfurey:
The athena lambda cannot currently log to cloudwatch because stack has been added as a prefix to the function name but the corresponding policy not updated.

This PR also removes 2 resources that are no longer needed because the corresponding lambdas have been deleted.


frankie297 (approved):
👍

davidfurey:
Logs are now visible in cloudwatch logs for this lambda.

PR #574, opened 2021-12-14
Update sbt version to 1.5.6 to fix security vulnerability

frankie297:
## What does this change?

Updates the sbt version to 1.5.6 to fix a security vulnerability introduced through log4j

## How to test

run sbt and test locally


marjisound (approved):
Great work 👍

PR #578, opened 2021-12-20
Upgrade mobile-logstash-encoder version

umguardian:
## What does this change?

The `mobile-logstash-encoder` library has been released with a few enhancements / bug fixes. Updating n10n to use that upgraded library.

## How to test

Deploy branch in CODE; log lines should continue to include EC2 instance ID.


groakland (approved):
LGTM

PR #580, opened 2022-01-12
Convert the notification worker lambda cloudformation to CDK

paulmr:
## Introduction

The aim of this pull request is to reimplement the Cloudformation configuration for the notification lambdas using CDK.

The key motivation behind this is for ticket [LIVE-3406](https://theguardian.atlassian.net/jira/software/c/projects/LIVE/boards/36?modal=detail&selectedIssue=LIVE-3406), which describes an issue with intermittent failures to deploy this project.

We think the reason for this failure is a race condition caused by the large number of deploys we are doing within this riff-raff project, combined with the slightly longer duration of those deploys. There is more detail on that specific problem in the ticket.

Anyway, the reason we are doing so many deploys is really just because the 5 or so notification lambdas that we are creating and which are pretty much identical, except for a few minor changes such as the handler function that is called (e.g. one of them will send messages to Apple, the other to Google etc) and we just want to reuse 90% of the cloudformation defined in `sender-worker-cfn.yaml`. We achieve this by setting up separate deploys in the riff-raff config which deploys the same template multiple times:

``` yaml
  mobile-notifications-ios-worker-cfn:
    type: cloud-formation
    app: ios-worker
    parameters:
      prependStackToCloudFormationStackName: false
      cloudFormationStackName: ios-worker-cfn
      templatePath: sender-worker-cfn.yaml
  mobile-notifications-ios-edition-worker-cfn:
    type: cloud-formation
    app: ios-edition-worker
    parameters:
      prependStackToCloudFormationStackName: false
      cloudFormationStackName: ios-edition-worker-cfn
      templatePath: sender-worker-cfn.yaml
# ... etc etc
```

So we need to find an approach that avoids these multiple deploys in riffraff, but which still allows us to resue the cloudformation for each notification lambda &mdash; which seems like a good fit for CDK.

## Interesting details

Most of this code change is just a straight-forward implementation of the existing `sender-worker-cfn.yml` in Typescript-based CDK, using the Guardian's CDK tools.

There are, however, a few things to note.

### Harvester Paramstore Parameters

As can be seen in the architecture diagram at the root of this repo, the notification lambdas are triggered by Harvester which writes the notifications onto the approriate queue. Therefore, Harvester needs to know the location of these queues for each of the lambdas.

As these queues are generated automatically by the cloudformation, we don't know in advance what these ARNs and URLs will be, and so we have to set up some parameters within the parameter store which are read by harvester to find the location.

These were presumably just inserted by hand when the notification architecture was initially set up, and so whenever there is a change to these ARNs we would need to know that we need to update these parameters so that harvester knows where the new locations are.

However, it would be better if these system parameters were updated automatically, and the connection between them was documented within the repo. Therefore, these system parameters are now created within the notification cloudformation, and automatically populated with the correct values:

``` typescript
    // this advertises the name of the sender queue to the harvester app
    new ssm.StringParameter(this, 'SenderQueueSSMParameter', {
      parameterName: `/notifications/${scope.stage}/workers/harvester/${opts.paramPrefix}LiveSqsCdkUrl`,
      simpleName: false,
      stringValue: this.senderSqs.queueUrl,
      tier: ssm.ParameterTier.STANDARD,
      dataType: ssm.ParameterDataType.TEXT
    })
```

The only snag is that we can't create these parameters in the cloudformation when they already exist outside of it, and so I have had to make a small tweak to the name of the parameter so that they don't clash (adding `Cdk`) and there has been a corresponding change in the Harvester code to account for this:

``` scala
// notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Configuration.scala
  def fetchHarvester(): HarvesterConfiguration = {
    val config = fetchConfiguration("harvester")
    HarvesterConfiguration(
      jdbcConfig = jdbcConfig(config),
      iosLiveSqsUrl = config.getString("iosLiveSqsCdkUrl"),
      iosEditionSqsUrl = config.getString("iosEditionSqsUrl"),
      androidLiveSqsUrl = config.getString("androidLiveSqsCdkUrl"),
      androidEditionSqsUrl = config.getString("androidEditionSqsUrl"),
      androidBetaSqsUrl = config.getString("androidBetaSqsUrl")
    )
  }
```

Once we are happy with this deployment, and we tear down the previous non-CDK stack, we can consider renaming these and redeploying (which would be a follow-up PR).

### Exported Values

There is much more detail about this in the README that is being added to the `cdk` directory for this PR, but essentially another connection with Harvester is the IAM policy that is set up there: Harvester needs to be given permission to write to these SQS streams, so currently this is a parameter in the cloudformation that has to be copy-and-pasted in when the cloudformation is deployed.

Instead of having this hidden dependency again, this fix changes it so that this value is now exported from the stack, and then imported by harvester. This can be seen (currently only in CODE where this deployed) either in the console or with a command such as this:

``` shell
$ aws cloudformation list-exports | jq -r '.Exports | map(select(.Name == "NotificationSenderWorkerQueueArns-CODE")) | .[0].Value'
arn:aws:sqs:eu-west-1:[...]
```

This means that:

+ The dependency is now automatic, i.e. you can't forget to update it
+ The dependency is documented in the repo by being included in the source code
+ AWS will bug you if you change it because of this dependency (and you will then need to follow the instructions in the README)


frankie297 (approved):
👍  Great Work

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Configuration.scala
```diff
@@ -68,11 +68,18 @@ object Configuration {
     val config = fetchConfiguration("harvester")
     HarvesterConfiguration(
       jdbcConfig = jdbcConfig(config),
+      // currently this will ignore the CDK stack. To change to CDK, we should
+      // comment the next 5 lines, and uncomment the 5 lines following them.
```
frankie297, line 72: Will this be in it's own PR then?

notificationworkerlambda/ecr.yaml
```diff
@@ -1,10 +1,15 @@
...
+  RepositoryName:
+    Description: Name of the ECR repository
+    Type: String
+    Default: notificationworker-lambda-images
 Resources:
   NotificationLambdaRepository:
     Type: AWS::ECR::Repository
     Properties:
-      RepositoryName: notificationworker-lambda-images
+      RepositoryName: !Ref RepositoryName
```
frankie297, line 12: I imagine this is fine as you've tested, but is it ok to call both the property and the parameter exactly the same?

paulmr: good point, I hadn't even noticed, but as you suggested, it works in testing, so I guess the answer is yes, it is ok :-)

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Configuration.scala
paulmr: Yeah, I will immediately follow up with a new PR once everything looks good.

paulmr:
There a few different steps for deploying this. The main reason for that is that we are creating a new stack, and so it needs to initially be created manuall, adding the appropriate tags, before it can be deployed with riff-raff.

In addition, I also want to take a bit of time to double-check things before it is activated. 

* [ ] upload the new CDK stack from the generated cloudformation, making sure it has the relavent tags
* [ ] merge to master - which will trigger an automated deploy
* [ ] check that the system params are still pointing to the correct place
* [ ] check that the exec role perms includes both of the sets of ARNS
* [ ] wait for a notification to go out and confirm that it ran correctly on the old stack's lambdas
     - https://dashboard.ophan.co.uk/notifications

PR #581, opened 2022-01-19
Wire up the CDK that was introduced in PR #580

paulmr:
PR #580 introduced a CDK version of the n10n sender lambdas, but in order to deploy it with an easy backout route, it was merged and deployed but not activated (i.e. Harvester is still using the old SQS queues which exist in parallel).

Everything looks like it went smoothly (and there have been numerous notifications sent out since it was deployed), so I think we are good to do the switchover now. 

As such, this change will activate the new CDK stack, after which time, once everything is confirmed to be working, we can tear down the old stack. This change is currently deployed on CODE and I have tested that it works and that it causes the notifications to switch over to the new stack.

### DIFF
--- a/notificationworkerlambda/cdk/lib/senderworker.ts
+++ b/notificationworkerlambda/cdk/lib/senderworker.ts
@@ -161,7 +161,7 @@ class SenderWorker extends cdk.Construct {
 
     // this advertises the name of the sender queue to the harvester app
     new ssm.StringParameter(this, 'SenderQueueSSMParameter', {
-      parameterName: `/notifications/${scope.stage}/workers/harvester/${opts.paramPrefix}LiveSqsCdkUrl`,
+      parameterName: `/notifications/${scope.stage}/workers/harvester/${opts.paramPrefix}SqsCdkUrl`,
       simpleName: false,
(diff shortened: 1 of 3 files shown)


notificationworkerlambda/cdk/lib/senderworker.ts
```diff
@@ -241,8 +241,8 @@ export class SenderWorkerStack extends GuStack {
      * platform or app by talking to a different lambda handler function
      */
 
-    addWorker("ios", "ios", "com.gu.notifications.worker.IOSSender::handleChunkTokens")
-    addWorker("android", "android", "com.gu.notifications.worker.AndroidSender::handleChunkTokens")
```
paulmr, line 245: I did actually notice a bug here that I am fixing with this change, the editions params don't have the word `Live` in them so it is not a common suffix for all of them.

paulmr:
This has now been deployed, and any future notifications should be sent via the new CDK stack. Will confirm that all is well once a PROD notifcation has been sent.

paulmr:
A number of notifications have now been sent since this was deployed. The numbers look good, and I can see that it was the new lambdas which were correctly invoked:

image

image

image

PR #582, opened 2022-02-01
replace the accidentally removed cleaning worker

paulmr:
When updating the `riffraff.yaml` for PR #580 , we accidentally removed one of the lambdas from the riffraff file which was still being used.

The effect of this is only that it has not been being redeployed, it wasn't actually deleted from AWS.

But we do need to reinsert this missing line in case we want to update its code later on.

I am also updating the description as it doesn't seem to match what this lambda does.


groakland (approved):
LGTM

PR #586, opened 2022-03-31
Add snapshot and testing tools for cdk

frankie297:
## What does this change?
- Add jest for testing
- Add test for cloudformation
- Create a snapshot file to test against
- Add a github action for running against the cdk directory


.github/workflows/cdk-test.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  cdk-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "16"
```
jacobwinch, line 28: Minor: it might be worth adding a `.nvmrc` file* and [providing the `node-version-file` input](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file). 

This would help to ensure that developers (when running this locally) and CI use the same version of node.

*Personally I would put this at the project root - do you think that makes sense in this context @JamieB-gu?

.github/workflows/cdk-test.yml
```diff
@@ -0,0 +1,36 @@
+name: cdk test
```
jacobwinch, line 1: Generally I prefer GHA but I note that the rest of this project is using TeamCity for CI. Using GHA therefore has a couple of drawbacks in this context:

1. It means developers need to think about 2 CI platforms when working on a single repo
2. It means that you can deploy a change with Riff-Raff _even when this test is failing_

You could probably make this a [required status check](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#require-status-checks-before-merging) as a workaround for 2.

.github/workflows/cdk-test.yml
JamieB-gu: > Minor: it might be worth adding a .nvmrc file* and [providing the node-version-file input](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file).

I hadn't realised they'd added this to the action! Yes, this sounds like a great idea.

> Personally I would put this at the project root - do you think that makes sense in this context

👍. I _think_ nvm will check parent directories for an `.nvmrc`, so even if you run `nvm use` from inside the `cdk` directory it _should_ still work? @frankie297 it might be worth testing this though.

frankie297: 👍

.github/workflows/cdk-test.yml
frankie297: 👍 added cdk-test as a required status check

.github/workflows/cdk-test.yml
frankie297: Added a .nvmrc file at the root of the project and tested

DavidLawes (approved):
We paired on this and it all made sense to me :)

PR #600, opened 2022-05-17
Snyk does not support the new addDependencyTreePlugin annotation

waisingyiu:
## What does this change?

The snyk execution in the teamcity build failed because it does not support the new addDependencyTreePlugin annotation in `plugins.sbt`.  This PR is to change the way we added the dependency tree plugin to the project as instructed by Snyk.

## How to test

We started the build `mobile-notifications-snyk` in teamcity and it was successful.


DavidLawes (approved):
👍

PR #602, opened 2022-05-24
Resolve snyk issues by bumping jackson modules to latest version

DavidLawes:
## What does this change?

Bumping the jackson dependencies to resolve high severity security alerts from snyk.

In MAPI we found that we had errors if not all jackson dependencies were on the same version, hence why we're bumping all 3 jackson dependencies here in one go.


waisingyiu (approved):
LGTM

PR #605, opened 2022-05-26
Update firebase admin, make google oauth a direct dependency

frankie297:
## What does this change?
To fix security vulnerabilities introduced in the notification worker lambda we:
- Updated firebase admin version to the latest (8.1.0)
- Relied on google oauth client as a direct dependency to fix the transitive security vulnerability

## How to test
Deploy to CODE and send a test notification, making sure it works

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -26,5 +26,3 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.0")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.9.4")
 
 addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
-
-addDependencyTreePlugin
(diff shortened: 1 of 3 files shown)


frankie297:
Tested on CODE and it worked

project/plugins.sbt
```diff
@@ -26,3 +26,5 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.0")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.9.4")
 
 addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
+
+addDependencyTreePlugin
```
DavidLawes, line 30: Seemed to need this in order to successfully use `dependencyBrowseTree`

waisingyiu: We removed `addDependencyTreePlugin` in this PR #600 because snyk does not support it.

PR #610, opened 2022-06-09
Bump netty io libraries

waisingyiu:
## What does this change?

This PR bumps the netty IO libraries versions to fix several vulnerabilities found by Snyk.  We set the version explicitly in the subproject `common` as it has transitive dependency on netty IO.  As a result, we removed the explicit dependency declaration in `notificationworkerlambda` as it inherits the same dependencies from `common`.

## How to test

Snyk local testing confirmed the vulnerabilities have been resolved. 
It passed all the test cases.

Paired with @DavidLawes in snyk mobbing session.


DavidLawes (approved):
👍

PR #659, opened 2022-06-24
LIVE-4234 increase shard size

waisingyiu:
## What does this change?

We are exploring ways to optimise the mobile notification delivery pipeline.  One suggestion is to increase the shard size so that the batch of notification requests increases in term of the number of device tokens and thus the overall number of the batches are reduced.  As a result, the harvester lambdas make fewer connections to the registration database (as well as execute fewer SQL queries).

This PR doubles the shard size from 10,000 to 20,000.  So we expect that it will half the number of database connection and double the average number of device tokens in each shard.

## How to test

It passed all the unit tests.  One unit test verifies the number of shards created against the new shard size.

We may experiment the PR on `CODE`.

It is considered a success if a breaking news notification only results in half the number of database connection it used to make while the overall performance improves.

If an invocation of harvester lambda fails, it will potentially have a bigger impact on the number of devices being unable to receive the notification.


groakland (approved):
LGTM. Let's get this tested on CODE and, fingers crossed, out to PROD. Great work

DavidLawes (approved):
Nice! 😄

waisingyiu:
We discussed the deployment time and agreed to ship this now.  We will keep our eye on the service for the afternoon and this PR can be rolled back easily if we need to.

PR #660, opened 2022-06-24
Add riffRaff upload settings to sbt

waisingyiu:
## What does this change?

We saw that we could not deploy a number of mobile-n10n modules (e.g. notification API, registration) in riff-raff.  Further investigation showed that some configurations were missing in the sbt:

Screenshot 2022-06-24 at 12 11 39

This PR adds the required settings to sbt so that the riffRaffUpload could work for all modules.

## How to test

All mobile-n10n modules could be shown in riffraff deployment page after teamcity build.  We succeeded to deploy `notification` via riff-raff.

Yes, no risks were found as it changes the riffraff upload settings only.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -40,6 +40,9 @@ val standardSettings = Seq[Setting[_]](
     "Guardian GitHub Releases" at "https://guardian.github.com/maven/repo-releases",
     "Guardian GitHub Snapshots" at "https://guardian.github.com/maven/repo-snapshots"
   ),
+  riffRaffManifestProjectName := s"mobile-n10n:${name.value}",
+  riffRaffUploadArtifactBucket := Option("riffraff-artifact"),
+  riffRaffUploadManifestBucket := Option("riffraff-builds"),
   libraryDependencies ++= Seq(
     "com.github.nscala-time" %% "nscala-time" % "2.24.0",
     "com.softwaremill.macwire" %% "macros" % "2.3.3" % "provided",


DavidLawes (approved):
Thanks Wai Sing :) If I understand correctly:
- the standard settings will now specify the riffraff project name and s3 buckets
- the standard settings are used by our play apps (registration and notification)
- when team city builds, for the play apps it'll now have the configuration needed to deploy assets

One of thing we could think about (in a different ticket/pr, after we've discussed as a team if we think it could be useful) is using [prout](https://github.com/guardian/prout)

build.sbt
```diff
@@ -40,6 +40,9 @@ val standardSettings = Seq[Setting[_]](
     "Guardian GitHub Releases" at "https://guardian.github.com/maven/repo-releases",
     "Guardian GitHub Snapshots" at "https://guardian.github.com/maven/repo-snapshots"
   ),
+  riffRaffManifestProjectName := s"mobile-n10n:${name.value}",
+  riffRaffUploadArtifactBucket := Option("riffraff-artifact"),
+  riffRaffUploadManifestBucket := Option("riffraff-builds"),
```
akash1810, line 45: If you're on the latest [version](https://index.scala-lang.org/guardian/sbt-riffraff-artifact/artifacts/sbt-riffraff-artifact) of `sbt-riffraff-artifact`, you can drop these. Since https://github.com/guardian/sbt-riffraff-artifact/pull/63, they have sensible defaults 🎉 .

DavidLawes: Thanks :) If I've understood correctly [this](https://github.com/guardian/mobile-n10n/pull/661) PR implements your suggestions

PR #663, opened 2022-06-27
LIVE-4203: log system metrics for the notifications app

DavidLawes:
## What does this change?

This change adds some custom logging into the notification app (the harvester and workers apps will have additional logging in subsequent PRs). The aim is to increase observability of the performance of the notification send flow and forms part of a task to implement a test rig in CODE that enables us to understand/measure changes to our system.

Some rationale for the abstraction of the custom logging fields:
- I moved it into common because multiple apps will probably want to log with custom fields
- The custom log fields rely on a change to the index pattern in our elk stack. The additional fields that have been manually added include a specified type, which now matches the types for the `value` in the case classes.

The custom logs will power a [dashboard](https://logs.gutools.co.uk/s/mobile/goto/2ee47100-f637-11ec-a2fe-3d2cf57c73cd)

The notification start time on its own may not be super helpful (it's also being sent as a long, milliseconds since the epoch) - however, in the worker lambda I plan to log the eventual end time. I'm hoping I'll be able to diff the start and end times to get a ms duration for the overall send time.

## How to test

When a notification is sent we should see additional properties included, eg:
Screenshot 2022-06-27 at 17 36 28

We can utilise the custom log fields to power our dashboard.

### DIFF
--- a/common/src/main/scala/models/NotificationLogging.scala
+++ b/common/src/main/scala/models/NotificationLogging.scala
@@ -1,0 +1,30 @@
+package models
+
+import net.logstash.logback.marker.LogstashMarker
+import net.logstash.logback.marker.Markers.appendEntries
+import org.slf4j.Logger
+
+import java.util.UUID
+
+import scala.jdk.CollectionConverters.MapHasAsJava
+
+sealed trait LoggingField {
+  val name: String
+  val value: Any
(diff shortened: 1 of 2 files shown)


frankie297 (approved):
👍

PR #669, opened 2022-07-05
Bump postgresql version

DavidLawes:
## What does this change?

Extracts transitive dependency that has a security vulnerability. For this dependency we tested in CODE to ensure interaction with the postgres db is not impacted.

## How to test

Send a dryRun push notification and validate no errors in harvester.


waisingyiu (approved):
LGTM

PR #677, opened 2022-07-12
Revert "LIVE-4234 increase shard size"

DavidLawes:
Reverts guardian/mobile-n10n#659

Looking at the stats for our 90% in 2mins target, it appears that this change hasn't improved the situation:

90-in-2-Shard-Change-Before-And-After

We agreed to revert and continue to monitor.


waisingyiu (approved):
LGTM

PR #682, opened 2022-07-15
Use logstash encoder for lambda logging

DavidLawes:
## What does this change?

In the mobile team we want to spend the next quarter working towards our notification send SLO: 90% of our readers receive a notification within 2mins.

In order to support this we need to measure our current system performance and enable easy test/measure iteration cycles.

Being able to log custom fields/markers is key to us building visualisations from the log documents. In the scala play apps there was already a working pattern for achieving this.

For scala lambdas a slightly different solution was needed:
- remove all usage of log4j
- integrate [logback-logstash-encoder](https://github.com/logfellow/logstash-logback-encoder)

In doing so our scala lambdas can emit json with custom fields. Previously we'd tried to use [aws powertools](https://awslabs.github.io/aws-lambda-powertools-java/) but the custom fields could only be type `string`, reducing the usefulness of the data we could log (and subsequently query).

I've deployed the changes to CODE and verified the json source of the log documents and confirmed that type `string` and type `long` were correctly parsed, so I believe this implementation meets our current needs.

Screenshot 2022-07-15 at 09 39 13

## How to test

I pushed to CODE and verified logs included the custom fields, where applicable. Also validated the custom fields had the correct type.

The [logback-logstash-encoder](https://github.com/logfellow/logstash-logback-encoder) library is used in other projects at the Guardian, so I don't believe we're introducing new tools or processes.


notificationworkerlambda/src/main/resources/logback.xml
```diff
@@ -0,0 +1,14 @@
...
+<configuration>
+    <statusListener class="ch.qos.logback.core.status.NopStatusListener" />
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+            <pattern>
+                %d{yyyy-MM-dd HH:mm:ss} [%t] %X{AWSRequestId} %-5p %c{1}:%L - %m%n
+            </pattern>
+        </encoder>
+    </appender>
+    <root level="info">
```
DavidLawes, line 11: When the level was "all" we got a lot of trace log data, which I believe cluttered up our view of what was happening in our lambdas. For now, I believe this level logs logger.info and logger.error statements, so should be equivalent to our previous log configuration.

waisingyiu (approved):
Thank you for your great work!

DavidLawes:
@waisingyiu I've been validating this change in CODE:

Although the logback.xml specifies a root value of `info` we do continue to see error log messages, this observation is confirmed by the [docs](https://logback.qos.ch/manual/architecture.html#effectiveLevel)

In the previous log4j2.xml we had an encoder that applied a common prefix to all log messages, eg
`2022-07-18 13:41:51 [scala-execution-context-global-25]  INFO  SqlRegistrationRepository:100 - Result:...`

I don't think this prefix is being pulled through correctly so we don't see any prefix for the new lambda logs, we just see:
`Result:...`

^ I don't _think_ this is a huge problem/degradation because we now include additional fields for the log messages we care about.

What do you think? If it doesn't look like too much of an issue to you, I will remove the `<pattern>` tag from the logback.xml and then merge this PR.

waisingyiu:
I agree with you that it is not much of an issue.    I saw that the information are available in the additional fields such as `thread_name`, `level` and `logger_name`.

PR #685, opened 2022-07-18
Increase harvester logging

DavidLawes:
## What does this change?

In this PR we add some additional log messages to the harvester so we can power a [dashboard](https://logs.gutools.co.uk/s/mobile/goto/2f6111c0-06a3-11ed-b0f3-ed9979428e02):

Screenshot 2022-07-18 at 15 56 36

Knowing this kind of information has a couple of benefits:
- It will allow us to benchmark our current performance in both CODE and PROD
- It will support analysis when testing new features in CODE (this dashboard replaces some manual work previously required to search logs and perform the equivalent aggregations)
- It will allow us to keep a track of how new features affect performance after deployment to PROD

Note that this implementation for logging custom markers does not use the common notification logging class from `common`. This implementation requires less boiler plate code as it slightly easier to read. It's also the approach used by other [teams](https://github.com/guardian/ophan-geoip-db-refresher/blob/ac2e939b16edea5bf68b676037fceed372f00b11/src/main/scala/ophan/geoip/db/refresher/Lambda.scala)

In future PRs I will revisit the logging for the notification play app and refactor.

## How to test

- Send a notification
- Validate in the 'discover' section of kibana that harvester logs have been generated
- Compare the logs in the 'discover' tab to the aggregates calculated in the [dashboard](https://logs.gutools.co.uk/s/mobile/goto/2f6111c0-06a3-11ed-b0f3-ed9979428e02)

- Logs that we generate from the harvester provide us with adequate information to support analysis and decision-making around new features we'd like to implement.

We are logging some extra stuff so index size may increase slightly, I don't think this is something we need to worry about because many other teams also log custom markers.


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -79,14 +80,14 @@ trait HarvesterRequestHandler extends Logging {
   def queueShardedNotification(shardedNotifications: Stream[IO, ShardedNotification], tokenService: TokenService[IO]): Stream[IO, Unit] = {
     for {
       shardedNotification <- shardedNotifications
-      _ = NotificationLogging.logInfoWithCustomMarkers("Processing notification", List(NotificationIdField(shardedNotification.notification.id)))
+      notificationId = shardedNotification.notification.id
```
waisingyiu, line 83: I suppose if the log "Processing notification" is not needed?

waisingyiu (approved):
Thank you for your great work!

PR #687, opened 2022-07-21
Document testing to reduce lambda duration

DavidLawes:
## What does this change?

Documents results and conclusions about how we could reduce the duration of our worker lambdas.


waisingyiu (approved):
Thank you for your great write-up!

PR #688, opened 2022-07-22
Add logging across stack

DavidLawes:
## What does this change?

It would be useful to know some more insights about the notification send flow, for example:
- how long do the worker apps take to send messages to apns?
- what is the total time taken to send a notification, from when it's received by the play app to when the last worker completes sending to apns/firebase?

These log messages enable this analysis and support an updated dashboard

For each notification we can calculate the total duration:
Screenshot 2022-07-22 at 12 50 29

For each notification we can generate some stats for the worker lambdas:
Screenshot 2022-07-22 at 12 51 26

Changes:
- Remove notification logging class and replace with a more simple implementation
- Ensure consistent naming of log fields across applications
- Include additional log fields in the worker lambdas


waisingyiu (approved):
Thank you David.  It would be very useful to have these figures on the dashboard.

DavidLawes:
> Thank you David. It would be very useful to have these figures on the dashboard.

Yes, good spot! I'd added these figures but hadn't saved the dashboard - I have saved it now so they should be visible to everyone :)

PR #689, opened 2022-07-22
Document testing of provisioned concurrency

DavidLawes:
## What does this change?

No code changes, only documenting the results of defining provisioned concurrency for the harvester lambdas.


waisingyiu (approved):
LGTM

PR #690, opened 2022-07-22
LIVE-4316 Document registrations database tuning and the test results

waisingyiu:
## What does this change?

Documents results and conclusions about the registrations database tuning.


DavidLawes (approved):
This is amazing and super helpful for my learning, thank you!

PR #692, opened 2022-07-28
Add stream to notifications-reports dynamo table

tomrf1:
For the apps metering project we need an event stream from the notifications table. We can use this to exclude articles from metering.
Because the notifications service writes to the dynamo table before initiating the send, we should get these events ahead of the mobile apps.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-dynamodb-streamspecification.html


frankie297 (approved):
👍

PR #693, opened 2022-07-28
Document testing to reduce lambda cold start time

DavidLawes:
## What does this change?

Document the testing we've done to try and reduce the lambda cold start time (no code changes)

### DIFF
--- a/docs/testing/00-notification-improvement-plan.md
+++ b/docs/testing/00-notification-improvement-plan.md
@@ -21,9 +21,11 @@ Current ideas and initiatives for the fine-tuning phase:
 - [Increasing shard size](./01-shard-size.md)
 - [RDS proxy](./02-rds-proxy.md)
 - Aurora and postgres 14
-- Query performance improvements 
-- ~~Handling DB connection errors more gracefully~~ parked for now since errors reduced after DB vacuum (including re-indexing) 
-- Lambda concurrency
+- [Query performance improvements ](../architecture/02-database-tuning.md)
+- ~~Handling DB connection errors more gracefully~~ parked for now since errors reduced after DB vacuum (including re-indexing)
+- [Lambda concurrency](./04-reducing-lambda-duration.md)
+- [Reducing lambda execution time](./06-reducing-lambda-cold-start-times.md)
+- Change batch size for sender lambdas (increase throughput to apns/fcm)
 
 ## Product improvements
 
@@ -36,6 +38,8 @@ Ideas for potential product improvements that would likely require some integrat
 
 Current ideas for re-architecting:
 
-- [Microsplinters](https://github.com/itsibitzi/n10n-poc/blob/main/n10n-broker/src/main.rs): preshard the DB and use these for querying 
+- [Microsplinters](https://github.com/itsibitzi/n10n-poc/blob/main/n10n-broker/src/main.rs): preshard the DB and use these for querying
 - Using Go to handle concurrency
 - Using ZIO to handle concurrency
+- Change lambda runtime: there are [studies](https://filia-aleks.medium.com/aws-lambda-battle-2021-performance-comparison-for-all-languages-c1b441005fd1) demonstrating that jvm-based lambdas are the worst performing, we could get benefits from migrating to typescript lambdas (or go, rust)
+- Pre-cache breaking news tokens: create a cron job that queries db to grab tokens for our breaking news topics to avoid these queries being executed as part of our send flow. Different flow for breaking news notifications: when requested we skip the harvester and immediately populate the sender queues with tokens to send.
--- a/docs/testing/06-reducing-lambda-cold-start-times.md
+++ b/docs/testing/06-reducing-lambda-cold-start-times.md
@@ -1,0 +1,128 @@
+# Reducing Lambda Cold Start Times
+
+The hypothesis we tested was that by:
+- minimising lambda dependencies, and
+- reverting to deploying lambdas with a .zip archive, and
+- using the latest aws sdk (v2), and
+- compiling for the latest java version (java 11 corretto), and
+- enabling only level 1 tiered compilation
+
+then the result would be reduced lambda start up and execution times.
+
+## Summary
+
+Splitting out the ios (and android) sender lambdas reduced the size of the assembled code to ~80MB.
+
+Given the reduced size of the assembled code we could revert to deploying the lambda with a zip archive (as opposed to a container image). This resulted in a 0.7s saving during cold starts.
+
+Migrating to the latest version of the aws java sdk (v2) resulted in further 0.5s reduction in cold start time.
+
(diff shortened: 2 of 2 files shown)


frankie297 (approved):

mchv:
Few other things you could try from simple to more difficult:
- Use `arm64` as runtime (only one config change) see [documentation](https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/)
- As you mentioned, change the tiered compilation to `1` (a single JVM parameter) see [documentation](https://aws.amazon.com/blogs/compute/optimizing-aws-lambda-function-performance-for-java/) but you need first to reduce dependencies (see below)
- reduce dependencies 
  - do not depends on `common` 
  - do not depends on `joda`
  - review if you can drop some of the `json` library as seems different are used.
- Update some dependencies
- You should update to `scala` to `2.13.8` which is unlikely to have an immediate impact but still contains some fixes
- You can update to latest `cats` 2.x (`2.8.0`) and review if you can update cats effect to `3.x` which I think is a transitive dependency and will provide faster implementation but can be a big effort.

Please be careful in the the idea of rewriting in another language.The  `jvm` is really good at managing concurrent IO with excellent performance, while expressing logic at a higher level. I don't think rewriting in another language and achieving good performance (aside of fast cold start) is trivial, but of course remain possible.

DavidLawes:
Thanks for these suggestions @mchv :) I'll take another look at the tiered compilation setting and how to reduce the dependencies. I'd also tested using arm64 as the runtime but didn't notice an impact, I missed this out of my test results (sorry!)

Great insight about jvm vs other languages. At the moment I don't think we have enough data/evidence to suggest migrating the sender lambdas to another language would solve our problem. Our next set of steps include understanding more about how the sender lambdas are operating (inc more detailed breakdown of how time is being spent/lost) & trying to increase the concurrency of our lambdas (e.g. by defining a provisioned concurrency). 

If you have any other suggestions or thoughts, would love to hear them!

mchv:
@DavidLawes You are welcome, 2 other small suggestions:
- Merge #670 because it contains a [performance fix](https://github.com/pgjdbc/pgjdbc/pull/2495) since [42.3.5](https://github.com/pgjdbc/pgjdbc/blob/master/CHANGELOG.md#4235-2022-05-04-084835--0400)
- Try [`sbt-proguard`](https://github.com/sbt/sbt-proguard) to shrink the jar size, without obfuscating the code (see [`-dontobfuscate` option](https://www.guardsquare.com/manual/configuration/usage)). I think @johnduffell as used it in the past.

johnduffell:
I did try sbt-proguard and something else similar once but it only works if you can accurately define the "roots" of your code.  Where things rely on reflection to find classes (usually loggers and I think AWS client (v1?) did) then you have to also define those also.  
In the end I decided it is a nice idea but in reality it would be flaky and worse still it could fail at runtime due to a missing class, so would require careful testing each deploy.
I think but I'm not sure that "unnecessary code that's executed" has a bigger impact than "code that's in the Jar but not used."

Mario G did a load of work on making lambdas fast, he did a simple one with minimal libraries (upickle/ujson, built in Java http client, built in logger) and it was super fast. I can dig out some info if it's useful.  Also he (and Adam F separately and possibly also Regis) did experimentation with native compilation (graalvm) and had some success, but again with some issues around reflection in AWS library.  Again I can try to dig more info out if it's useful.

If it is the cold start time that is the issue, I would consider whether it would be economical to either run the lambda in a dry-run mode on a schedule to keep it warm, or pay for a burstable ec2 instance to run all the time which would only be tens of dollars a month (plus cost to maintain it) and might be more efficient in terms of dev time?

mchv:
Thanks @johnduffell, this is super helpful!

DavidLawes:
Thanks @johnduffell - we've been chatting about your ideas :) We think the idea of the dry run to keep the lambdas warm is a great idea, we're definitely going to test it out!

johnduffell:
glad it's useful - let me know if you want any further thoughts or review anything based on my experience!

mchv:
Another quick things to try regarding performance generally (this will not help directly cold start problem) is [AWS CodeGuru profiler](https://docs.aws.amazon.com/codeguru/latest/profiler-ug/what-is-codeguru-profiler.html). You need to [set-up for the lambda](https://docs.aws.amazon.com/codeguru/latest/profiler-ug/setting-up-short.html) and then [code integration is trivial](https://docs.aws.amazon.com/codeguru/latest/profiler-ug/scala-language-support.html).

DavidLawes:
> Another quick things to try regarding performance generally (this will not help directly cold start problem) is [AWS CodeGuru profiler](https://docs.aws.amazon.com/codeguru/latest/profiler-ug/what-is-codeguru-profiler.html). You need to [set-up for the lambda](https://docs.aws.amazon.com/codeguru/latest/profiler-ug/setting-up-short.html) and then [code integration is trivial](https://docs.aws.amazon.com/codeguru/latest/profiler-ug/scala-language-support.html).

Ah, this is super interesting, thank you for sharing! I will add this to our list of initiatives about how to improve lambda performance.

We did some investigation into our lambda [timings](https://github.com/guardian/mobile-n10n/blob/main/docs/testing/07-lambda-timing.md) of our lambdas and we also raised a support case with aws. We believe we may have a problem with concurrent executions of our lambdas (this is backed up by cloudwatch [metrics](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2:graph=~%28metrics~%28~%28~%27AWS*2fSQS~%27ApproximateAgeOfOldestMessage~%27QueueName~%27mobile-notifications-harvester-PROD-Sqs-NPP9X15G7WAO%29~%28~%27...~%27notification-sender-workers-cdk-PROD-iosSenderSqsA94CB0A0-PT5FWNUIRURH%29%29~view~%27timeSeries~stacked~false~region~%27eu-west-1~start~%27-PT168H~end~%27P0D~stat~%27Average~period~60%29;query=~%27*7bAWS*2fSQS*2cQueueName*7d*20PROD*20Age*20ios) showing how long we have sqs messages waiting on the queue to be processed)

Our next set of tests include:
- validating this theory by testing whether, by defining provisioned concurrency for our lambdas, the performance improves
- reducing the number of lambdas we're attempting to run in parallel by e.g.  increasing batch size

As an aside, I believe our lambda code could be improved. We've seen some lambdas starting to time out after 1m30s when trying to send notifications to firebase. At the moment we make 1 api request per device, but firebase now support making [batch](https://firebase.google.com/docs/reference/admin/java/reference/com/google/firebase/messaging/FirebaseMessaging) requests which could improve our performance/efficiency.

Thought i'd note our current thoughts/plans in case it sparks any other ideas :)

PR #695, opened 2022-07-29
Update dependencies & small improvement

mchv:
## What does this change?

- Update some dependencies

  - SBT to `1.7.1` a2a6b49e6276a0d4b185f147aa2786045ba883d7
  - Scala to `2.12.8` 395af10461f4d060f934c28cc896959dd54bf1c9

- Use fast exception ([because normal ones are expensive](https://www.baeldung.com/java-exceptions-performance)) for control flow e9b70014d9f47846786e2087bfef983037c8eccf

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -8,7 +8,7 @@ import com.typesafe.sbt.packager.docker.{Cmd, ExecCmd}
 val projectVersion = "1.0-latest"
 
 organization := "com.gu"
-ThisBuild / scalaVersion := "2.13.2"
+ThisBuild / scalaVersion := "2.13.8"
 
 val compilerOptions = Seq(
   "-deprecation",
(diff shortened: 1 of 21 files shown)


DavidLawes (approved):
Thanks :) I think these changes make sense to me. Thinking specifically about how we handle exceptions in the worker lambdas we're now not going to log the stack trace in some cases (which does make sense actually, I've seen examples where our logs are cluttered with excessive stack traces).

I notice we explicitly change the error handling in the fcm client but not for apns, that code is new to me but I'm guessing some of the error handling is abstracted by the pushy library so there are no changes we need to make there?

mchv:
>  notice we explicitly change the error handling in the fcm client but not for apns, that code is new to me but I'm guessing some of the error handling is abstracted by the pushy library so there are no changes we need to make there?
The change I have made is because the compiler became better at detected unhandled case in pattern matching, so I think you are right.

mchv:
I will not merge this before next week.

PR #698, opened 2022-08-01
Add logging to chunked tokens, to help calculate invocation count

frankie297:
## What does this change?
Add logging to chunked tokens, to help calculate invocation count

## How to test
Send test notification and see total invocation count


DavidLawes (approved):
🎉

PR #700, opened 2022-08-02
Increase chunked token batch size

frankie297:
## What does this change?
Increased the chunked token batch size to 5000 from 1000, to decrease the amount of invocations needed to send a notification. In the hope we can ease the pressure we have on concurrency.

| Before (invocation count) | After (invocation count) |
| --------------------------| -------------------------|
| 1600 (approx) | 400 (approx)

## How to test
Send notification before and after batch size change to see difference in invocation count. Should we double batch size to 2000 to check performance? It's hard to test true performance until in PROD, so we could roll this change, monitor and rollback if necessary.


DavidLawes (approved):
I think it's worth pushing this to prod and evaluating, in prod we can evaluate how this change impacts lambda invocations, concurrency and total duration. In code we couldn't understand the impact as clearly (we use dryRun in code which results in 0 api calls to apns/firebase).

Note that we have already increased our logging in the sender lambdas so our dashboard now shows the total number of sender invocations as well as the average chunk token size that each invocation has to process, with this increased observability I believe we can assess the impact of this change quite clearly, eg:

Screenshot 2022-08-02 at 12 04 05

Our initial hesitation was around the value of the chunk token size we're suggesting: multiplying by 5 may have unexpected results (considering any trade off between the total tokens each invocation has to process vs any performance improvement due to reduced concurrency). However, we'll be continually monitoring so if we severely degrade the performance we can easily revert.

waisingyiu (commented):
It is a good idea.

I suppose the expected duration of one sender invocation will not exceed the timeout of lambda?

DavidLawes:
> It is a good idea.
> 
> I suppose the expected duration of one sender invocation will not exceed the timeout of lambda?

This is a good point. At the moment the timeout is defined as 1m30s. The average duration of the lambdas seems to range from 70ms - 18s.

I can see 2 instances in the last week where the maximum duration of the lambda was 1m30 (and so was probably terminated by aws).

I think it might be good to:
- reduce `chunkN` value to 2000 (so we're only doubling the potential max number of tokens processed by a single invocation)
- double the lambda timeout defined for these functions

jacobwinch:
> I think it might be good to:
> 
> * reduce `chunkN` value to 2000 (so we're only doubling the potential max number of tokens processed by a single invocation)
> * double the lambda timeout defined for these functions

This sounds like a sensible idea. If you are increasing the timeout for the function, you probably also want to increase the [visibility timeout](https://github.com/guardian/mobile-n10n/blob/357a9ba521392141f67f152afeb3cad1517477da/notificationworkerlambda/cdk/lib/senderworker.ts#L41) to [prevent messages from being processed twice](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-sqs-queue.html#cfn-sqs-queue-visibilitytimeout).

PR #704, opened 2022-08-10
Results from testing provisioned concurrency

DavidLawes:
## What does this change?

No code changes as a result of this PR, it's just for documentation purposes.


DavidLawes:
cc @frankie297 - I've tried to write up where we got to so far with the test using 100 provisioned concurrency. Will also attempt to fill in some data for the testing at 200 before end of day.

DavidLawes:
NB: we're going to re-run some tests and will update this PR after they're complete. Will hold off merging until that

DavidLawes:
It's worth noting the recommendations from this piece of work:

- test additional configuration that we can alter to improve concurrency (batch window and visibility timeout), which will be picked up in the current sprint
- advice to editorial: waiting 2 mins between notifications significantly improves send times

groakland (approved):
Very clear and readable. Great work

PR #716, opened 2022-08-12
LIVE-3899 fix RDS proxy secrets

waisingyiu:
## What does this change?

When we set up a RDS proxy in  #666 , we should manually populate the `secrets` created in Secrets Manager with the actual database passwords.  It creates a problem where we may change some parameters in the cloudformation stack `registrations-db-proxy` and as a result the AWS re-initialises the passwords in `secrets` to randomly generated strings.  Manual operations are then required to populate the secrets again.

 Newer version of AWS CDK allows us to reference the secret values stored in SSM parameter store and populates `secrets` with the actual passwords upon cloudformation creation / update.  The generated cloudformation JSON does not contain the passwords.  It just shows the path of the parameter store which is then accessed by AWS during cloudformation provisioning.

This PR bumps the GuCDK to the latest version v47 and AWS CDK v2.35 so that it uses this feature to reference SSM to set the passwords in `secrets`.

## How to test

We checked the generated cloudformation JSON and tested it on `CODE`.

### DIFF
--- a/cdk/lib/__snapshots__/registrations-db-proxy.test.ts.snap
+++ b/cdk/lib/__snapshots__/registrations-db-proxy.test.ts.snap
@@ -60,11 +60,8 @@ Object {
       "DeletionPolicy": "Delete",
       "Properties": Object {
         "Description": "Secrets for accessing registration database from cleaner lambdas",
-        "GenerateSecretString": Object {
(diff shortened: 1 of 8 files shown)


waisingyiu:
This approach of referencing SSM does not work on secrets manager.

PR #724, opened 2022-08-19
Notification app metrics

DavidLawes:
## What does this change?

The reliability team has provided a very useful [doc](https://docs.google.com/document/d/10AnOZ4MLjuTO7mXaySoVO2SwmroLmns4gGmsfY7egmw/edit?usp=sharing) about how to incorporate metrics into our notification platform. 

This PR adds metrics to the notification app only:
- notification app processing time
- count of breaking news notifications

This PR also changes the logic for how duration is calculated: previously the duration didn't account for the Future completing so the duration looked to always be ~0ms. Now I'm seeing more realistic durations (600-1700ms).

I can see that the metrics are being pushed to cloudwatch as expected:
Screenshot 2022-08-19 at 15 06 04

## How to test

I pushed to CODE, sent a couple of test notifications and checked:
- logs
- kibana dashboard
- cloudwatch metrics

Given we publish these metrics I think we'll have the observability we need to monitor the notification app.

We're pushing custom metrics which incurs additional cost, but I'm not using additional dimensions which hopefully controls this additional cost ($0.30 per month, $0.01 per 1,000 requests).

### DIFF
--- a/notification/app/notification/controllers/Main.scala
+++ b/notification/app/notification/controllers/Main.scala
@@ -22,7 +22,6 @@ import tracking.SentNotificationReportRepository
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 import scala.jdk.CollectionConverters.MapHasAsJava
-import scala.util.{Failure, Success}
 
 final class Main(
   configuration: Configuration,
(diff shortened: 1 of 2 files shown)


frankie297 (approved):
LGTM 👍

jacobwinch (approved):
Great work! 👍

PR #726, opened 2022-08-22
Update the procedure around the logical replication

waisingyiu:
## What does this change?

When we attempted to perform the database upgrade, we found that the `replica identity` was required for the initial data copy phase of logical replication when the source database was constantly changing.

This PR updates the database upgrade procedure as below:
1. create primary key constraint before starting logical replication.  The primary key is the replica identity by default.
2. switch over the Registrations API to the new RDS instance before stopping logical replication.  The purpose is to minimise the data loss during switchover.

This PR changes the documentation only.


DavidLawes (approved):
👍

groakland (commented):
LGTM

PR #727, opened 2022-08-23
Document test results from increasing thread pool size

DavidLawes:
## What does this change?

No code changes, just documenting the results and recommendations from the testing around increasing the thread pool size.

Branch for reference: https://github.com/guardian/mobile-n10n/tree/sender-lambda-custom-execution-context

PR #729, opened 2022-08-23
Sender lambda custom execution context

DavidLawes:
## What does this change?

As discussed in https://github.com/guardian/mobile-n10n/pull/727 there can be performance improvements by increasing the thread pool size.

We did some testing in code to evaluate the change:
- we defined new SSM variables for the thread pool size, allowing us to easily increase/decrease the value (without requiring code changes)
- we saw no errors or out memory issues when testing in code and identified the key metrics that allow us to understand if the change is successful

Before this change is merged we need to define new parameters in SSM:
- /notifications/PROD/workers/ios/apns.threadPoolSize
- /notifications/PROD/workers/ios-edition/apns.threadPoolSize
- /notifications/PROD/workers/android/fcm.threadPoolSize
- /notifications/PROD/workers/android-edition/fcm.threadPoolSize
- /notifications/PROD/workers/android-beta/fcm.threadPoolSize

It's suggested to initially set all the thread pools to a size of 10.

Once deployed we will monitor the following metrics:
- memory usage of the lambdas
- alerts or errors from the lambdas
- processing time / 90in2 statistic

It's suggested that we increase only the thread pool size for the `ios` and `android` workers. It's suggested to gradually increase the thread pool size (over a number of days) and monitor the results.

## How to test

Once deployed ensure lambdas correctly initialise (including all ssm variables being pulled from systems manager).

Evaluate the metrics mentioned above and check for any related alerts.

There can be risks by increasing the thread pool size, e.g. out of memory issues. We intend to gradually increase the thread pool size. We will only increase if we see performance improvements over a number of days. We can easily change the thread pool size by changing the configuration in SSM (no code change required).


groakland (approved):
LGTM

rtyley:
I think increasing the thread pool size may have improved performance because threads were being _blocked_, which is often avoidable with async code. It turns out that this `asScala` method (used in several places in the lambda to convert the Java `Future`s returned by Google API SDK's into _Scala_ `Future`'s) is accidentally blocking because [`f.get()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Future.html#get()) blocks, ie it stops all execution on the thread until the result is available:

https://github.com/guardian/mobile-n10n/blob/fbc66f46284a1c994ee6b07dca9935f03f867e97/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala#L98-L103

When code is blocking inside a Scala Future like this, it's tying up a thread, and so that limits other work from occurring unless other threads are allocated. _This_ PR (#729) increased the size of the thread pool to allow more operations to happen at once, increasing performance. Depending on the type of `ExecutionContext` in use you could also wrap the blocking code in the [`scala.concurrent.blocking()`](https://docs.scala-lang.org/overviews/core/futures.html#blocking-inside-a-future) construct in order to indicate to the `ExecutionContext` that it should allocate an additional thread while the code is blocking.

Even better than that tho' is to write code that is non-blocking... how can we do that here?

### How can we write a better, non-blocking method to convert Java `ApiFuture`s to Scala `Future`s?

The [`ApiFuture`](https://cloud.google.com/java/docs/reference/api-common/latest/com.google.api.core.ApiFuture) used by Google APIs is _"similar to Guava's [ListenableFuture](https://guava.dev/releases/21.0/api/docs/com/google/common/util/concurrent/ListenableFuture.html), but redeclared"_. There's some good resources talking about [how to convert a `ListenableFuture` to a Scala `Future`](https://stackoverflow.com/q/18026601/438886), and we can use the same stuff for `ApiFuture`, eg:

```scala
import com.google.api.core.{ApiFuture, ApiFutureCallback, ApiFutures}

implicit class RichApiFuture[T](val af: ApiFuture[T]) {
  def asScala(implicit ec: ExecutionContext): Future[T] = {
    val p = Promise[T]()
    ApiFutures.addCallback(af, new ApiFutureCallback[T] {
      def onFailure(t: Throwable): Unit = p failure t
      def onSuccess(result: T): Unit = p success result
    }, (command: Runnable) => ec.execute(command)))
    p.future
  }
}
```


#### Further background

[_"The Case For Effect Systems"_](https://www.youtube.com/watch?v=qgfCmQ-2tW0) by Daniel Spiewak, is an amazing summary/introduction to callbacks, Futures, and even what is _beyond_ Futures, worth checking out!

PR #737, opened 2022-08-31
LIVE-4421 RDS instance classes

waisingyiu:
## What does this change?

We are exploring ways to improve the performance of our notification delivery pipeline.

We tested the performance of harvester with different RDS instance classes, and summarised the work in a document.

This PR adds the document to the repository. 

## How to test

No changes to the program code were made in this PR.


DavidLawes (approved):
👍

PR #739, opened 2022-09-01
Clean unregistered android tokens

DavidLawes:
## What does this change?

When attempting to send android users a notification, this changes which error responses from firebase are parsed as an `InvalidToken`. Parsing error responses as an `InvalidToken` will trigger android tokens to be cleaned from the registrations db.

We can see in the logs that we have a large number of errors from the android sender lambda (where peaks in these error messages occur when sending breaking news notifications to a large number of recipients):
Screenshot 2022-09-01 at 12 40 04

We can see in the google console that these error responses have been present for over a month:
Screenshot 2022-09-01 at 12 40 42

We can see that ios sender lambdas trigger about 1.7k tokens a week to be cleaned:
Screenshot 2022-09-01 at 12 41 45

Whereas android sender lambdas trigger only about 10 tokens a week to be cleaned (significantly less, even accounting for the difference in android vs ios users):
Screenshot 2022-09-01 at 12 41 19

Looking at the errors in the android sender lambdas in more detail, we can see that the error responses all share some common data:
- error code: `NOT_FOUND`
- error cause: `Requested entity was not found.`
- stack trace includes message:

```
Caused by: com.google.api.client.http.HttpResponseException: 404
POST https://fcm.googleapis.com/v1/projects/push-notifications-prod-15cbe/messages:send
{
  "error": {
    "code": 404,
    "message": "Requested entity was not found.",
    "status": "NOT_FOUND",
    "details": [
      {
        "@type": "type.googleapis.com/google.firebase.fcm.v1.FcmError",
        "errorCode": "UNREGISTERED"
      }
    ]
  }
}
```

## How to test

- We have provided unit tests to cover some of the pre-existing functionality as well as the new logic implemented.

In CODE we can test by:
- inserting a registration into the test rig db for a topic with no other subscribers
- trigger a notification send to the given topic in code, without using dryRun
- confirm that the token is successfully retrieved from the db and is attempted to be sent by the android sender lambda
- inspect the response from firebase and confirm, where the response information matches the log messages we've seen in prod, that the token is successfully parsed as an `InvalidToken` and that it is eventually cleaned from the db. 

- Successful responses from firebase continue to be parsed as expected
- More tokens are parsed as invalid and successfully cleaned from the db
- Number of errors in our logs decreases
- Number of error responses shown in the google console also decrease

We've tried to mitigate the risk by introducing unit tests covering both pre-existing and new functionality. We will also test in code to confirm the behaviour is expected.

We will also attempt to confirm our understanding of the error code with google/firebase before pushing this change (TBC).


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
```diff
@@ -14,12 +14,12 @@ import com.gu.notifications.worker.utils.UnwrappingExecutionException
...
 import java.util.UUID
-import java.util.concurrent.Executor
-import scala.concurrent.{ExecutionContextExecutor, Future, Promise}
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future}
 import scala.util.control.NonFatal
 import scala.util.{Failure, Success, Try}
 
-class FcmClient private (firebaseMessaging: FirebaseMessaging, firebaseApp: FirebaseApp, config: FcmConfig)
+
+class FcmClient (firebaseMessaging: FirebaseMessaging, firebaseApp: FirebaseApp, config: FcmConfig)
```
frankie297, line 22: Are we sure it's ok to make this class public?

DavidLawes: Yeah, a great question. I wondered about this too and here's my current thinking/understanding:
- The `FcmClient` object exposes a public method which creates a success/failure of a new instance of the `FcmClient`
- The public apply method also handles creating the connection to firebase and initialising our "firebase app", which is a prerequisite to being able to instantiate a new `FcmClient`
- This public method is only used in the AndroidSender when we first start the lambda: https://github.com/guardian/mobile-n10n/blob/main/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala#L23

To me the flow looks to have been implemented this way in order to:
- encapsulate the option-setting, app-initialisation and class instantiation all inside a `Try`
- provide a guide to users of the `FcmClient` about how to use the class

I think for our purposes, making the class public:
- allows us to test more easily (fewer mocks and simpler tests)
- doesn't impact how the AndroidSender currently interfaces with the client object and class

If in the future we needed to re-write the android sender lambda we would always need to instantiate with the given options, the current mechanism (using the public object's apply method) is a neat abstraction and making the class public doesn't stop us from continuing to use this.

Sorry, a bit of a ramble - what do you think? Maybe there are other concerns/considerations I've accidentally overlooked!

frankie297: I think thats a great explanation thank you for going into so much detail. 👍

frankie297 (approved):
I think everything looks sound to me

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
DavidLawes: I've also tried to use scalatest's PrivateMethodTester but failed spectacularly, possibly because I need to invoke the private apply method and then use the resulting instance to invoke sendNotification...

PR #746, opened 2022-09-07
LIVE-4500 Study the timing of harvester lambda execution

waisingyiu:
## What does this change?

This PR added a document which summarises the findings of the study about the timing of the harvester lambda execution.

## How to test

No changes to the code are made in the PR.


DavidLawes (approved):
👍 

Really reassuring that you reached the same kind of conclusions as me about lambda timings/concurrency!!

Definitely agree about the next step and very interested to see what results you get! When I tried increasing the batch size for the sender lambdas, my invocation count reduced (and so did the time messages spent on the queue) but then each invocation took n times longer (because it had to process n more messages per invocation). Overall, I didn't _appear_ to get a performance improvement, although I believe there could be other benefits from increasing the batch size (e.g. cost?)

waisingyiu:
Thank you, David, for sharing the result of increasing batch size in sender worker.  It appears that the harvester made use of `fs2` streaming library to process multiple records in parallel.  Not sure if it helps with the performance, but I think it may be worth testing it on CODE.

PR #747, opened 2022-09-08
Live 4419 batch api calls

frankie297:
## Accessibility


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/DeliveryService.scala
```diff
@@ -86,4 +90,59 @@ class DeliveryServiceImpl[F[_], C <: DeliveryClient] (
...
+        client.sendBatchNotification(
+          notification.id,
+          token,
+          payload,
+          notification.dryRun.contains(true) || client.dryRun
+        )(cb)
+      }
+    }
+
+    def sending(client: C)(token: List[String], payload: client.Payload): Stream[F, Either[DeliveryException, C#Success]] = {
```
DavidLawes, line 107: A lot of this logic was copypasta from the existing method for sending

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
```diff
@@ -80,6 +104,24 @@ class FcmClient (firebaseMessaging: FirebaseMessaging, firebaseApp: FirebaseApp,
       Left(UnknownReasonFailedRequest(notificationId, token))
   })
 
+  def parseBatchSendResponse(
+    notificationId: UUID, tokens: List[String], triedResponse: Try[BatchResponse]
+  )(cb: Either[Throwable, Success] => Unit): Unit = triedResponse match {
+    case Success(batchResponse) =>
+      // From firebase sdk docs: the order of the response list corresponds to the order of the input tokens
+      batchResponse.getResponses.asScala.toList.zip(tokens).foreach { el => {
```
DavidLawes, line 112: We need to associate any failure response to the corresponding token. The batch response alone doesn't facilitate this, but the docs indicate that the order of the batch responds corresponds to the order of the input.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/tokens/TokenService.scala
```diff
@@ -12,9 +12,11 @@ import play.api.libs.json.{Format, Json}
 import scala.concurrent.ExecutionContextExecutor
 
 case class IndividualNotification(notification: Notification, token: String)
+case class BatchNotification(notification: Notification, token: List[String])
 
 case class ChunkedTokens(notification: Notification, tokens: List[String], range: ShardRange) {
   def toNotificationToSends: List[IndividualNotification] = tokens.map(IndividualNotification(notification, _))
+  def toBatchNotificationToSends: List[BatchNotification] = tokens.grouped(500).map(BatchNotification(notification, _)).toList
```
DavidLawes, line 19: This is the bit of logic that batches the tokens from the SQS message body into groups of 500. I think it was suggested to maybe group slightly less that firebase's maximum, so we could think about reducing this by 10%, for example?

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -22,4 +31,24 @@ class AndroidSender extends SenderRequestHandler[FcmClient] {
   override val deliveryService: IO[Fcm[IO]] =
     FcmClient(config.fcmConfig).fold(e => IO.raiseError(e), c => IO.delay(new Fcm(c)))
   override val maxConcurrency = 100
+
+  //override the deliverChunkedTokens method to validate the success of sending batch notifications to the FCM client. This implementation could be refactored in the future to make it more streamlined with APNs
+  override def deliverChunkedTokens(chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant)]): Stream[IO, Unit] = {
```
DavidLawes, line 36: Thinking about how to test/experiment in production, we could think about implementing a switch here too:
- define a list of allowed topics
- only allowed topics are sent by the batch method
- increase allowed topic list as confidence increases

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Configuration.scala
```diff
@@ -104,7 +107,8 @@ object Configuration {
         debug = config.getBoolean("fcm.debug"),
         dryRun = config.getBoolean("dryrun")
       ),
-      config.getInt("fcm.threadPoolSize")
+      config.getInt("fcm.threadPoolSize"),
+      config.getStringList("fcm.allowedTopicsForBatchSend").asScala.toList
```
DavidLawes, line 111: This new parameter will need to be added for every android worker in each environment. We need to define the type as a "StringList".

Suggest we start by sending only to breaking/international for now? Plus maybe sport?

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -22,4 +31,33 @@ class AndroidSender extends SenderRequestHandler[FcmClient] {
...
+    for {
+      (chunkedTokens, sentTime, functionStartTime) <- chunkedTokenStream
+      isAllowedToSendBatch = chunkedTokens.notification.topic.forall(topic => config.allowedTopicsForBatchSend.contains(topic.toString))
```
DavidLawes, line 39: I initially wondered if I should create unit tests for this but the initial configuration/mocking wasn't simple. This logic would never exist in the long term, so instead I'll validate in the CODE environment (have already validated locally too).

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Configuration.scala
waisingyiu: `breaking/international` has around 550,000 recipients so it may be a good choice.  But not many alerts were sent to this topic only (we got 3 breakers in the last 10 days) so we may need to wait

waisingyiu: Yes, we may add sport as well.  `breaking/uk-sport` has 10,000 recipients, and `breaking/international-sport` has around 5,000.  They occurs quite frequently - at least once every day in the past 10 days except 18-Oct.  
Maybe we add one of them

DavidLawes: Thanks Wai Sing :) Maybe we start with 2 allowed topics? 

For ref: I couldn't get the StringList parameter to parse correctly, for this temporary experiment I've decided to go with a simple string that I split using regex `","`

waisingyiu: I agree to allow two topics, maybe `breaking/international` and `breaking/international-sport`?

It sounds good to me to use a simple string with comma delimiter for a temporary experiment

waisingyiu (approved):
👍

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
waisingyiu: Sounds good to me

PR #751, opened 2022-09-09
LIVE-4521 summarise the test with harvester batch size

waisingyiu:
## What does this change?

We have tested the performance of harvester with increased batch size / increased shard size.

The PR adds a document which summaries the test result and the findings.

## How to test

No changes are made to the programs.


groakland (approved):
Great write up, thanks

PR #752, opened 2022-09-12
LIVE-4538 Restrict JVM tiered compilation level to 1 for harvester

waisingyiu:
## What does this change?

We have experimented with the JVM settings about tiered compilation level in #756.  The [test results](https://github.com/guardian/mobile-n10n/blob/main/docs/testing/11-tiered-compilation.md) suggest that restricting the tiered compilation level to 1 is likely to improve the overall duration of harvester lambda.  

This PR applies this JVM setting to the harvester lambda function via an environment variable.

## How to test

This PR does not change any business logics.

We ran performance tests on `CODE` to evaluate the impact of this JVM setting.

We measure success using the 90in2min indicator, which is the percentage of recipients who are able to receive the notification under 2 mins.

It may give poorer performance as the configuration and the workload are different in `PROD` from `CODE`, but there is no evidence to indicate so.

In any events, we can revert this PR to restore to the original settings.


groakland (approved):
LGTM

groakland (approved):
LGTM

PR #756, opened 2022-09-14
LIVE-4538 Summarised test result of restricting tiered compilation level

waisingyiu:
## What does this change?

We have done some experiments with the JVM settings of restricting tiered compilation level and studied its impact on the performance of our lambda functions in the notification delivery pipeline.

This PR adds a document which summarised the test results.

No changes to the source code were made.


frankie297 (approved):
Great! 👍

PR #763, opened 2022-09-20
LIVE-4549: update cdk version

frankie297:
## What does this change?
This updates the AWS CDK from v1 to v2 (in the workers lambdas) and GU CDK version (across the project) to the latest so we can fix security vulnerabilities currently being blocked by this migration.

## How to test
Make sure notifications send as expected


waisingyiu (approved):
Thank you Frankie for the great work!

PR #767, opened 2022-09-22
Add documentation for batch api calls experiment

frankie297:
## What does this change?
Adds documentation for the batch api calls experimentation, with the view to promote this to a feature change rather than an experiment.


waisingyiu (approved):
👍

PR #771, opened 2022-09-28
Include multiple partition_dates if notification sent just before midnight

DavidLawes:
## What does this change?

We pushed the slomonitor lambda into production to collect data about the accuracy of the 90in2 count. The results revealed an edge case that severely impacted data accuracy for a specific scenario:
- A notification was sent at 23.59
- The athena query only counted results from the same day
- The majority of the notifications were received within the 2nd minute after sending, i.e. 00.01 the following day, and so were accidentally excluded from the 90in2 count

This PR attempts to handle the edge case to improve the data accuracy of slomonitor's 90in2 count:

- extract and extend the logic that generates the athena query string
- create test cases for query string generation
- refactor slomonitor object to extend from a trait allowing env vars to be overridden in tests

## How to test

We've added unit tests to cover the expected change in logic.

After pushing the change we should continue to monitor the 90in2 accuracy, if the edge case scenario happens again we should expect to see a 90in2 count that matches that seen via bigquery.

### DIFF
--- a/slomonitor/src/main/scala/com.gu.notifications.slos/SloMonitor.scala
+++ b/slomonitor/src/main/scala/com.gu.notifications.slos/SloMonitor.scala
@@ -89,5 +89,4 @@ trait SloMonitor {
 
 object SloMonitor extends SloMonitor {
   val stage: String = System.getenv("STAGE")
-
 }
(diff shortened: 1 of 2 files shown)


jacobwinch (approved):
Great work 💯 👍

PR #775, opened 2022-09-29
Fix failing slomonitor

DavidLawes:
## What does this change?

slomonitor was failing to execute in aws because of missing class constructor, we believe it's caused by recent refactoring of trait/object in #771. Refactoring the tests and trait/object to fix this error.

Screenshot 2022-09-29 at 09 27 21

## How to test

Deployed to CODE, sent a breaking news notification, validated slomonitor executed correctly.

Screenshot 2022-09-29 at 09 29 31


frankie297 (approved):
👍

PR #776, opened 2022-09-29
Live 4580 document cache tokens

waisingyiu:
## What does this change?

This PR adds a document which summarises the feasibility study of caching all the device tokens in the memory of Notification APIs.

## How to test

No changes have been made to the source code.


DavidLawes (approved):
Agree with the conclusions, it sounds like the caching of tokens might not be the most optimum approach:
- it would increase the complexity of the design
- it may not continue to scale well in the future as the number of tokens grow

I don't have a good grip on the possible performance improvements: the harvester streams tokens to the workers. The first tokens begin to be processed by the workers ~2s after the harvester starts processing. The harvester completes after ~40s. If we were to cache tokens then the best improvement would be 40s (for some of the tokens), the worst could be just 2s. Also, putting all the tokens on a queue at the same time would likely increase the concurrency-related issues in the worker lambdas...

PR #793, opened 2022-10-19
LIVE-4702 create AWS stack for EC2-based sender worker

waisingyiu:
## What does this change?

This PR creates the stack using GuCDK.  The stack mainly defines:

1. an autoscaling group to host the new sender worker process, and
2. five SQS queues for harvester to send messages to the new sender worker 

Based on current mechanism where harvester gets the URLs of the sender queues, it creates five parameters in SSM for the new queues:

- `/notifications/CODE/workers/harvester/androidBetaSqsEc2Url`
- `/notifications/CODE/workers/harvester/androidEditionSqsEc2Url`
- `/notifications/CODE/workers/harvester/androidLiveSqsEc2Url`
- `/notifications/CODE/workers/harvester/iosEditionSqsEc2Url`
- `/notifications/CODE/workers/harvester/iosLiveSqsEc2Url`

It also exports the ARNs of the new queues as the output of the cloudformation stack (`NotificationEc2SenderWorkerQueueArns-CODE`) so that the harvester's cloudformation gets the ARNs and grants the harvester function access to the queues.

No deployment setting (i.e. riffraff yaml) has been defined yet.  Manual deployment is required for the new stacks.

## Deployment

It requires manual deployment to create the stacks because the initial cloudformation parameters need to be supplied and it can only be done manually (at this moment).

After the stack has been created (and riff-raff has been set up), the stacks can be updated by riff-raff continuous deployment.

## How to test

A new stack has been created with the five new queues.

The stack is not used by existing programs in the notification service yet.  So no impact to production is expected.

### DIFF
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -50,6 +50,7 @@ export const senderCodeProps = {
 		maximumInstances: 2,
 	},
 	instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
+	targetCpuUtilization: 20,
 	notificationSnsTopic:
 		'arn:aws:sns:eu-west-1:201359054765:AutoscalingNotificationsCODE',
 	cleanerQueueArn:
@@ -66,6 +67,7 @@ export const senderProdProps = {
 		maximumInstances: 6,
 	},
 	instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
+	targetCpuUtilization: 20,
 	notificationSnsTopic:
 		'arn:aws:sns:eu-west-1:201359054765:AutoscalingNotificationsPROD',
 	cleanerQueueArn:
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -63,8 +63,8 @@ export const senderProdProps = {
 	stack: 'mobile-notifications-workers',
 	stage: 'PROD',
 	asgCapacity: {
-		minimumInstances: 3,
-		maximumInstances: 6,
+		minimumInstances: 1,
+		maximumInstances: 2,
 	},
 	instanceType: InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL),
 	targetCpuUtilization: 20,
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -1,6 +1,8 @@
 import 'source-map-support/register';
 import { App } from 'aws-cdk-lib';
+import { InstanceClass, InstanceSize, InstanceType } from 'aws-cdk-lib/aws-ec2';
 import { RegistrationsDbProxy } from '../lib/registrations-db-proxy';
+import { SenderWorkerStack } from '../lib/sender-stack';
 import { SloMonitoring } from '../lib/slo-monitoring';
 
 const app = new App();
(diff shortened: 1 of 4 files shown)


cdk/lib/sender-stack.ts
```diff
@@ -0,0 +1,145 @@
...
+
+export class SenderWorkerStack extends GuStack {
+	constructor(scope: App, id: string, props: SenderStackProps) {
+		super(scope, id, props);
+
+		const sqsMessageVisibilityTimeout = Duration.seconds(100);
+		const sqsMessageRetentionPeriod = Duration.hours(1);
+		const sqsMessageRetryCount = 5;
+		const instanceType = InstanceType.of(InstanceClass.T4G, InstanceSize.SMALL);
+		const targetUtilizationPercent = 40;
```
DavidLawes, line 39: I don't have a feel for what this number should be. Did we choose this based on your experience or some advice?

cdk/lib/sender-stack.ts
```diff
@@ -0,0 +1,145 @@
...
+
+		const vpc = GuVpc.fromIdParameter(
+			this,
+			AppIdentity.suffixText({ app: props.appName }, 'VPC'),
+		);
+		const distributionRole = new GuInstanceRole(this, {
+			app: props.appName,
+			additionalPolicies: [
```
DavidLawes, line 47: do we get permission to emit logs (to cloudwatch?) by default?

cdk/bin/cdk.ts
```diff
@@ -38,3 +40,35 @@ new SloMonitoring(app, 'SloMonitor-PROD', {
 	stack: 'mobile-notifications',
 	stage: 'PROD',
 });
+
+export const senderCodeProps = {
+	appName: 'sender-worker',
+	stack: 'mobile-notifications-workers',
+	stage: 'CODE',
+	asgCapacity: {
+		minimumInstances: 1,
```
DavidLawes, line 49: during the testing phase, would we consider creating code and prod with the same settings?

DavidLawes (approved):
👍

cdk/lib/sender-stack.ts
waisingyiu: Sorry I should give some explanation.

`sqsMessageRetentionPeriod` refers to the duration for how long a message can stay in the queue.  After this period elapses, the message get deleted even if it is not processed.

`sqsMessageRetryCount` refers to the number of times a message can get picked up before being sent to DLQ.  So it can be seen as the maximum retry count.

For now, it seems sensible to use the same values for the old queues, but we can review them later.

For `instanceType`, I think we can start with a small box size that is reasonable.  `t4g.small` is chosen because it is used by Apps-rendering on `PROD`.  But definitely we can get a better idea what this value should be after we implement the sender and run some testing.

For `targetUtilizationPercent`, it defines the level of CPU utilisation at which the scaling policy scales up or down the autoscaling group.  You have raised a good point here.  `t4g.small` is a burstable instance class and its baseline performance is 20%.  Going over this usage consumes the CPU credit.  We should probably set the target CPU utilisation to 20.  (And it is the same value as in MAPI).

Please let me know if you have any comments.

DavidLawes: Ah, the target utilisation makes sense now, thank you :) Given how aws burst the instance, changing this value to 20 seems reasonable!

cdk/lib/sender-stack.ts
waisingyiu: The GuCDK class grants the permission to write to the aggregated log stream (`mobile-log-aggregation-${stage}` in mobile account).  Usually in EC2, programs write log to a log file, and an external program `aws-kinesis-agent` is used to move the logs to the kinesis stream.

cdk/bin/cdk.ts
waisingyiu: Yes, that sounds good for the testing phase.

PR #794, opened 2022-10-19
Removing blocking code from the firebase client

DavidLawes:
## What does this change?

In #729 we increased the thread pool size as a way of increasing performance in the notification system. This probably wouldn't have been necessary if we did not have blocking code.

As discussed [here](https://github.com/guardian/mobile-n10n/pull/729#issuecomment-1284068735) the blocking code has been identified and a solution discussed.

In this PR we've refactored sendNotification() and extracted parseSendResponse() so that we can test the response parsing more easily (without having to mock all the Future-related interactions).

## How to test

We have validated that the unit tests still correctly parse the response from firebase.

We ran the lambda locally and could see the future completing.

We should monitor the n10n [dashboard](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#dashboards:name=n10n-profiling) which shows lambda duration. After the change is released we should not expect to see a degradation in performance.

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
@@ -1,7 +1,7 @@
 package com.gu.notifications.worker.delivery.fcm
 
 import _root_.models.Notification
-import com.google.api.core.ApiFuture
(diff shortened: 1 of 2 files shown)


rtyley (approved):
@DavidLawes & I worked on this and I think it's a fairly innocuous change that should definitely benefit Android notifications!

I worked on the refactoring of the tests, so I'm probably not an ideal reviewer, but hey, now there are few lines! 
image

PR #796, opened 2022-10-24
LIVE-4705 set up deployment for EC2-based sender

waisingyiu:
## What does this change?

The previous PR #793 created a new cloudformation stack for the sender worker we are looking to build.

This PR set up a sub-project in the repo with the corresponding riff-raff configuration for deployment.

Specifically, it contains the following changes:

1. Create a new subproject `senderworker` which is a Scala project
2. It depends on `notificationworkerlambda`.  The JAR is included in the output package of `senderworker`
3. It is packaged as Debian package and deployed as a Systemd service (just like MAPI microservice)
4. At this moment, the Scala program does nothing (except printing some messages to confirm it is executed)
5. `AWS-kinesis-agent` is set up to copy logs from log file to the `mobile-aggregate-log-$(stage)` stream
6. Associate the EC2 instance with the default security group so that we can use ssm-ssh to log in the instance

## How to test

We deployed it to `CODE` and it worked as expected.

The stack is not used by existing programs in the notification delivery pipeline yet. So no impact to production is expected.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -441,7 +441,7 @@ lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notifica
     riffRaffArtifactResources += (file("cdk/cdk.out/RegistrationsDbProxy-PROD.template.json"), s"registrations-db-proxy-cfn/RegistrationsDbProxy-PROD.template.json")
 )
 
-lazy val ec2SenderWorker = Project("senderworker", file("senderworker"))
+lazy val ec2SenderWorker = Project("sender-worker", file("senderworker"))
   .dependsOn(common, notificationworkerlambda, commontest % "test->test")
   .enablePlugins(SystemdPlugin, RiffRaffArtifact, JavaServerAppPackaging)
   .settings(standardSettings: _*)
--- a/build.sbt
+++ b/build.sbt
@@ -441,6 +441,23 @@ lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notifica
     riffRaffArtifactResources += (file("cdk/cdk.out/RegistrationsDbProxy-PROD.template.json"), s"registrations-db-proxy-cfn/RegistrationsDbProxy-PROD.template.json")
 )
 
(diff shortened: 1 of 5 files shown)


frankie297 (approved):

senderworker/riff-raff.yaml
```diff
@@ -0,0 +1,23 @@
+stacks: [mobile-notifications-workers]
+regions: [eu-west-1]
+allowedStages:
+  - CODE
+  - PROD
+deployments:
+  senderworker-cfn:
+    type: cloud-formation
+    app: sender-worker
```
DavidLawes, line 9: Would we expect to see logs under this `appName` in kibana?

build.sbt
```diff
@@ -441,6 +441,23 @@ lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notifica
     riffRaffArtifactResources += (file("cdk/cdk.out/RegistrationsDbProxy-PROD.template.json"), s"registrations-db-proxy-cfn/RegistrationsDbProxy-PROD.template.json")
 )
 
+lazy val ec2SenderWorker = Project("sender-worker", file("senderworker"))
+  .dependsOn(common, notificationworkerlambda, commontest % "test->test")
```
DavidLawes, line 445: If my understanding is correct, if we depend on notificationworkerlambda then we can access the classes defined in that project inside the `sender-worker` project?

senderworker/riff-raff.yaml
waisingyiu: Not yet.  The logging configuration will be wired up in [LIVE-4736](https://theguardian.atlassian.net/browse/LIVE-4736).  After that, its logs are under this `appName`

build.sbt
waisingyiu: Yes, that's right.  Again there are more work needed to be able to call `notificationworkerlambda` - the SSM configuration things.  It will be handled in LIVE-4736.

PR #798, opened 2022-10-26
LIVE-4736 add configuration and logging to EC2 sender worker

waisingyiu:
## What does this change?

After PR #793 and PR  #796, we have a subproject which creates a EC2-based cloudformation stack and deploys a _empty_ `sender-worker` via `riff-raff`.

The PR wires up the configuration setting and logging facility and makes changes so that the sender worker lambda functions can be reused directly in the EC2-based sender.

## Logging

The _slf4j_ logging configuration is defined in `resources/logback.xml`.  Like the registrations API / notificatIons API, it writes logs to a log file on the file system (`sender-worker/sender-worker.log`).  In the EC2 instance, a `aws-kinesis-agent` is used to copy the logs from the log file to the Kinesis stream `mobile-log-aggregate-$(stage)`.  Logs are then shipped to Kibana.

## Configuration

It seems reasonable to expect that the lambda-based sender and EC2-based sender may have different configuration due to the difference in their underlying infrastructure.  A new set of configuration has been created for the EC2-based sender under the SSM parameter store `notifications/$(stage)/ec2workers`, but it has the same structure of parameter keys as lambda-based sender's.

## Local Development

Local deployment is enabled.  It requires a local configuration file to store the whole set of parameters.  A shell script (`scripts/create-local-conf-from-CODE.sh`) is provided to help creating one from the parameters in `CODE`:

1. Execute the shell script by: ```. ./scripts/create-local-conf-from-CODE.sh```
2. Copy the output to `~/.gu/sender-worker.conf`.  Note that it contains secret data such as the Firebase account key and APNS certificates.

The two steps above need to be done once only.

4. Run 'sbt'
5. Run 'project sender-worker` in `sbt` shell
6. Run `run`
 
## Changes to `notificationworkerlambda`

To allow the EC2 sender to use the sender functions defined in `notificationworkerlambda` directly, the following changes have been made to the lambda sender:

1. The `IOSSender` class has been changed to have another constructor which takes `ApnsWorkerConfiguration` config from the argument. 

2. The `AndroidSender` class has been changed to have another constructor which takes `FcmWorkerConfiguration` config and an optional FirebaseApp name from the argument.

3. The `FcmClient.apply` method accepts a new optional argument for the name of FirebaseApp.  Each FirebaseApp instance can be initialised once only, but for EC2 sender, it may instantiate multiple`FirebaseApp` for `Android`, `AndroidBeta` and `AndroidEdition`.  In this case, we need to initialise FirebaseApp with a unique name (unique within this process).

## How to test

The `AndroidSender` and `IOSSender` were instantiated successfully.  The sender attempted to send a test `Notification`  to an invalid token and the FIrebase / APNS returns an expected error of invalid token in both `CODE` and local environment.

Logs from `sender-worker` `CODE` can be found in `Kibana`.  Logs from local deployment can be found in `log/sender-worker.log`.

I will run a test of sending notifications on `CODE` to make sure that it has no impact on the existing lambda sender worker.


senderworker/src/main/scala/com/gu/notifications/ec2worker/Configuration.scala
```diff
@@ -0,0 +1,67 @@
...
+
+    FcmWorkerConfiguration(
+      config.getString("cleaningSqsUrl"),
+      FcmConfig(
+        serviceAccountKey = config.getString("fcm.serviceAccountKey"),
+        debug = config.getBoolean("fcm.debug"),
+        dryRun = config.getBoolean("dryrun")
+      ),
+      config.getInt("fcm.threadPoolSize"),
+      getStringList("fcm.allowedTopicsForBatchSend")
```
waisingyiu, line 62: It mainly duplicates the logic of building `FcmWorkerConfiguration` and `ApnsWorkerConfiguration` from the configuration, but the `Config` in EC2 sender contains all the configuration of each platform so the logic in lambda sender cannot be reused directly.

Given that we rarely add or remove configuration items and we will switch over to (or dispose of) EC2 sender at the end, maybe it is acceptable as a temporary solution.  But any idea would be great!

senderworker/scripts/create-local-conf-from-CODE.sh
```diff
@@ -0,0 +1,56 @@
+function readStringParameter() {
```
DavidLawes, line 1: This is a really useful function! It might be worth moving into a common/shared location because it's likely this would be useful for locally running the lambdas too

senderworker/src/main/resources/logback.xml
```diff
@@ -0,0 +1,26 @@
+<configuration>
```
DavidLawes, line 1: My understanding is that this logback configuration allows for structured json logging. Any existing logs we use to populate dashboards in kibana will continue to work, albeit they should be logged under a different `appName`.

In the sender lambdas though we emit metrics using the embedded log structure (logs that have a key `_aws`). I _think_ the logging configuration for ec2s means the logs are not parsed by cloudwatch so we will not emit the same metrics when the ec2 service is running.

Perhaps not a consideration for now, but if we're not parsing/emitting metrics via cloudwatch we may want to consider setting up a grafana dashboard that could consume/collate metrics from different sources, OR we could change the existing lambda functions to just use the PUT metric api instead of the structured logs...

senderworker/src/main/scala/com/gu/notifications/ec2worker/Configuration.scala
DavidLawes: I think this duplication is fair - like you said this should only exist in this medium-term. After we decide on an infrastructure (lambda or ec2) we won't need both Config files. 

Thank you for pointing this out though, hopefully we won't need to add new remote parameters, but if we do we will need to parse them in 2 places in the codebase (for now at least).

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
```diff
@@ -0,0 +1,87 @@
...
+  logger.info("Sender worker - AndroidBeta started")
+  val androidBetaSender = new AndroidSender(Configuration.fetchFirebase(config, AndroidBeta), Some(AndroidBeta.toString()))
+  androidBetaSender.handleChunkTokens(sqsEvent, null)
+
+  logger.info("Sender worker - AndroidEdition started")
+  val androidEditionSender = new AndroidSender(Configuration.fetchFirebase(config, AndroidEdition), Some(AndroidEdition.toString()))
+  androidEditionSender.handleChunkTokens(sqsEvent, null)
+
+  logger.info("Sender worker all started")
+  Thread.sleep(60*60*1000)
```
DavidLawes, line 86: For now, when the servers start we should expect to see in the logs these messages once? And then the process terminates/waits, but doesn't continue to `handleChunkTokens`? Maybe I've misunderstood the implementation

DavidLawes: I think this raises an interesting point, which may be worth testing/experimenting/validating. If an error is thrown in the loop that listens to the sqs queue, is the service automatically restarted, and yes after how long? I had a previous experience where unhandled errors when listening to the queue terminated the server but there was no auto-restart process (server remained dead)

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
```diff
@@ -125,17 +125,19 @@ class FcmClient (firebaseMessaging: FirebaseMessaging, firebaseApp: FirebaseApp,
...
     Try {
       val firebaseOptions: FirebaseOptions = FirebaseOptions.builder()
           .setCredentials(GoogleCredentials.fromStream(new ByteArrayInputStream(config.serviceAccountKey.getBytes)))
           .setHttpTransport(new OkGoogleHttpTransport)
           .setConnectTimeout(10000) // 10 seconds
           .build
-      FirebaseApp.initializeApp(firebaseOptions)
+      firebaseAppName match {
+        case None => FirebaseApp.initializeApp(firebaseOptions)
+        case Some(name) => FirebaseApp.initializeApp(firebaseOptions, name)
```
DavidLawes, line 137: I understand why we're doing this, to give each firebase client a unique name. Is this a restriction of the firebase sdk? I'm wondering why we don't need equivalent functionality for apns?

DavidLawes (approved):
👍

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
waisingyiu: Yes, it is a [requirement of FirebaseSDK](https://firebase.google.com/docs/reference/admin/java/reference/com/google/firebase/FirebaseApp).  

I guess it may be just a matter of implementation.  For apns, it appears that Apple provides a web service URLs instead of a SDK (I may be wrong) and each call to the web service is probably independent.  We use third party library `pushy` and it does not have this restriction.

senderworker/scripts/create-local-conf-from-CODE.sh
waisingyiu: It is great that it can be useful!  I move the script folder up a level to be directly under `mobile-n10n` directory

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
waisingyiu: In my first version where the service just quits after calling the sender stuff, it gets started up every minute.  But I also recall that I saw once the process was not started up again after some time.  So I suspect that we may not be able to rely on this auto-restart process completely, but need to implement some kind of [health check](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-health-checks.html#instance-health-detection) for the autoscaling group to start up a new EC2 instance when the process does not work properly.

One idea may be that the service sends some kinds of cloudwatch metric regularly inside the loop.  When the metric is missing for a certain period, the health check fails.

I have created a ticket to look into this. LIVE-4744](https://theguardian.atlassian.net/browse/LIVE-4744)

senderworker/src/main/resources/logback.xml
DavidLawes: I agree, it can be a follow-up task :)

senderworker/src/main/resources/logback.xml
```diff
@@ -0,0 +1,26 @@
...
+
+  <contextName>sender-worker</contextName>
+
+  <appender name="FILE" class="ch.qos.logback.core.rolling.RollingFileAppender">
+    <file>logs/sender-worker.log</file>
```
DavidLawes, line 6: As a small aside (just in case other people look at this PR), in other logback configs we've specified the log file destination as `logs/application.log`, which I think then makes the script in the user data a little easier to follow. For example, I've been wondering why the `appName` was repeated, but I guess it's because we've named the log file the same as the `appName`,

eg
`/opt/aws-kinesis-agent/configure-aws-kinesis-agent ${this.region} mobile-log-aggregation-${this.stage} /var/log/${props.appName}/${props.appName}.log`

becomes:
`/opt/aws-kinesis-agent/configure-aws-kinesis-agent ${this.region} mobile-log-aggregation-${this.stage} /var/log/${props.appName}/application.log`

waisingyiu: Thanks David.  This sounds good to me, as the `appName` has already been in the directory name.  It also makes the log file name consistent with other services in `n10n`.

senderworker/src/main/resources/logback.xml
waisingyiu: I have created a JIRA ticket [LIVE-4746](https://theguardian.atlassian.net/browse/LIVE-4746) for this follow-up task.

PR #803, opened 2022-11-01
Update parsing logic for multicast message response

DavidLawes:
## What does this change?

In https://github.com/guardian/mobile-n10n/pull/747 we enabled multicast messaging for specific topics. The multicast message endpoint allows us to send up to 500 tokens in a single http request (as opposed to a single token per http request in our pre-existing implementation). The intention was to test whether batching tokens would increase the rate at which lambdas could process tokens.

Batch sending has been limited to topics `breaking/international` and `breaking/international-sport`. Since deploying the previous change some notifications have been sent using the multicast message endpoint. The general results show:
- the 90in2 metric has not changed (but we don't expect it to because ios is excluded from this test).
- the number of android users receiving notifications for these two topics is as expected (indicating the sending of the notification via multicast messaging is taking place successfully).

However, the logs made validating the flow a little hard to follow, eg:

Screenshot 2022-11-01 at 14 09 40

We can see the logs report that only 2 notifications were successfully sent by the given lambda invocation. However, for the notification above, id `0e1280bb-62d7-41f6-b131-9d9e3beb0139`, we can see:
- sent to `breaking/international` topic
- there were 392 function invocations
- each function invocation processed an average of 867 tokens

Given we send to the multicast messaging endpoint in groups of 500 tokens, we'd expect most of the lambda invocations to make 2 http requests.

It appeared that the logging events handled only 1 response per batch. This PR addresses the parsing logic to ensure that every response from a multicast message is handled successfully - this is important because it affects not just the logging but also the cleaning of invalid tokens.

### Explanation

Looking at the sending flow, we currently write log events from this block of code:

```scala
        resultStream
          .broadcastTo(
            reportSuccesses(chunkedTokens, sentTime, functionStartTime),
            cleanupFailures,
            trackProgress(chunkedTokens.notification.id))
```

where `resultStream` is the stream of responses resulting from sending the push notification.

The `broadcastTo` function defines a series of pipes which accept the source stream (== `resultStream`). Each of these pipes accept the input type (`Stream[IO, Either[DeliveryException, DeliverySuccess]]`) and just execute side effects (reporting, logging, sending tokens for cleaning). It feels logical that there will only be elements to pull from the results stream when we start to receive http responses. 

We use [DeliveryService](https://github.com/guardian/mobile-n10n/blob/f85f3542112eed085a42315f2e2af98a99b2be71/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/DeliveryService.scala#L97) as a wrapper for making the http calls to firebase and apns. The [Async](https://typelevel.org/cats-effect/docs/typeclasses/async) wrapper is a cats-effect component and according to the docs it "creates a simple, non-cancelable ... instance that executes an asynchronous process on evaluation. The given function is being injected with a side-effectful callback for signalling the final result of an asynchronous operation".

This is important because cats-effect is providing us with a way of managing callbacks to the asynchronous process, the definition states that the callback signature should be `Either[Throwable, A] => ()`.

We use [FcmClient](https://github.com/guardian/mobile-n10n/blob/main/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala#L71) to send the batch notification to firebase and we can see that we invoke the callback function when the future has completed, e.g:

```scala
  def parseBatchSendResponse(
    notificationId: UUID, tokens: List[String], triedResponse: Try[BatchResponse]
  )(cb: Either[Throwable, Success] => Unit): Unit = triedResponse match {
    case Success(batchResponse) =>
      // From firebase sdk docs: the order of the response list corresponds to the order of the input tokens
      batchResponse.getResponses.asScala.toList.zip(tokens).foreach { el => {
        val (r, token) = el
        if (!r.isSuccessful) {
          parseSendResponse(notificationId, token, Failure(r.getException))(cb)
        } else {
          cb(Right(FcmDeliverySuccess(s"Token in batch response succeeded", token)))
        }
      }
    }
```

The parsed response has type `Either[Throwable, DeliverySuccess]` which matches the expected type for the inputs to our callback functions (the functions defined inside `broadcastTo`).

What I believe happens is:
- we make asynchronous call to firebase
- we define a callback that can be invoked when the response is received
- the responses are streamed onto a response stream
- the callbacks are equivalent to the "pipes" or functions defined inside `broadcastTo`

This is significant when considering the `parseSendResponse` function because we currently invoke the callback function for each response included in the batch:

```scala
  def parseSendResponse(
    notificationId: UUID, token: String, response: Try[String]
  )(cb: Either[Throwable, Success] => Unit): Unit = cb(response match {
    case Success(messageId) =>
      Right(FcmDeliverySuccess(token, messageId))
    ....
```

In the multicast message response our response stream is for a batch of individual responses, not just a single individual response (as it was previously).

What I think has happened is the callback has been invoked only once per batch, when the first token has been parsed.

In order to rectify this we could (as implemented in this PR), for multicast messages:
- invoke the callback after successfully parsing all the responses from the batch response
- update the logic for the callbacks to handle a List of parsed responses, instead of just a single parsed response

### Test locally

I did some testing locally:

Single batch
74 tokens included in fake sqs message (expected 1 http request to be made)
Only 2 were valid tokens
I could see expected result log
```
{
    "@timestamp": "2022-11-01T12:34:52.384Z",
    "message": "Processed 74 individual notification(s)",
},
{
    "@timestamp": "2022-11-01T12:34:52.389Z",
    "message": "Results (notification: 068b3d2b-dc9d-482b-a1c9-bd0f5dd8ebd7 ShardRange(0,1)): : Success: 2, Failure: 72, DryRun: 0 Total: 74",
    "notificationId": "068b3d2b-dc9d-482b-a1c9-bd0f5dd8ebd7",
}
```

Multiple batches:
74 tokens included in fake sqs message
Changed batch size to be 50 (expected 2 http requests to be made)

```
{
    "@timestamp": "2022-11-01T15:15:52.571Z",
    "message": "Processed 24 individual notification(s)",
}
{
    "@timestamp": "2022-11-01T15:15:52.648Z",
    "message": "Processed 50 individual notification(s)",
}
{
    "@timestamp": "2022-11-01T15:15:52.653Z",
    "message": "Results (notification: 068b3d2b-dc9d-482b-a1c9-bd0f5dd8ebd7 ShardRange(0,1)): : Success: 2, Failure: 72, DryRun: 0 Total: 74",
    "notificationId": "068b3d2b-dc9d-482b-a1c9-bd0f5dd8ebd7",
}
```

### Test in CODE 

For a dryRun notification (pointing to `testRig` db) the logs appear as expected:

Screenshot 2022-11-01 at 15 38 00

^ we can see that the tokens have been correctly grouped and the parsing logic for the response has been corrected.

For a real notification send (pointing to `code` db) we expect `breaking/international` topic to send to 8 devices, 6 of which are android.

From the logs we can see that, due to the sharding, we invoke 6 lambda functions, each with their own set of expected results:

Screenshot 2022-11-01 at 16 01 08

^ although this doesn't help validate successful batch parsing like the prior test, it does confirm that the multicast message still correctly sends the notification.

Given this change is deployed we expect to see:
- log messages revealing the expected number of tokens (not batches) processed by each lambda invocation
- log messages revealing that _some_ tokens were invalid and processed accordingly

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
@@ -1,6 +1,6 @@
 package com.gu.notifications.worker.models
 
-import com.gu.notifications.worker.delivery.{DeliveryException, DeliverySuccess}
+import com.gu.notifications.worker.delivery.{BatchDeliverySuccess, DeliveryException, DeliverySuccess}
 
 case class SendingResults(
   successCount: Int,
@@ -19,6 +19,12 @@ object SendingResults {
     case Right(success) if success.dryRun => previous.copy(dryRunCount = previous.dryRunCount + 1)
     case Right(_) => previous.copy(successCount = previous.successCount + 1)
     case Left(_) => previous.copy(failureCount = previous.failureCount + 1)
+  }
+
+  def aggregateBatch(previous: SendingResults, batchSize: Int, res: Either[Throwable, BatchDeliverySuccess]): SendingResults = res match {
+    case Right(batchSuccess) =>
+      batchSuccess.responses.foldLeft(previous)((acc, resp) => SendingResults.aggregate(acc, resp))
+    case Left(_) => SendingResults(0, batchSize, 0)
   }
 }
 
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
@@ -24,7 +24,7 @@ object SendingResults {
   def aggregateBatch(previous: SendingResults, batchSize: Int, res: Either[Throwable, BatchDeliverySuccess]): SendingResults = res match {
     case Right(batchSuccess) =>
       batchSuccess.responses.foldLeft(previous)((acc, resp) => SendingResults.aggregate(acc, resp))
-    case Left(_) => SendingResults(0, batchSize, 0)
+    case Left(_) => SendingResults(previous.successCount, previous.failureCount + batchSize, previous.dryRunCount)
   }
 }
 
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -70,7 +70,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
     val notificationLog = s"(notification: ${chunkedTokens.notification.id} ${chunkedTokens.range})"
     input
       .fold(SendingResults.empty) { case (acc, resp) => SendingResults.aggregateBatch(acc, chunkedTokens.tokens.size, resp) }
-      .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform), prefix = s"Results $notificationLog: "))
+      .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform, isIndividualNotificationSend = false), prefix = s"Results $notificationLog: "))
       .through(cloudwatch.sendMetrics(env.stage, Configuration.platform))
   }
 
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -1,20 +1,19 @@
 package com.gu.notifications.worker
 
 import cats.effect.{ContextShift, IO, Timer}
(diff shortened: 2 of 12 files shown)


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
```diff
@@ -20,5 +20,11 @@ object SendingResults {
     case Right(_) => previous.copy(successCount = previous.successCount + 1)
     case Left(_) => previous.copy(failureCount = previous.failureCount + 1)
   }
+
+  def aggregateBatch(previous: SendingResults, batchSize: Int, res: Either[Throwable, BatchDeliverySuccess]): SendingResults = res match {
+    case Right(batchSuccess) =>
+      batchSuccess.responses.foldLeft(previous)((acc, resp) => SendingResults.aggregate(acc, resp))
+    case Left(_) => SendingResults(0, batchSize, 0)
```
waisingyiu, line 27: I just wonder why the `previous` result is completely discarded when the current result is failure?

DavidLawes: Good catch - I'll update :)

waisingyiu (approved):
👍

PR #805, opened 2022-11-02
LIVE-4704: Add ec2 worker sender queue url parameter

frankie297:
## What does this change?
Adds a new parameter to store the EC2 Workers SQS Queues URLs. This is so we can use the parameter to retrieve messages from the SQS queues. We know the parameter already exists for the harvester, the neatest way is to duplicate that parameter and modify the name for the ec2 worker namespace.

## How to test
Deploy to CODE and see if the parameter is generated


DavidLawes (approved):
This was deployed to CODE and we could see the corresponding ssm parameters had been successfully created.

Given the parameters now exist we can reference the queue urls in the ec2worker config, which is needed when polling for sqs messages

waisingyiu (approved):
It sounds like a good idea!

PR #807, opened 2022-11-03
Log multicast message request latency

DavidLawes:
## What does this change?

This adds a log event when the multicast message api request has completed. The log event includes the latency of the api call. 

Knowing the latency of the api call itself will help us validate that the function processing rates we're calculating are valid (they're currently much larger than expected).

I actively didn't include the same log event for the latency of the individual notification send because we already have this kind of data via the firebase google cloud console.

I tried to visualise the latency of the multicast message api call via the firebase google cloud console, but the latency appears to be limited to the `SendMessage` endpoint (or the latency shown is aggregated across all push notification endpoints being used, it's not clear).

Screenshot 2022-11-03 at 09 27 44

NB: The latency we can see from the graph above correlates with the general timings we're seeing when sending individual notifications.

## How to test

I tested locally and I could see the log message being generated when the api call had completed, the log event included the batch request latency.

```
{
    "@timestamp": "2022-11-03T09:31:57.246Z",
    "@version": "1",
    "message": "Batch send request completed",
    "logger_name": "com.gu.notifications.worker.delivery.fcm.FcmClient",
    "thread_name": "pool-33-thread-8",
    "level": "INFO",
    "level_value": 20000,
    "worker.batchRequestLatency": 571,
}
```


waisingyiu (approved):
👍

PR #809, opened 2022-11-03
LIVE-4704: Add sqs url parameters to apns and firebase config

frankie297:
## What does this change?
Adds ec2 url parameter to apns and firebase config, this will be used later for polling a queue to receive messages. Also adds an sqs url parameter to the old worker config.

## How to test
Deploy to CODE and check it loads configuration correctly, check the lambda still works as expected.


DavidLawes (approved):
👍

PR #810, opened 2022-11-03
LIVE-4701 create stack switch in harvester

waisingyiu:
## What does this change?

In option 1, we plan to have both the lambda sender worker and EC2-based sender worker running in parallel during the migration and testing phase, and we route the notification tokens based on the topics of the notification.

This PR implements this stack switching feature in the harvester worker.

It makes the following changes
1. Read the new parameters about the URLs of the new SQSs from SSM
2. Read the new parameter `allowedTopicsForEc2Sender`
3. Instantiate two sets of `SqsDeliveryService`, for the old SQSs and for the new ones.
4. Dispatch the tokens to the set of `SqsDeliveryService` based on the topic.  If all topics of the notification are on the list, it goes to the new stack.
5. Update test cases

### Configuration

The parameters for the URLs of the new SQS were created in this PR #793 

The parameter path for allowed topics is
`/notifications/CODE/workers/harvester/allowedTopicsForEc2Sender` for `CODE` and
`/notifications/PROD/workers/harvester/allowedTopicsForEc2Sender` for `PROD`.

This parameter is optional.  If this parameter does not exist, it means no topics are allowed to use the EC2 stack.

Its type is string.  For multiple topics, it should be a comma-separated list.

## How to test

Test cases have been added to test the stack switching logics.  It has also been tested on CODE.

### DIFF
--- a/notificationworkerlambda/harvester-cfn.yaml
+++ b/notificationworkerlambda/harvester-cfn.yaml
@@ -115,6 +115,16 @@ Resources:
                 - ","
                 - "Fn::ImportValue":
                     "Fn::Sub": 'NotificationSenderWorkerQueueArns-${Stage}'
+      - PolicyName: SQSOutputForEc2
+        PolicyDocument:
+          Statement:
+            Effect: Allow
+            Action: sqs:*
+            Resource:
+              "Fn::Split":
+                - ","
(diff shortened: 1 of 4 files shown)


groakland (approved):
LGTM

PR #811, opened 2022-11-03
Log individual notification request latency

DavidLawes:
## What does this change?

In https://github.com/guardian/mobile-n10n/pull/807 we started logging the latency of batch requests to firebase. To make an easy/fair comparison between the two firebase apis we use to send notifications, in this PR we add a similar log event for the latency of individual notification send requests.

This change will add a single log event per token we send. Therefore we will collect a lot of data very quickly about the general request latency. My intention is to revert this change by the end of today when we have enough data.

## How to test

I tested locally, the logs show the new log event:

```
{
    "@timestamp": "2022-11-03T11:54:32.51Z",
    "@version": "1",
    "message": "Individual send request completed",
    "logger_name": "com.gu.notifications.worker.delivery.fcm.FcmClient",
    "thread_name": "pool-36-thread-18",
    "level": "INFO",
    "level_value": 20000,
    "worker.individualRequestLatency": 2455,
}
```


waisingyiu (approved):
👍

PR #818, opened 2022-11-10
LIVE-4746 use AWS metric API in sender lambda

waisingyiu:
## What does this change?

The current implementation of sender lambda relies on Cloudwatch Logs to emit metrics.  When this implementation is invoked in EC2-based sender, the log events go to log files (instead of Cloudwatch Logs), and then they are shipped to Kibana via Kinesis stream.  As a result, metrics are not emitted.

Note that even without this change, we can still capture the metric data as they are sent to Kibana via log events.

To support the upcoming experimentation with EC2-based sender, it may be useful to emit metrics to AWS as metrics are retained in the AWS for much longer than they are in Kibana.

However, it makes additional API call to AWS to send metrics.  So if we revert to sender lambda at the end, we may consider changing back to the way we emit metrics via Cloudwatch Logs.

## Changes
- define a case class for the performance metrics
- create a function to emit metrics via AWS API
- pass the performance metrics data from the logging method to the new function created in the last point

## How to test

I deployed to CODE and sent a `dryRun` notification (temporary enabled AWS metric for `dryRun` mode).  The AWS metrics are validated against the corresponding metrics in Kibana.

Screenshot 2022-11-10 at 12 41 52

Screenshot 2022-11-10 at 12 42 16


DavidLawes:
Sorry in advance, I've probably missed something obvious 😅 ... will there be a way to differentiate the metrics in aws between the lambda workers and the ec2 workers? (and if we can't differentiate between them, is that important?)

Within kibana I think we will be able to segregate the log events by `app.keyword`. The `app` would be `sender-worker` for the ec2 workers? And for the lambda workers I think `app` is the same as `platform`, so "android", "ios", etc. In kibana we could therefore generate aggregations for the key system metrics for both the ec2 workers and the lambda workers (I think?).

I couldn't work out if the same would be true for the aws metrics? I thought that, for the current metrics, the relevant dimension was the `platform` ("android", "ios", etc), but I didn't _think_ this would differentiate between the app (lambdas or sender-worker)?

waisingyiu:
Thank you David for your comments!  I totally agree with you that at this moment we are able to get all the metrics from Kibana (no matter we run the sender in lambda or EC2), and we are able to differentiate the metrics between them.

I guess one of the reasons we sent the metrics to AWS in the first place is that we could look at the metrics further back because the retention period is longer in AWS.  It appears that it _may be_ useful when we run the EC2 worker experiment on PROD because it may last for quite a while.  Please let me know if I got it wrong.

And I agree that there is no way to differentiate the metrics in AWS between the lambda workers and the EC2 workers yet.  To make it useful, we may need to change the metric a little, for instance, adding a new dimension.  My initial thought is that this PR just aims to restore the ability to emit metric when running in EC2.  [Further work](https://theguardian.atlassian.net/browse/LIVE-4707) is probably needed to design what additional metrics are needed for the experiment.

Happy to withhold if you think otherwise!

DavidLawes:
> Further work

Ahh, amazing! I hadn't realised there were follow-on tasks. Given there's something in the pipeline to potentially add another dimension, then this all seems very sensible :)

DavidLawes (approved):
👍

PR #823, opened 2022-11-15
Process multiple sqs event records concurrently

DavidLawes:
## What does this change?

This PR is intended to test the performance improvement of multicast message sending by enabling more multicast message http requests to take place concurrently. The theory is that we have the ability to parallelise within our lambda, but it is currently under-used when sending groups of tokens to firebase.

We want to increase the function processing rate of our worker lambdas because they are the bottleneck for achieving our 90in2 SLO. We recently tried [testing](https://github.com/guardian/mobile-n10n/pull/747) the impact of multicast message sending our tokens to firebase. We observed the following results:
- the latency of a multicast message request was higher than the latency of a request to send an individual token (2600ms vs 70ms, respectively)
- the latency for sending 500 tokens is only ~38 times greater than sending for an individual token, so there is a potential efficiency gain by sending in batches, however:
- tokens were sent in groups of 500, meaning each lambda invocation would make, at most, only 2 http requests in parallel
- given the parallelism we already achieve, the overall result meant that making many shorter requests concurrently was quicker than making fewer longer requests (verified because the average lambda duration is typically smaller than the latency of a single average multicast message request).

NB: thinking about the parallelism we already achieve when sending individual notifications, we could estimate this as:
- ( average latency of a request * average number of requests made) / average total duration of lambda execution
- ( 70 ms * 1000 ) / 2400 ms
- ~29

There are some options to increase the number of parallel batch notification http requests:
- increase the number of tokens each sqs event contains: we cannot significantly increase the number of tokens in each single sqs message due to the sqs message limit (256kb, but we already send about 140kb of data)
- increase the batch size (the number of messages each lambda invocation processes): this is a good option but when testing we could see a linear relationship between duration and batch size, indicating that batches of sqs messages were being processed sequentially, not concurrently. 

A change to the `deliverChunkTokens` functions has been made to ensure multiple sqs events are processed concurrently. This was confirmed via local testing. A test sqs event was configured to included 3 sqs messages (where each message contained only 1 token). The below are extracts from the logs showing the sequence before and after this change:

| Before change                                                                                                              | After change                                                                                                                 |
|----------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------|
| Parsed chunk tokens                                                                                                        | "Parsed chunk tokens"                                                                                                        |
| Parsed chunk tokens                                                                                                        | "Parsed chunk tokens"                                                                                                        |
| Parsed chunk tokens                                                                                                        | "Parsed chunk tokens"                                                                                                        |
| Sending batch notification                                                                                                 | "Sending batch notification"                                                                                                 |
| Batch send request completed                                                                                               | "Sending batch notification"                                                                                                 |
| Processed 1 individual notification(s)                                                                                     | "Sending batch notification"                                                                                                 |
| Results (notification: 1145fad9-4825-4b95-9696-233ea1cdefaa ShardRange(0,1)): : Success: 0, Failure: 1, DryRun: 0 Total: 1 | "Batch send request completed"                                                                                               |
| Sent 0 tokens for deletion via SQS                                                                                         | "Processed 1 individual notification(s)"                                                                                     |
| Sending batch notification                                                                                                 | "Results (notification: ebe9a525-431f-4333-8277-b6f31b818439 ShardRange(0,1)): : Success: 0, Failure: 1, DryRun: 0 Total: 1" |
| Batch send request completed                                                                                               | "Sent 0 tokens for deletion via SQS"                                                                                         |
| Processed 1 individual notification(s)                                                                                     | "Batch send request completed"                                                                                               |
| Results (notification: d4953d04-58e3-42e3-b057-d5a2ced68641 ShardRange(0,1)): : Success: 0, Failure: 1, DryRun: 0 Total: 1 | "Processed 1 individual notification(s)"                                                                                     |
| Sent 0 tokens for deletion via SQS                                                                                         | "Results (notification: ed9e7175-8d7d-4586-a499-a4d3038251aa ShardRange(0,1)): : Success: 0, Failure: 1, DryRun: 0 Total: 1" |
| Sending batch notification                                                                                                 | "Sent 0 tokens for deletion via SQS"                                                                                         |
| Batch send request completed                                                                                               | "Batch send request completed"                                                                                               |
| Processed 1 individual notification(s)                                                                                     | "Processed 1 individual notification(s)"                                                                                     |
| Results (notification: 8b726f8c-989c-4e9e-a24b-8e05dc1701f3 ShardRange(0,1)): : Success: 0, Failure: 1, DryRun: 0 Total: 1 | "Results (notification: c861d37b-659d-46fe-9201-14b4ae6bf307 ShardRange(0,1)): : Success: 0, Failure: 1, DryRun: 0 Total: 1" |
| Sent 0 tokens for deletion via SQS                                                                                         | "Sent 0 tokens for deletion via SQS"                                                                                         |

^ here we can see that before the change we waited for the the request(s) associated with a single sqs message to return before attempting to process any subsequent sqs messages. After the change we can see all associated requests being made concurrently at the beginning of the process.

Based on the levels of parallelism we have for individual notification sends we believe that a single lambda invocation could easily parallelise 20 sqs messages. This significantly increases our token processing rate (estimated at 7400 tokens/s if all multicast messages were initiated concurrently and took the average expected time).

## How to test
 
I've tested locally and the change does what's intended (processes multiple messages concurrently).

We will also confirm the changes work as anticipated in the CODE environment.

We should expect to see an improvement in the performance of batch notification sends - we already have metrics in place to monitor this (including function processing rate, overall duration, etc).

We are increasing the work each lambda invocation has to do which will (likely) increase the total lambda execution time (although hopefully less than a factor of 20). Our current lambda timeout is currently unchanged but open to opinions about whether this timeout should be increased. For ref:
- current timeout = 90s
- 99% of lambda invocations complete within 20s
- average lambda duration tends to be about 2.4s

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -36,22 +36,22 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
   override val maxConcurrency = 100
 
   //override the deliverChunkedTokens method to validate the success of sending batch notifications to the FCM client. This implementation could be refactored in the future to make it more streamlined with APNs
-  override def deliverChunkedTokens(chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant)]): Stream[IO, Unit] = {
+  override def deliverChunkedTokens(chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant, Int)]): Stream[IO, Unit] = {
     chunkedTokenStream.map {
-      case (chunkedTokens, sentTime, functionStartTime) =>
+      case (chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize) =>
         val isAllowedToSendBatch = chunkedTokens.notification.topic.forall(topic => config.allowedTopicsForBatchSend.contains(topic.toString))
 
         if (isAllowedToSendBatch) {
           logger.info(s"Allowed to send notification ${chunkedTokens.notification.id} to topics ${chunkedTokens.notification.topic} in batches")
           deliverBatchNotificationStream(Stream.emits(chunkedTokens.toBatchNotificationToSends).covary[IO])
             .broadcastTo(
-              reportBatchSuccesses(chunkedTokens, sentTime, functionStartTime),
+              reportBatchSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
               cleanupBatchFailures(chunkedTokens.notification.id),
               trackBatchProgress(chunkedTokens.notification.id))
         } else
           deliverIndividualNotificationStream(Stream.emits(chunkedTokens.toNotificationToSends).covary[IO])
             .broadcastTo(
-              reportSuccesses(chunkedTokens, sentTime, functionStartTime),
+              reportSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
               cleanupFailures,
               trackProgress(chunkedTokens.notification.id))
     }.parJoin(maxConcurrency)
@@ -64,11 +64,11 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
       .evalTap(Reporting.logBatch(s"Sending failure: "))
   } yield resp
 
-  def reportBatchSuccesses[C <: DeliveryClient](chunkedTokens: ChunkedTokens, sentTime: Long, functionStartTime: Instant): Pipe[IO, Either[DeliveryException, BatchDeliverySuccess], Unit] = { input =>
+  def reportBatchSuccesses[C <: DeliveryClient](chunkedTokens: ChunkedTokens, sentTime: Long, functionStartTime: Instant, sqsMessageBatchSize: Int): Pipe[IO, Either[DeliveryException, BatchDeliverySuccess], Unit] = { input =>
     val notificationLog = s"(notification: ${chunkedTokens.notification.id} ${chunkedTokens.range})"
     input
       .fold(SendingResults.empty) { case (acc, resp) => SendingResults.aggregateBatch(acc, chunkedTokens.tokens.size, resp) }
-      .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform, isIndividualNotificationSend = false), prefix = s"Results $notificationLog: "))
+      .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform, isIndividualNotificationSend = false, sqsMessageBatchSize), prefix = s"Results $notificationLog: "))
       .through(cloudwatch.sendMetrics(env.stage, Configuration.platform))
   }
 
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -37,10 +37,10 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
 
   //override the deliverChunkedTokens method to validate the success of sending batch notifications to the FCM client. This implementation could be refactored in the future to make it more streamlined with APNs
   override def deliverChunkedTokens(chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant)]): Stream[IO, Unit] = {
-    for {
-      (chunkedTokens, sentTime, functionStartTime) <- chunkedTokenStream
-      isAllowedToSendBatch = chunkedTokens.notification.topic.forall(topic => config.allowedTopicsForBatchSend.contains(topic.toString))
-      resp <- {
+    chunkedTokenStream.map {
+      case (chunkedTokens, sentTime, functionStartTime) =>
+        val isAllowedToSendBatch = chunkedTokens.notification.topic.forall(topic => config.allowedTopicsForBatchSend.contains(topic.toString))
+
         if (isAllowedToSendBatch) {
           logger.info(s"Allowed to send notification ${chunkedTokens.notification.id} to topics ${chunkedTokens.notification.topic} in batches")
           deliverBatchNotificationStream(Stream.emits(chunkedTokens.toBatchNotificationToSends).covary[IO])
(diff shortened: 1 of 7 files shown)


waisingyiu:
Thank you David for the clear explanation.  It sounds a very brilliant idea!

As an aside only, if I understand correctly, one of the impacts may be that if a lambda invocation fails for whatever reasons, all 20 batches will become visible in the queue even if some of them may have been processed by the lambda.

For the lambda timeout, I suspect that we need to gather some statistics on the actual execution duration after running it on PROD in order to determine what value this timeout should be.  My feeling is that if a lambda function times out, it usually results from some exceptions (such as network connection problem) rather than from the normal processing of data.

Just in case if we want to change the lambda timeout, (I guess you are probably aware of it, but just in case) we need to change the visibility timeout of the SQS as well

waisingyiu (approved):

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -37,26 +37,24 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
...
-        else
+        } else
           deliverIndividualNotificationStream(Stream.emits(chunkedTokens.toNotificationToSends).covary[IO])
             .broadcastTo(
               reportSuccesses(chunkedTokens, sentTime, functionStartTime),
               cleanupFailures,
               trackProgress(chunkedTokens.notification.id))
-      }
-    } yield resp
+    }.parJoin(maxConcurrency)
```
rtyley, line 57: This is a change that becomes necessary as we increase the SQS batchSize of records above 1 : we don't want to process those records (each containing at most 1000 device ids) one after the other, we want to process them in parallel!

The [`parJoin()`](https://github.com/typelevel/fs2/blob/86e27ca52fc5767c066b91b55d30357764ad8401/core/shared/src/main/scala/fs2/Stream.scala#L4014-L4044) method merges a stream of streams (the stream of records, each record yielding a stream of tokens) and runs multiple record streams in parallel, rather than all sequentially. We already use `parJoin` in several places, eg:

https://github.com/guardian/mobile-n10n/blob/07259d294e34b184c41ecf34a0a4785e79d11332/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala#L64-L67

Note that the change in this file only affects the `AndroidSender`, which thanks to #747 already has a custom overridden `deliverChunkedTokens()` method to allow it to batch requests. To keep iOS in line with the new `parJoin()` behaviour, the original method in the `SenderRequestHandler` base trait has also been updated.

rtyley (approved):
@DavidLawes & I paired on this, so I'm approving it but I'm sure someone else from the MSS team will want to take a look!

frankie297 (approved):

DavidLawes:
I did some testing on CODE:
- We cannot increase the batch size without also defining a [batch window](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html).
- The batch window defines an duration during which we attempt to gather messages before invoking the lambda function.
- The minimum we can set the batch window is 1 second, which is what I've gone for, i.e. we'll wait a second to attempt to gather 20 messages from the queue before invoking the lambda function.
- Looking at the sqs metrics, we know from age of the messages on the worker queues, and the number of messages visible, that during peak times we currently have more messages on the queue than we can instantaneously process.
- Given we have lots of messages in the queue at any one time, it seems reasonable that during peak processing we'll easily be able to gather 20 messages from the queue  in 1 second for a single lambda invocation to then process.
- I'll add a log event to capture the number of messages each lambda invocation processes incase we need to tune the batch window later on.

Screenshots from the PROD android worker queue metrics:

Screenshot 2022-11-16 at 12 52 21

Screenshot 2022-11-16 at 12 52 07

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/NotificationParser.scala
```diff
@@ -39,13 +39,7 @@ object NotificationParser extends Logging {
 
   private def parseChunkedTokens(json: JsValue): ChunkedTokens = {
     json.validate[ChunkedTokens] match {
-      case JsSuccess(chunkedTokens, _) => {
```
DavidLawes, line 42: The number of tokens each sqs message contains is now included in a downstream log event, so this isn't needed anymore

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
```diff
@@ -68,26 +68,35 @@ trait SenderRequestHandler[C <: DeliveryClient] extends Logging {
...
 
     deliverChunkedTokens(chunkedTokenStream)
       .compile
       .drain
       .unsafeRunSync()
+
+    logger.info(Map(
+      "sqsMessageBatchSize" -> sqsMessageBatchSize,
+      "totalTokensProcessed" -> totalTokensProcessed,
+      "invocation.functionProcessingRate" -> { totalTokensProcessed.toDouble / Duration.between(startTime, Instant.now).toMillis * 1000 },
```
DavidLawes, line 99: I thought this was important to capture, to get an understanding of how well the _entire_ invocation performs

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
```diff
@@ -68,26 +68,35 @@ trait SenderRequestHandler[C <: DeliveryClient] extends Logging {
...
 
   def handleChunkTokens(event: SQSEvent, context: Context): Unit = {
-    val chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant)] = Stream.emits(event.getRecords.asScala)
+    val sqsMessageBatchSize = event.getRecords.size
+    val startTime = Instant.now
+    val totalTokensProcessed: Int = event.getRecords.asScala.map(event => NotificationParser.parseChunkedTokenEvent(event.getBody)).foldLeft(0)((acc, r) => acc + r.tokens.size)
+    val chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant, Int)] = Stream.emits(event.getRecords.asScala)
       .map(r => (r.getBody, r.getAttributes.getOrDefault("SentTimestamp", "0").toLong))
-      .map { case (body, sentTimestamp) => (NotificationParser.parseChunkedTokenEvent(body), sentTimestamp, Instant.now) }
+      .map { case (body, sentTimestamp) => (NotificationParser.parseChunkedTokenEvent(body), sentTimestamp, startTime, sqsMessageBatchSize) }
```
DavidLawes, line 89: Ensuring the sqsMessageBatchSize is included in our log event was a fiddle to implement in this current configuration. However, we have a ticket in our backlog to improve how we're achieving observability within this project.

PR #826, opened 2022-11-16
LIVE-4771: Add the ability for EC2 sender worker to continuously poll the Sender SQS queues for new messages

frankie297:
## What does this change?
This PR builds upon our previous implementation https://github.com/guardian/mobile-n10n/pull/814 which could process events and deletions successfully but we needed the ability for the EC2 workers to listen for changes to SQS queues, to pick up new messages that are received. 

I initially tried to implement a continuous loop, but have decided to take advantage of Amazons JMS ([Java Message Service)](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-java-message-service-jms-client.html). JMS is a service which allows us to create connections and sessions with SQS queues, as well as a consumer to send and receive messages from an SQS queue. But crucially, it allows us to create listeners to poll continuously SQS queues.

This PR achieves continuous polling with the following:

- Updates the AWS SDK version to make sure we can use the JMS library
- Adds the JMS library as a dependency
- JMS library requires the queue name rather than the queue url, so a new parameter is created in CDK
- The parameter is also added for the worker as well as the ec2 worker
- The `OpenConnection` class uses the JMS library to configure and open a connection to the sender worker SQS queues
- `ReceiverCallback` class parses a JMS message (received from a queue) to an SQS message and returns an SQS event.
- This SQS event can then be passed in to our `onMessage` function, which calls the `handleChunkTokens` method with our parsed event
- `message.acknowledge()` deletes a message from the queue after receipt and successful processing. (when we see a notification has been successfully sent)

## How to test
Tested on CODE and saw expected logs

- The EC2 sender worker can successfully receive messages from the worker SQS queues
- The EC2 sender worker can successfully process and send notifications to the APNs and Firebase endpoints
- Unsuccessful messages/ messages that cannot be processed are sent to the relevant DLQ, and do not crash the service

There might be increased costs with continuous polling. We might want to consider how we can open and close connections more dynamically, if costs are too expensive. The library allows us to listen to the queues for a period of time (a waiting time for queue silence) before closing a connection. We could do something similar in a future PR? How would we open the connection again?

Example code here from JMS:
```        public void waitForOneMinuteOfSilence() throws InterruptedException {
            for(;;) {
                long timeSinceLastMessage = System.nanoTime() - timeOfLastMessage;
                long remainingTillOneMinuteOfSilence = 
                    TimeUnit.MINUTES.toNanos(1) - timeSinceLastMessage;
                if( remainingTillOneMinuteOfSilence < 0 ) {
                    break;
                }
                TimeUnit.NANOSECONDS.sleep(remainingTillOneMinuteOfSilence);
            }

        callback.waitForOneMinuteOfSilence();
        System.out.println( "Returning after one minute of silence" );

        // Close the connection. This closes the session automatically
        connection.close();
        System.out.println( "Connection closed" );

```


build.sbt
```diff
@@ -421,8 +421,9 @@ lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notifica
       "com.google.firebase" % "firebase-admin" % "9.0.0",
       "com.google.protobuf" % "protobuf-java" % "3.19.2",
       "com.amazonaws" % "aws-lambda-java-events" % "2.2.8",
-      "com.amazonaws" % "aws-java-sdk-sqs" % awsSdkVersion,
+      "com.amazonaws" % "aws-java-sdk-sqs" % "1.12.338",
```
DavidLawes, line 424: Out of curiosity (most because I can't remember), did we need to bump just this library specifically?

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
```diff
@@ -4,57 +4,91 @@ import com.amazonaws.services.lambda.runtime.events.SQSEvent.SQSMessage
...
+    override def onMessage(message: jms.Message): Unit = {
+      println("JMS message received")
+      val parsedEvent = parsedReceivedMessages(message)
+      parsedEvent match {
+        case Some(event) => {
+          handleChunkTokens(event, null)
+          message.acknowledge()
+        }
+        case None => {
+          logger.error("Cannot parse message")
```
DavidLawes, line 76: We tried to test what happens if we reached this branch, but we couldn't (why testing locally). For now, we think what'll happen is:
- we'll log a message
- no error will be thrown
- no message.acknowledge --> message remains on queue (?)

We thought that maybe this would be thought about a little more at a later stage

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
```diff
@@ -4,57 +4,91 @@ import com.amazonaws.services.lambda.runtime.events.SQSEvent.SQSMessage
...
     }
 
-    val parsedEvent: SQSEvent = parsedReceivedMessages(receiveMessages)
+    override def onMessage(message: jms.Message): Unit = {
+      println("JMS message received")
+      val parsedEvent = parsedReceivedMessages(message)
+      parsedEvent match {
+        case Some(event) => {
+          handleChunkTokens(event, null)
+          message.acknowledge()
```
DavidLawes, line 73: If handleChunkTokens throws an error:
- we do not acknowledge the message
- message remains on queue
- message is reprocessed by the sender-worker
- if the message cannot be processed 5 times then it is put on the dlq (according to our cdk configuration)

DavidLawes (approved):

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
```diff
@@ -4,57 +4,91 @@ import com.amazonaws.services.lambda.runtime.events.SQSEvent.SQSMessage
...
-    parsedEvent.getRecords.forEach(m => {
-      sqsClient.deleteMessage(queue, m.getReceiptHandle)
-      logger.info("Successfully deleted message")
-    })
+    val connectionCallback = new ReceiverCallback(queueUrl, handleChunkTokens)
 
-    logger.info(s"Finished polling $appName")
+    new OpenConnection(queueName) {
+      consumer.setMessageListener(connectionCallback)
+    }
```
waisingyiu, line 89: If a message is received from SQS, will the callback function be executed in the same thread or in another thread?  I just wonder that if there are multiple messages, will they be handled concurrently?

If the callback are executed in separate threads, is it possible for us to tune the size of thread pool?

frankie297: I have now made a change so the `onMessage` function returns a future, which means the messages are received asynchronously and using different threads. It is using the thread pool size of the lambda, which is currently 50. Do you think this should be more?

| Application Logs Before | Application Logs After |
| --- | --- |
|| |

build.sbt
frankie297: I have now bumped all the libraries that rely on the aws sdk version

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
DavidLawes: It's a great question - in our current setup we define the ec2 instance type as [t4g small](https://aws.amazon.com/ec2/instance-types/?trk=adf88fbf-e631-4352-979f-e3812ceb7322&sc_channel=ps&s_kwcid=AL!4422!3!536451532736!e!!g!!aws%20ec2%20instance%20types&ef_id=Cj0KCQiAg_KbBhDLARIsANx7wAwMGv0grfOqMTLnqxh4RpgIcntayYhvgsJqWnTlxUA4-vMxsFbOZwUaAhnyEALw_wcB:G:s&s_kwcid=AL!4422!3!536451532736!e!!g!!aws%20ec2%20instance%20types), which has a standard vcpu of 2, but with burstable cpu up to 24.

I thought the general rule scala had was to create a thread pool per cpu, but in the lambda we bumped this above what vcpu we thought we'd have because of the relatively small amount of work we want each thread to do (mostly just making a http request, then receiving/processing the response).

The benefit of the ec2 model is that we can vertically scale, so I think it might be worth:
- keeping the default thread pool size as it is for now (if we start to see out of memory errors during testing we can always reduce the thread pool size)
-  if we see slow processing of messages, increase the default number of instances in our target group

waisingyiu: Thanks Frankie!  I guess it may be a parameter which needs tuning when we test it on CODE

waisingyiu (approved):

DavidLawes (approved):

senderworker/src/main/scala/com/gu/notifications/ec2worker/SenderWorker.scala
waisingyiu: I agree.  I think there are multiple dimensions for tuning: 
- instance size
- no. of instances in ASG
- no. of threads etc

PR #828, opened 2022-11-18
Apply sqs batch size changes to android worker only

DavidLawes:
## What does this change?

We are testing the performance impact when concurrently handling multiple sqs messages inside a single sqs event.

Yesterday we found that performance degraded when handling 20 sqs messages and sending individual notifications (not in groups).

This change scopes the sqs batch size configuration android only, only the android worker will have:
- an sqs batch size of 20, all other workers revert to a batch size of 1
- a max batching window of 1 second
- a timeout of 180 seconds, all other workers revert to the original timeout of 90 seconds
- a visibility timeout of 200 seconds, all other workers revert to the original visibility timeout of 100 seconds

## How to test

 After deploy, confirm the lambda configuration has been updated correctly for the non-android workers.

### DIFF
--- a/notificationworkerlambda/cdk/lib/senderworker.ts
+++ b/notificationworkerlambda/cdk/lib/senderworker.ts
@@ -23,7 +23,8 @@ interface SenderWorkerOpts {
   tooFewInvocationsEnabled: boolean,
   cleanerQueueArn: string,
   platform: string,
-  paramPrefix: string
+  paramPrefix: string,
+  isBatchingSqsMessages: boolean,
 }
 
 class SenderWorker extends cdkcore.Construct  {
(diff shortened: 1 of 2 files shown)


notificationworkerlambda/cdk/lib/senderworker.ts
```diff
@@ -110,13 +111,13 @@ class SenderWorker extends cdkcore.Construct  {
...
-      timeout: cdk.Duration.seconds(180),
+      timeout: props.isBatchingSqsMessages ? cdk.Duration.seconds(180) : cdk.Duration.seconds(90),
       reservedConcurrentExecutions: props.reservedConcurrency
     })
 
     const senderSqsEventSourceMapping = new lambda.EventSourceMapping(this, "SenderSqsEventSourceMapping", {
-      batchSize: 20,
-      maxBatchingWindow: cdk.Duration.seconds(1),
+      batchSize: props.isBatchingSqsMessages ? 20 : 1,
+      maxBatchingWindow: props.isBatchingSqsMessages ? cdk.Duration.seconds(1) : cdk.Duration.seconds(0),
```
DavidLawes, line 120: I _think_ this is needed? I wondered if the default was 0 if it wasn't specified, but if we only gett 1 sqs message per event, maybe the maxBatchingWindow is irrelevant anyway

groakland (approved):
LGTM

waisingyiu (approved):
👍

notificationworkerlambda/cdk/lib/senderworker.ts
waisingyiu: I agree!

PR #831, opened 2022-11-18
Set up GuCDK stack for Registration and deploy stage-specific templates

jacobwinch:
## What does this change?

We would like to use the new constructs for error budget alerting (https://github.com/guardian/cdk/pull/1576) so that we receive warnings if we are at risk of missing SLO(s) for the Registration service.

This PR imports the Registration YAML via GuCDK in order to configure these alerts, as doing so will make it easier[^1] to reference the correct metrics in TypeScript code and also simplifies deploying the error budget alarm configuration via Riff-Raff. Finally, the work here is the [first step in migrating the Registration app to GuCDK](https://github.com/guardian/cdk/blob/main/docs/migration-guide.md#introduce-cdk-to-the-repository-with-cicd)[^2], so we will need to do it at some point anyway! 

cc @guardian/devx-reliability 

## How to test

This shouldn't change any behaviour - I've [deployed to `CODE`](https://riffraff.gutools.co.uk/deployment/view/4837ee12-d059-4ef3-855b-dab753eaf464) and can see that resources are being updated (but, crucially, not deleted/replaced!) by this change:

image

The updates are just GuCDK automatically tagging resources e.g.

image

* Everything should continue working as before, but we'll be able to make use of `GuErrorBudgetAlarmExperimental` more easily in a future PR.
* There is slightly less work to do when we decide to migrate Registration to GuCDK.

We are moving to two separate infrastructure definition files here (one for `CODE` and one for `PROD`) so there is a risk that I've made a typo and wired things up incorrectly (although I will test in `CODE` there are now some `PROD`-specific errors that would be missed via testing, so a careful review would be much appreciated!).

Lastly, operating in this partially-migrated state means that the process for updating the infrastructure for Registration could be slightly more confusing than before (you need to update the YAML and then update the snapshot tests); I've tried to mitigate this with comments to explain the new process.

[^1]: That said, it is not strictly necessary to do this now - we could just deploy a standalone alerting stack and reference the SLI metrics by hardcoding the names as strings. If you'd prefer not to import the current template let me know and I'll take a different approach.
[^2]: It's a slightly amended version of the steps described in the GuCDK migration docs, since this repo already includes the GuCDK library; the end result is the same. In other words, when you decide to fully migrate Registration to GuCDK, you'll be able to start [here](https://github.com/guardian/cdk/blob/main/docs/migration-guide-ec2.md).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -151,7 +151,8 @@ lazy val registration = project
       "org.tpolecat" %% "doobie-h2"        % doobieVersion % Test
     ),
     riffRaffPackageType := (Debian / packageBin).value,
-    riffRaffArtifactResources += (file(s"registration/conf/${name.value}.yaml"), s"${name.value}-cfn/cfn.yaml"),
+    riffRaffArtifactResources += (file(s"cdk/cdk.out/Registration-CODE.template.json"), s"registration-cfn/Registration-CODE.template.json"),
+    riffRaffArtifactResources += (file(s"cdk/cdk.out/Registration-PROD.template.json"), s"registration-cfn/Registration-PROD.template.json"),
     Debian / packageName := name.value,
     version := projectVersion
   )
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -1,11 +1,26 @@
 import 'source-map-support/register';
+import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import { App } from 'aws-cdk-lib';
 import { InstanceClass, InstanceSize, InstanceType } from 'aws-cdk-lib/aws-ec2';
+import { Registration } from '../lib/registration';
 import { RegistrationsDbProxy } from '../lib/registrations-db-proxy';
 import { SenderWorkerStack } from '../lib/sender-stack';
 import { SloMonitoring } from '../lib/slo-monitoring';
 
 const app = new App();
+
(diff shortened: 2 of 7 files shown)


waisingyiu (approved):
Thank you Jacob for doing this for us!  I agree - it is better to import the cloudformation and then define error budget alarms in the same cloudformation.   I think that the budget alarms may make some references to existing resources in registration and it would be much easier to understand and manage if they are in the same cloudformation.

jacobwinch:
> I think that the budget alarms may make some references to existing resources in registration

Yes, that's exactly right. We will be able to follow the [pattern that you used before to access the CloudFormation load balancer](https://github.com/guardian/mobile-apps-api/pull/1832/files#diff-456f6d20a2c9a271b2a8c2b88b8d6274c2f4d3b33dccb9f3c9f7ac71bebb70b4R248-R250) and then we can access the properties that we'll need to identify metrics e.g.

```typescript

const regisitationLoadBalancer = cfnTemplate.getResource(
  'LoadBalancerToPrivateASG',
) as CfnLoadBalancer;

const loadBalancerName = regisitationLoadBalancer.loadBalancerName;

```

jacobwinch:
`PROD` CloudFormation update looks correct (i.e. no replacements):

image

And the Availability SLI also looks fine since we've merged this:

image

So this seems to have worked as expected 🎉

PR #825, opened 2022-11-16
LIVE-4707 study how to monitor EC2 stack rollout

waisingyiu:
## What does this change?

The PR summaries the study on how to monitor the new EC2 stack after it is rolled out.

The document is in `docs/architecture/06-monitor-ec2-stack-rollout.md`

No changes to the application code were made.

PR #832, opened 2022-11-21
LIVE-4729 Separate AWS metric namespaces for sender lambda and EC2 sender

waisingyiu:
## What does this change?

This PR is part of the work to create metrics and alarms to monitor EC2 stack after it is rolled out.

It enables the sender function to create AWS metrics under different namespaces so that we will be able to differentiate between metrics from worker lambda and EC2 sender.

It also adds a _start_ log before processing a SQS message.  This log is needed to create the concurrency level graph as proposed in #825 (the 6th item). 

## How to test

I deployed the worker lambda to CODE and verified that the AWS metrics were created under the correct namespace `Notifications/$stage/workers` and the `start` logs were created.  I also updated the sample Kibana dashboard ( https://logs.gutools.co.uk/s/mobile/app/dashboards#/view/75c0d220-6447-11ed-be34-2d379af36700?_g=(filters:!()) ) with the created logs.


DavidLawes (approved):
Neat!! Will be super useful, thanks so much :)

PR #833, opened 2022-11-21
Reduce SQS batch size for android worker

DavidLawes:
## What does this change?

We deployed a [change](https://github.com/guardian/mobile-n10n/pull/823) that enabled the android worker lambda to concurrently process batches of SQS messages.

Since that deploy we have observed the following:
- increased number of errors received
- new error types being returned to us, including api client timeout errors
- android lambdas failing to process entire batches of tokens, and then having to retry

Given the SQS batching is the most recent change, this PR reduces the SQS batch size to try and reduce the errors that we're receiving.

## How to test

Given the change is deployed, confirm that the android worker lambda has an SQS batch size of 10.

Reduction is errors received when sending breaking news notifications.


waisingyiu (approved):
👍

PR #837, opened 2022-11-22
Send all notifications using multicast messaging

DavidLawes:
## What does this change?

We created an [experiment](https://github.com/guardian/mobile-n10n/pull/747) to test if multicast messaging improved the speed of delivering app notifications.

We then introduced a further [change](https://github.com/guardian/mobile-n10n/pull/823) to allow the lambda to concurrently process multiple SQS messages per invocation.

When sending messages in batches, this appears to be _ok_, however, when sending messages using the old API method (`sendAsync`) we run into a problem:
- lambda picks up multiple messages from the queue
- lambda attempts to process all messages concurrently
- lambda attempts to make (max) 10 * 1000 tokens http requests
- we run out of threads and throw an error

Screenshot 2022-11-22 at 11 17 37

Since the change to concurrently process SQS messages we can see a spike in out of memory issues in the android worker, that do not correlate to when we send breaking news notifications.

Given we've now validated that multicast messaging is working as we expect, it's suggested that we no longer utilise an allowed list of topics to control whether we send via `sendMulticastAsync` or `sendAsync`. This change now means that ALL notifications to android devices will be sent via multicast messaging.

NB: if this current change is successful in production I'll follow-up with another PR to tidy up our SSM configuration to remove parameter `allowedTopicsForBatchSend`

## How to test

Given this change has been deployed, we expect all notifications to send using multicast messaging.

Reduction in out of memory errors.

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -39,21 +39,12 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
   override def deliverChunkedTokens(chunkedTokenStream: Stream[IO, (ChunkedTokens, Long, Instant, Int)]): Stream[IO, Unit] = {
     chunkedTokenStream.map {
       case (chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize) =>
-        val isAllowedToSendBatch = chunkedTokens.notification.topic.forall(topic => config.allowedTopicsForBatchSend.contains(topic.toString))
-
-        if (isAllowedToSendBatch) {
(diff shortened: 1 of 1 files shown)


waisingyiu:
Oh It is interesting  I thought the number of threads was controlled by the size of the thread pool (e.g. #729) and it would just wait until threads are available, but it seems that it was more complicated than that.

Given the great result with multicast API call, it is great to apply it to all notifications!

waisingyiu (approved):

frankie297 (approved):

frankie297:
I agree, I think now we have successfully tested I see no immediate risk to adding all topics to the allowed list

DavidLawes:
> Oh It is interesting I thought the number of threads was controlled by the size of the thread pool (e.g. #729) and it would just wait until threads are available, but it seems that it was more complicated than that.
> 
> Given the great result with multicast API call, it is great to apply it to all notifications!

Yes, I was under the same impression as you - it wasn't until I saw the error message explicitly ("unable to create native thread: possibly out of memory") that I questioned my understanding.

PR #846, opened 2022-11-23
LIVE-4729 create alarms for EC2 stack rollout

waisingyiu:
## What does this change?

We are building a EC2 implementation of sender worker, which is one of our experiments to further improve the performance of the notification delivery pipeline.

This PR creates alarms to help us monitor the health of the EC2 stack after it is released to PROD.  For each of the platforms (`android, ios`, `android-beta`, `android-edition`, `ios-edition`), it creates four alarms according to the [proposal](https://github.com/guardian/mobile-n10n/blob/main/docs/architecture/06-monitor-ec2-stack-rollout.md) in #825  

## Threshold

I examined similar metrics on the existing worker lambdas to figure out what reasonable values we should use for the thresholds initially:

### Age of the oldest message
Screenshot 2022-11-24 at 10 39 46

The largest value for this metric was 650.  I set the threshold of the alarm to 700

### Number of messages in flight
Screenshot 2022-11-24 at 10 37 50

The largest value for this metric was 350.  I set the threshold to 400.

### Processed messages percentage
It is worth noting that there is a time lapse between the time a message is received and the time a message has been processed, and that is the reason why the percentage sometimes drops to zero and then goes up well above 100%.

Despite this, we could still make use of this metric to monitor whether the EC2 _is processing_ SQS messages at all.  To make it useful, I set the evaluation period to 3 to avoid false alarms as described above, and set the threshold to 50%.

### Failure rate
I notice high failure rate was high on Android occasionally, but it happened probably because we were running an experiment on Android platform  For ios, it is consistently below 1%.  So the failure rate threshold is set to 1%.

Screenshot 2022-11-24 at 12 16 00

## How to test

I deployed it to `CODE` and verified the created cloudwatch alarms and the metrics behind them.

An alarm is raised when the EC2 stack stops working.

### DIFF
--- a/cdk/bin/cdk.ts
+++ b/cdk/bin/cdk.ts
@@ -53,6 +53,8 @@ export const senderCodeProps = {
 	targetCpuUtilization: 20,
 	notificationSnsTopic:
 		'arn:aws:sns:eu-west-1:201359054765:AutoscalingNotificationsCODE',
+	alarmSnsTopic: 'mobile-server-side',
+	alarmEnabled: false,
 	cleanerQueueArn:
 		'arn:aws:sqs:eu-west-1:201359054765:mobile-notifications-registration-cleaning-worker-CODE-Sqs-1CFISZQCN49SR',
 };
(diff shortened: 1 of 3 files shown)


groakland (approved):
LGTM

PR #849, opened 2022-11-28
Increase sqs batch size for android worker

DavidLawes:
## What does this change?

This change increases the sqs batch size for the android worker from 10 to 20. The intention is to test how much parallelism we can achieve within each function invocation.

Looking at the current duration, 99% of invocations complete within ~20s, so even if the extra sqs messages can't be processed concurrently a doubling in duration wouldn't cause a timeout (currently set to 180s).

Screenshot 2022-11-28 at 10 04 29

I reduced the visibility timeout too, we probably only need a 10s buffer between the longest possible invocation (180s) and when the message could be picked up again by another invocation.

## How to test

When this change is deployed we expect to see:
- the lambda configuration updated to show a batch size of 20 for the sqs trigger
- invocations picking up more than 10 messages from the queue at a time (where possible)


waisingyiu (approved):
👍

PR #854, opened 2022-11-30
Results from experimenting with batching

DavidLawes:
## What does this change?

No code changes - just updating docs to describe the results and conclusions from experimenting with batching inside the android worker lambdas.

The results on batching are probably best viewed when the page is [rendered](https://github.com/guardian/mobile-n10n/blob/3fe458a092807fd4a298b18ba16d0e08e6c707f5/docs/testing/16-batching-sqs-messages-and-api-calls.md)

The number of the docs had got out of sync, so I updated that too.


waisingyiu (approved):
Thank you for the great write-up!

PR #855, opened 2022-12-01
Add first draft of error budget alerts for subscribe/unsubscribe SLO

jacobwinch:
cc @waisingyiu, @DavidLawes, @frankie297, @jorgeazevedo and @michaelwmcnamara (who all worked on this PR).

## What does this change?

This PR utilises a new GuCDK construct (https://github.com/guardian/cdk/pull/1576) to implement error budget alerting. 

The alarm configuration that we're adding here follows Google's recommendations from the [SRE Workbook](https://sre.google/workbook/alerting-on-slos/) to implement `Multiwindow, Multi-Burn-Rate Alerts`. There is also another nice explanation of this alerting strategy [here](https://developers.soundcloud.com/blog/alerting-on-slos).

Note that the SLO target (which has been set to a placeholder value `0.999` i.e. 99.9%) and the SNS topic for alerts (currently `jacob-test`) will need to be updated when we have agreed on the real target. 

I think it's worth merging this initial PR now as:

1. It will allow us to evaluate the alert performance against real production traffic (e.g. to confirm that there are no false alarms)
1. It prevents this branch from becoming stale

## How to test

We've deployed this alarm configuration to `CODE` and confirmed that simulating an outage trips the appropriate alarm. More extensive testing was performed whilst building the GuCDK construct and I don't think we need to repeat that testing here.

We have all of the alert configuration in place for the subscribe/unsubscribe SLO. It will be trivial to update the target, SNS topic and snapshot tests when we are ready to start monitoring the real SLO target.

The SNS topic that we are using for testing doesn't email anyone, so we can safely merge this without adding noise to inboxes/Chat channels.


DavidLawes (approved):
👍 

Agree that we can manually verify whether the alarm has fired to determine if we've set a reasonable SLO target

PR #856, opened 2022-12-01
Ensure that alarm names are unique across different stages

jacobwinch:
## What does this change?

Fixes a problem with https://github.com/guardian/mobile-n10n/pull/855 which caused `PROD` deployments to fail.

Alarm names must be unique, so the fact that we have already deployed these alarms in `CODE` means that we cannot deploy them to `PROD` 🤦‍♂️ 

The SLO name that we pass into the construct is used to generate the alarm names in CloudWatch, so we can fix this problem by adding the stage to the SLO name. The GuCDK construct should probably do this automatically (and I might add that in the future), but that requires a library release so this PR is a faster fix!

## How to test

I will deploy to `CODE` and then `PROD` and hope that it works this time!

We should be able to deploy to `PROD` again!

This should be pretty low risk; it will recreate all alarms but that should be perfectly safe.

### DIFF
--- a/cdk/lib/__snapshots__/registration.test.ts.snap
+++ b/cdk/lib/__snapshots__/registration.test.ts.snap
@@ -116,9 +116,9 @@ Object {
     },
   },
   "Resources": Object {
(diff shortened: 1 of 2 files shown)


DavidLawes (approved):
👍

PR #859, opened 2022-12-01
LIVE 4882: add metrics listener for pushy client

DavidLawes:
## What does this change?

Before we start tuning the apns pushy client we thought it would be a good idea to capture some additional performance data. The pushy library allows us to define a [metrics listener](https://github.com/jchambers/pushy/tree/master/dropwizard-metrics-listener) when creating a new client. The metrics listener defines some methods which will be invoked when certain events occur.

This change defines does a couple of things:
- when we create a new apns pushy client we define a metrics listener
- the metrics listener we define is an instance of a new class we've added: IOSMetricsRegistry
- instantiating the IOSMetricsRegistry class creates a new metrics registry: I understand this to be something like a mutable map of the metrics we capture
- instantiating the IOSMetricsRegistry class creates 7 new metrics, all of which are Counters that are incremented when the corresponding default methods are invoked by the pushy client
- when the lambda invocation is complete we want to log the metric data. We want to log the data for IOS only (we don't yet use a metrics registry for android) and so I've defined a new logging method to allow this abstraction for iOS only

NB: in this PR we're not yet capturing this data as aws metrics, but maybe that could be a follow up pr if we think the data is worth retaining in the longer-term and that it captures everything we need.

## How to test

I deployed to CODE and I can see the new fields included in the log event:

Screenshot 2022-12-01 at 15 21 13


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/apns/models/IOSMetricsRegistry.scala
```diff
@@ -0,0 +1,31 @@
...
+  val notificationRejected: Counter = registry.counter("pushy.notificationRejected")
+  val connectionAdded: Counter = registry.counter("pushy.connectionAdded")
+  val connectionRemoved: Counter = registry.counter("pushy.connectionRemoved")
+  val connectionCreationFailed: Counter = registry.counter("pushy.connectionCreationFailed")
+
+  override def handleWriteFailure(apnsClient: ApnsClient, notificationId: Long): Unit = writeFailures.inc()
+
+  override def handleNotificationSent(apnsClient: ApnsClient, notificationId: Long): Unit = notificationSent.inc()
+
+  override def handleNotificationAccepted(apnsClient: ApnsClient, notificationId: Long): Unit = notificationAccepted.inc()
```
waisingyiu, line 21: Do you think it would be useful to write a log when notification is accepted / sent so that we can get something like the rate of notification accepted / sent over a period of time?  Maybe we can then understand whether the Pushy is able to send notification as fast as we call Pushy API?
(Here I assume the `notificationAccepted` happens when we call Pushy API, and `notificationSent` happens when Pushy has sent the notifications to APNS.)

Anyway this PR looks great to me!

waisingyiu (approved):

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/apns/models/IOSMetricsRegistry.scala
waisingyiu: That sounds great!

DavidLawes: With the extra log messages I can see the request latency in the log events:
Screenshot 2022-12-02 at 15 18 49

PR #862, opened 2022-12-02
LIVE-4823: Define new configurable properties for iOS sender worker

DavidLawes:
## What does this change?

In this PR we make `maxConcurrency` and the number of concurrent connections configurable properties. This will enable testing and tuning of the ios sender worker as part of future experiments.

SSM variables will be created with the current values:
- maxConcurrency: 100
- number of concurrent connections: this wasn't previously specified, but the [default](https://github.com/jchambers/pushy/blob/master/pushy/src/main/java/com/eatthepath/pushy/apns/ApnsClientBuilder.java#L75) value is 1

## How to test

Given this change is deployed, we should expect to see no changes. The current values of the new configurable properties have been set to what we already use:
- maxConcurrency: 100
- number of concurrent connections: 1

We should expect both the sender lambdas and the ec2 workers to start up correctly, having retrieved the new properties successfully from SSM.


frankie297 (approved):
👍

PR #719, opened 2022-08-17
[90in2 SLO] Proof of concept for near real-time tracking

jacobwinch:
This PR is a proof of concept for tracking the 90in2 SLO in near real-time. If implemented fully, it would allow us to evaluate performance for each notification within minutes rather than days. 

_Note: this work is not complete and cannot be shipped in its current state (hence the draft PR); hopefully it demonstrates the potential architecture well enough to make a decision about whether to proceed with this work or not._

## Architecture

* Each time a breaking news notification is received by the `notification` app it sends a tracking message to a new SQS queue.
* There is a 10 minute delay (although this is configurable)[^1]. During this 10 minute window devices are receiving notifications and sending tracking information which is made available via Fastly/S3/Athena.
* After 10 minutes, the message is picked up by a new Lambda. 
* This Lambda queries Athena[^2] and calculates the % notifications delivered in 120 seconds.

image

## Next Steps

If we decide to pursue this type of tracking further, I would suggest the following next steps:

1. Figure out how to get the expected number of deliveries for each notification to this Lambda
2. Test in `PROD` to see whether percentages calculated align with current manual tracking process
3. Send structured logs / CloudWatch metrics so that we can visualise performance

[^1]: We could reduce this further to get faster feedback, but we need to wait long enough to be sure that tracking information can be queried via Athena.
[^2]: There was already code for querying Athena in this repo, I've just restructured things so that it can be shared across Lambdas. It's easiest to see that in https://github.com/guardian/mobile-n10n/pull/719/commits/116a098beca131396546503d4010fce536fd0d56.

PR #866, opened 2022-12-07
Collect metrics for notification delivery SLIs

jacobwinch:
## What does this change?

This PR updates the SLO monitoring code (first introduced via https://github.com/guardian/mobile-n10n/pull/719) to capture CloudWatch metrics related to notification delivery times.

The metrics captured here allow us to [calculate various SLIs](https://metrics.gutools.co.uk/goto/lw3X88FVk) related to notification delivery performance:

image
_N.B. the SLI values here are just examples based on test data_

We can also get platform specific SLIs (this seems useful given the potential for performance differences between Firebase and APNS):

image
_N.B. the SLI values here are just examples based on test data_

image
_N.B. the SLI values here are just examples based on test data_

## How to test

I've tested this by deploying to `CODE` and adding a [small hack](https://github.com/guardian/mobile-n10n/commit/c6547742484ce00e63989e72af650b991b13c281) which allows me to query production data. 

This will allow us to visualise some useful SLIs. In the future, we could also consider using these metrics as the basis for alerts.

* The current SLI calculations (in Grafana, not in this codebase) use the total number of deliveries metric rather than a (currently absent) _expected_ number of deliveries metric. We might want to amend this later as using the latter would allow us to pick up cases where we had a coverage problem as well as a performance problem (currently if we only deliver 1 notification but we deliver it very quickly, the SLI value will be 100%!).
* @DavidLawes mentioned that the MSS team is using the structured logging produced by this Lambda; I have left the original logging in place (although the calculation is slightly different) to avoid breaking anything.


slomonitor/src/main/scala/com.gu.notifications.slos/SloMonitor.scala
```diff
@@ -47,41 +92,76 @@ object SloMonitor {
...
+       |  END) AS less_than_240,
+       |  SUM(CASE
+       |    WHEN DATE_DIFF('second', from_iso8601_timestamp('$sentTime'), received_timestamp) < 270 THEN 1
+       |    ELSE 0
+       |  END) AS less_than_270,
+       |  SUM(CASE
+       |    WHEN DATE_DIFF('second', from_iso8601_timestamp('$sentTime'), received_timestamp) < 300 THEN 1
+       |    ELSE 0
+       |  END) AS less_than_300,
+       |  COUNT(*) AS total_deliveries
```
jacobwinch, line 137: N.B. that `total_deliveries` really means `total_deliveries_when_this_query_was_run` - in practice that means the number of deliveries [in the first 10 minutes](https://github.com/guardian/mobile-n10n/blob/63dd9b0ca1b3f1020cc65c372c0c72fd1cb31403/notification/app/notification/services/SloTrackingSender.scala#L18), at present.

slomonitor/src/main/scala/com.gu.notifications.slos/SloMonitor.scala
```diff
@@ -47,41 +92,76 @@ object SloMonitor {
...
-          logger.info(Map(
-            "notificationId" -> notificationId,
-            "deliveriesWithin2mins" -> actualDeliveries
-          ),
-          s"Notifications delivered within 120 seconds was $actualDeliveries")
-        })
-      )
+    val result = Athena.startQuery(Query("notifications", query, s"s3://aws-mobile-event-logs-${this.stage.toLowerCase()}/athena/slo-monitoring")).flatMap {
+      Athena.fetchQueryResponse(_, rows => {
+        val androidDeliveries = rows.head
```
jacobwinch, line 154: The response data looks like this:

```
"platform","less_than_30","less_than_60","less_than_90","less_than_120","less_than_150","less_than_180","less_than_210","less_than_240","less_than_270","less_than_300","total_deliveries"
"android","424","432","436","437","439","441","441","441","441","441","441"
"ios","701","702","703","704","704","704","704","704","704","704","704"
```

It looks like the `GROUP BY` is sorting alphabetically by default but perhaps we should add some validation or something here to ensure that the query results look roughly how we expect (e.g. there should be X rows, Y columns, one row starting with `android` and one with `ios` etc.)

slomonitor/src/main/scala/com.gu.notifications.slos/SloMonitor.scala
DavidLawes: I think this is a good enough for now. I know that we've been keen to explore the number of missing/lost deliveries too, so at some point in the future we may want to update how we define the "total".

slomonitor/src/main/scala/com.gu.notifications.slos/SloMonitor.scala
DavidLawes: Great idea to separate this by platform! I guess we could use metric math in the console to get an aggregated view of the 90in2 (if we ever wanted to do this)?

DavidLawes (approved):
Nice! This should help us get better insights more easily, thanks so much :)

As a side note, in our notification worker lambdas we've temporarily reverted to sending all metrics via the put metric api (instead of using embedded metric format) - this was to enable us to test the stack when running on ec2s instead of lambdas. Possibly when/if we revert that change we might experience the same logging issue you've come across here, so thanks for flagging it.

jacobwinch:
> Great idea to separate this by platform!

It was inspired by the excellent presentation yesterday afternoon!

>  I guess we could use metric math in the console to get an aggregated view of the 90in2 (if we ever wanted to do this)?

Yes exactly, here's an example of that calculation in Grafana:

image

`((iosLessThan120 + androidLessThan120) / (iosTotalDeliveries + androidTotalDeliveries)) * 100`

jacobwinch:
These metrics are coming through now; we can see the results so far in this [dashboard](https://metrics.gutools.co.uk/goto/NVfvdXK4k).

PR #867, opened 2022-12-07
Fix logging for 90in2 deliveries

jacobwinch:
## What does this change?

Fixes a bug introduced by https://github.com/guardian/mobile-n10n/pull/866, which is caused by using `+` with `String`s instead of `Double`s 🤦‍♂️:

image

`Notifications delivered within 120 seconds was 374718449495` 🤯 

## How to test

I have deployed to `CODE` and checked the logging: 

image

image

The logging should be truthful again after this PR is merged.

I think this is pretty low risk.


DavidLawes (approved):
Whoops, sure I've made this exact same error, sorry for not spotting it!

PR #868, opened 2022-12-08
LIVE-4830 invoke sender lambda by eventbridge

waisingyiu:
## What does this change?

AWS support gave the following advice to achieve high concurrency in our sender lambda: 

> We believe the delays you are currently experiencing relate to the fixed way [Lambda scales up against SQS messages](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#events-sqs-scaling). An alternative approach we discussed is to make use of [Amazon EventBridge](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-what-is.html) to directly trigger the Lambda functions rather than using SQS as the event source.

> With this method, the harvester would still place the tokens onto the SQS queue but then trigger an EventBridge event that would invoke the Lambda. There would be no need to ensure correlation between the number of SQS events and EventBridge events as you are pulling batched messages. If the SQS queue is empty when the Lambda is triggered then it would quickly exit.

This PR implements this suggestion from AWS.

The following specific changes are made:
1. In harvester function, we emit an event to EventBridge for every SQS message we writes
2. In sender lambda, we add an event rule which is triggered by the events in (1) and then invokes the sender lambda
3. Remove the SQS event source mapping
4. Define another entry point in sender lambda.   We read messages from SQS queues and then reuse the existing logics for processing.
5. Add a configuration `batchsize` which set the maximum number of message to receive when we read from SQS
6. Set a schedule to emit events in case some messages are left unprocessed in SQS queues (chances are very small, but it is possible because [duplicated messages may be received](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues.html) 

I set the following values for `batchsize` on CODE
- Android: 10 (we use batch size of 20 in the current event source mapping, but the `receiveMessage` API allows a maximum value of 10 only)
- others: 1 (to align with the current event source mapping, but we may consider changing it to 2 to make sure all messages on the queue are received and processed)

We notice that after this change, our sender lambda will hit the limit on the rate of Parameter Store API call.  There are two options:
1. Turn on higher parameter store throughput, which is a paid tier
2. Enable Snapstart (require further work and testing)

## How to test

On CODE, we verified that sender lambda functions were invoked at a much higher rate than it would be if we use SQS event source mapping.  All the SQS messages could be processed, and the logs indicated that they were processed correctly in dryRun mode.

It would be a success if it can -
1. lead to higher token processing rate
4. result in higher 90in2 metric

It may happen that the firebase and APNS may not be able to process the tokens at the rate we send the HTTP requests as we start up many lambda functions within a short period and all of them sends HTTP requests to firebase / APNS.  It may result in our API calls being throttled.   The lambda function has retry mechanism in place so it would mostly increase the duration if it happened.


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
```diff
@@ -17,16 +18,37 @@ import org.slf4j.{Logger, LoggerFactory}
...
+import com.amazonaws.services.sqs.AmazonSQS
+import scala.annotation.tailrec
+
+
+case class SqsMessage(body: String, sentTimestamp: Long)
+
+object SqsMessage {
+  def fromSqsApiMessage(message: Message) = SqsMessage(message.getBody(), message.getAttributes.getOrDefault("SentTimestamp", "0").toLong)
+
+  def fromSqsEventMessage(message: SQSMessage) = SqsMessage(message.getBody(), message.getAttributes.getOrDefault("SentTimestamp", "0").toLong)
```
waisingyiu, line 40: I want to run the same piece of code on a SQS message no matter whether we use EventBridge or SQS event source, but the classes representing a SQS message are different in `SQSEvent` and in `receiveMesage` API.  So I create a case class to capture relevant fields from SQS messages.

notificationworkerlambda/cdk/lib/senderworker.ts
```diff
@@ -115,15 +117,37 @@ class SenderWorker extends cdkcore.Construct  {
...
+      },
+    });
+    eventRule.addTarget(new LambdaFunction(senderLambdaCtr, {
+      maxEventAge: cdk.Duration.minutes(15), // Optional: set the maxEventAge retry policy
+      retryAttempts: 5, // Optional: set the max number of retry attempts
+    }));
+
+    const scheduledEventRule = new Rule(this, "ScheduledRerunRule", {
+      schedule: Schedule.cron({ minute: "0/10" }),
+    });
```
waisingyiu, line 139: Ideally the EventBridge events trigger our sender lambda which are able to process all SQS messages, but maybe in rare, unexpected cases, a few unprocessed SQS messages are left on the SQS.  I create a scheduled rule to invoke the sender lambda to deal with them if any.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
```diff
@@ -105,4 +130,25 @@ trait SenderRequestHandler[C <: DeliveryClient] extends Logging {
...
+
+    val receiveMessages = sqsClient.receiveMessage(new ReceiveMessageRequest()
+      .withQueueUrl(sqsQueue)
+      .withMaxNumberOfMessages(batchSizeToUse)
+      .withWaitTimeSeconds(1)
+      .withAttributeNames("SentTimestamp"))
+      .getMessages.asScala.toList
+
+    handleChunkTokensInMessages(receiveMessages.map(SqsMessage.fromSqsApiMessage(_)))
+    receiveMessages.foreach(msg => sqsClient.deleteMessage(sqsQueue, msg.getReceiptHandle()))
```
DavidLawes, line 152: Am I understanding it correctly: given we've successfully handled the chunked tokens, we delete the messages from the queue? And this functionality is no different from what we currently do: if any message in the batch throws an unhandled error, all messages from the batch will be retried?

notificationworkerlambda/cdk/lib/senderworker.ts
```diff
@@ -115,15 +117,37 @@ class SenderWorker extends cdkcore.Construct  {
...
+    scheduledEventRule.addTarget(new LambdaFunction(senderLambdaCtr, {
+      event: RuleTargetInput.fromObject({
+        source: "com.gu.notifications.worker",
+        "detail-type": "Scheduled sender",
+        resources: [`${this.senderSqs.queueUrl}`],
+        detail: {
+          platform: `${props.platform}`,
+          batchsize: 10,
```
DavidLawes, line 147: Is this the same for both ios and android? When we were testing previously the ios lambda didn't process a "big" batch size very well. Given the lambda code (and the underlying concurrency model?) is unchanged as a result of this PR I suppose we could see a similar degradation in  behaviour when processing lots of SQS messages per invocation. It's worth testing this though because the result may be different :)

DavidLawes: Oh, sorry, maybe this is a different batch size? (I was thinking of the sqs message batch size that the lambda uses to get messages from the queue)

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
waisingyiu: Yes, that's right.  

When we finish processing the chunked tokens, we will delete the messages at the end.  

If any exceptions are thrown and unhandled, the lambda will terminate, leaving the messages in the queue.  In this case, the lambda returns an error and the EventBridge will [retry calling the function again](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rule-dlq.html) (the maximum retry count is set to 5 in our CDK).

I think if unhandled exceptions happen, they probably results from JVM runtime error or getting AWS client instance.  It looks less likely to come from the processing of individual messages because we handle exceptions that may happen from APNS / Firebase API and put them in the output (`Either[DeliveryException, DeliverySuccess]`) of the stream.

notificationworkerlambda/cdk/lib/senderworker.ts
waisingyiu: It is a different batch size.  It is used in the scheduled execution of the sender lambda to process SQS messages which are left on the queue for some reasons. 

For invocation that are triggered by the EventBridge event (sent by Harvester), the lambda gets the batch size from a new SSM parameter `batchsize`.  It is set to 10 for Android (the max batch size allowed in `receiveMessage` API) and set to 1 for the rest (to be in line with our current settings).

PR #872, opened 2022-12-09
LIVE-4850 upgrade to Java 11

waisingyiu:
## What does this change?

We are exploring the new AWS `Snapstart` feature to reduce the cold start initialisation time of the sender lambda in our effort to increase the overall notification delivery performance.  The `Snapstart` feature requires Java11 runtime so we have to upgrade the lambda function from Java 8 to Java11.

Other important reasons are:
1.  it was reported that Graviton servers achieve better performance on Java 11 
2. more generally, newer Java versions support newer language and platform features
3. it is suggested that future Scala 3 versions may drop support for Java 8
4. the Play Framework, which is used in `registration`, `notification` and `report`, announced that the next major version (2.9) will drop support for Java 8
 
This PR upgrades the whole project to use Java11.

Major changes are:
1. Set the lambda runtime to Java11 (for those which do not use docker)
2. Change AMI recipe to `bionic-mobile-java11-ARM` for EC2 app.  This recipe is basically identical to the old recipe `bionic-mobile-ARM` except that Java 8 is replaced with Java 11.
3. Add `.tool-version` so that the `asdf` (if installed) attempts to use JDK11 when we compile and test the project locally

## How to test
All test cases passed.  All modules were deployed on CODE successfully.  I tried registering tokens and sending push notifications on CODE, and it worked as expected.

All services run properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,4 +29,3 @@ dynamodb-local-schedule-lambda/
 project/.bloop/
 project/metals.sbt
 .bsp/
-.tool-versions
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,1 @@
+java corretto-11.0.17.8.1
--- a/eventconsumer/cfn.yaml
+++ b/eventconsumer/cfn.yaml
@@ -128,7 +128,7 @@ Resources:
       Handler: com.gu.notifications.events.AthenaLambda::handleRequest
       MemorySize: 1024
       Role: !GetAtt ExecutionRole.Arn
-      Runtime: java8
+      Runtime: java11
       Timeout: 120
       Tags:
         - Key: Stage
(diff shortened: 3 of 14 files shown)


DavidLawes (approved):
Nice! For the lambdas, did we also need to update the java version used by team city, or was that unnecessary?

waisingyiu:
> Nice! For the lambdas, did we also need to update the java version used by team city, or was that unnecessary?

Yes, I think it is better to upgrade the Java version for compilation and for runtime.  I have created the Teamcity build configuration [mobile-notifications-all-java11](https://teamcity.gutools.co.uk/viewType.html?buildTypeId=Mobile_MobileNotifications_MobileNotificationsAllJava11) and used it on this branch.  (to avoid affecting the main branch which use Java8)

I will change the `mobile-notifications-all` build configuration to Java11 when I am about to merge the PR.

PR #881, opened 2022-12-16
Update error budget alerts for subscribe/unsubscribe SLO

jacobwinch:
cc @DavidLawes & @michaelwmcnamara (who also worked on this PR)

## What does this change?

This PR updates the error budget alarms (first created via https://github.com/guardian/mobile-n10n/pull/855) for subscribe/unsubscribe SLO to use the real SLO target (99.5%). It also updates the alarms so that notifications are sent to the `ServerAlerts` Chat channel (via the `mobile-server-side` topic and [SNS to GChat Lambda](https://github.com/guardian/google-chat-bots/tree/main/sns-to-gchat)).

## How to test

I've deployed this to `CODE` to confirm that the infrastructure changes can be deployed successfully.

We have alarms in place for the real SLO target, so we should be notified if we're at risk of missing the SLO.

We have checked the recent performance of the alerts to confirm that they wouldn't have fired unnecessarily when using this target (the slow and medium burn alarms _were_ firing sometimes for a 99.9% SLO target, which this service cannot currently achieve).


cdk/lib/__snapshots__/registration.test.ts.snap
```diff
@@ -193,7 +193,7 @@ Object {
             "ReturnData": false,
           },
         ],
-        "Threshold": 0.014400000000000013,
+        "Threshold": 0.07200000000000006,
```
jacobwinch, line 196: These thresholds are [automatically calculated](https://github.com/guardian/cdk/blob/9d5ff3552e1b95391415c4c0db0a0aac90d34f90/src/experimental/constructs/error-budget-alarm.ts#L173) by multiplying the error budget by various burn rates ([slow](https://github.com/guardian/cdk/blob/9d5ff3552e1b95391415c4c0db0a0aac90d34f90/src/experimental/constructs/error-budget-alarm.ts#L153)/[medium](https://github.com/guardian/cdk/blob/9d5ff3552e1b95391415c4c0db0a0aac90d34f90/src/experimental/constructs/error-budget-alarm.ts#L136)/[fast](https://github.com/guardian/cdk/blob/9d5ff3552e1b95391415c4c0db0a0aac90d34f90/src/experimental/constructs/error-budget-alarm.ts#L120)).

DavidLawes (approved):
I paired on this and it all makes sense to me :)

PR #884, opened 2022-12-19
Collect token delivery latency metrics

jacobwinch:
cc @michaelwmcnamara (who also worked on this PR)

## What does this change?

This PR allows us to calculate (and collect CloudWatch metrics for) token delivery latency.

This means that we can use a standard latency metric in CloudWatch to visualise internal stack performance (i.e. we can check how the stack is performing in isolation of Firebase/APNS delivery times, which may be unpredictable). For example, we can calculate the 99th, 95th & 90th percentiles for token delivery latency. 

image

We can also calculate an SLI based on this metric, for example:

Good events = _the number of token deliveries where the delivery took less than X seconds_
Valid events = _the number of token deliveries_
SLI = _(the number of token deliveries where the delivery took less than X seconds / the number of token deliveries) * 100_ 

IIUC a single token delivery directly corresponds to an individual user receiving a notification[^1]. Consequently, using token delivery latency is likely to be useful for an SLI (and potentially an SLO) because:

1. It is easy to understand/reason about the correlation between latency increasing and user happiness decreasing[^2]
2. It allows us to easily compare a (new) token delivery SLI with the (existing) 90in2 SLI, which also operates at the level of individual-user deliveries

## How to test

* I have deployed this to `CODE` and confirmed that the latency metrics being calculated look sensible
* Since this PR changes the JSON that is passed between the various microservices, I have also confirmed that:
  * A (dry-run) notification can be sent if `notification` is running an old build and the `workers` are running this new build
  * A (dry-run) notification can be sent if `notification` is running this new build and the `workers` are still running an old build
* I sent a single 'standard' notification (i.e. not a dry-run) to confirm that metrics were actually being pushed to CloudWatch. There were not many datapoints for this; I assume that we must have thousands of 'fake' tokens in `CODE`?
* We have added unit tests for the latency metric batching logic, as this was a little tricky to write

* We can better visualise performance of the stack via CloudWatch (and Grafana)
* We should also be able to create alerts related to this metric/SLI (if desired)

I can think of a few potential risks here:

1. We are adding a new field to the JSON that is passed between the different microservices (see https://github.com/guardian/mobile-n10n/commit/a8d26d171c249e229a2240bf472d9d183866d3c2). I have tried to mitigate this risk making the field optional (for consumers) in the first instance. This means that any messages that are already in the queue(s) should still be processed correctly, even though they lack this new field. I have also tested (see above) to confirm that the service can continue operating successfully even if this build is rolled out to different microservices at different times. Once this PR has been deployed, we'll be able to make the new field mandatory in all of the models and this complexity will be removed.

1. We are making more API calls to CloudWatch as a result of this PR. This has the potential to:
   1. Slow the worker function(s) down 
   1. ~Introduce invocation failures (e.g. due to a CloudWatch API call failing).~ Update: I don't think a CloudWatch call failing will cause the invocation to fail, see [this comment](https://github.com/guardian/mobile-n10n/pull/884#discussion_r1052310294).
   1. Increase AWS costs[^3]
  
In order to mitigate the above, we have implemented batching logic so that latency metrics are pushed to CloudWatch in batches of _150 unique values_. 

I think each worker is currently [processing 1000 tokens](https://github.com/guardian/mobile-n10n/blob/471357ac709baada6dc9cfb6830a4bdc88c3e4b3/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala#L67)? So in theory a worker invocation could result in 7 additional API calls.

However, as a unique value in this context is the number of seconds spent processing a notification and Lambda invocations are well below 150 seconds, I would expect each Lambda to make one additional API call in practice[^4].

Note that there are options for optimising further here (e.g. we could move to [embedded metric format](https://aws.amazon.com/about-aws/whats-new/2019/11/amazon-cloudwatch-launches-embedded-metric-format/), or combine [some of the other CloudWatch monitoring functionality](https://github.com/guardian/mobile-n10n/blob/471357ac709baada6dc9cfb6830a4bdc88c3e4b3/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/Cloudwatch.scala#L45-L77) into the same API calls).

[^1]: Assuming that the Firebase/APNS delivery succeeds eventually.
[^2]: I think it will be easier to explain this SLI to stakeholders than something more technical (e.g. throughput).
[^3]: Since we [pay per API call](https://aws.amazon.com/cloudwatch/pricing/).
[^4]: For example, when testing in `CODE` I observed that the Android Worker makes a single API call where there is only unique value - the number of seconds spent processing  (e.g. 45 seconds) - and the count is the number of tokens processed in that time (e.g. 1000).

### DIFF
--- a/common/src/main/scala/models/ShardedNotification.scala
+++ b/common/src/main/scala/models/ShardedNotification.scala
@@ -2,6 +2,8 @@ package models
 
 import play.api.libs.json.{Format, Json}
 
+import java.time.Instant
+
 case class ShardRange(start: Short, end: Short) {
   def range: Range = Range.inclusive(start, end)
 }
@@ -13,6 +15,7 @@ object ShardRange {
 case class ShardedNotification(
   notification: Notification,
   range: ShardRange,
+  notificationAppReceivedTime: Option[Instant]
 )
 
 object ShardedNotification {
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -5,7 +5,7 @@ import com.gu.notifications.worker.cleaning.CleaningClientImpl
 import com.gu.notifications.worker.delivery.DeliveryException.InvalidToken
 import com.gu.notifications.worker.delivery.{BatchDeliverySuccess, DeliveryClient, DeliveryException}
 import com.gu.notifications.worker.delivery.fcm.{Fcm, FcmClient}
-import com.gu.notifications.worker.models.SendingResults
+import com.gu.notifications.worker.models.{LatencyMetrics, SendingResults}
 import com.gu.notifications.worker.tokens.{BatchNotification, ChunkedTokens}
 import com.gu.notifications.worker.utils.{Cloudwatch, CloudwatchImpl, Reporting}
 import fs2.{Pipe, Stream}
@@ -45,6 +45,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
         deliverBatchNotificationStream(Stream.emits(chunkedTokens.toBatchNotificationToSends).covary[IO])
           .broadcastTo(
             reportBatchSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
+            reportBatchLatency(chunkedTokens, functionStartTime), // FIXME - we want the time that the MSS stack received the notification, not the function start time
             cleanupBatchFailures(chunkedTokens.notification.id),
             trackBatchProgress(chunkedTokens.notification.id))
     }.parJoin(maxConcurrency)
@@ -66,7 +67,17 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
     input
       .fold(SendingResults.empty) { case (acc, resp) => SendingResults.aggregateBatch(acc, chunkedTokens.tokens.size, resp) }
       .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform, sqsMessageBatchSize), prefix = s"Results $notificationLog: ").andThen(_.map(cloudwatch.sendPerformanceMetrics(env.stage, enableAwsMetric))))
-      .through(cloudwatch.sendMetrics(env.stage, Configuration.platform))
+      .through(cloudwatch.sendResults(env.stage, Configuration.platform))
+  }
+
+  def reportBatchLatency[C <: DeliveryClient](chunkedTokens: ChunkedTokens, notificationSentTime: Instant): Pipe[IO, Either[DeliveryException, BatchDeliverySuccess], Unit] = { input =>
+    val shouldPushMetricsToAws = chunkedTokens.notification.dryRun match {
+      case Some(true) => false
+      case _ => true
+    }
+    input
+      .fold(LatencyMetrics.empty) { case (acc, resp) => LatencyMetrics.aggregateBatchLatency(acc, resp, notificationSentTime) }
+      .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform))
   }
 
   def cleanupBatchFailures[C <: DeliveryClient](notificationId: UUID): Pipe[IO, Either[Throwable, BatchDeliverySuccess], Unit] = { input =>
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -76,7 +76,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
       case _ => true
     }
     input
-      .fold(LatencyMetrics.empty) { case (acc, resp) => LatencyMetrics.aggregateBatchLatency(acc, resp, notificationSentTime) }
+      .fold(List.empty[Long]) { case (acc, resp) => LatencyMetrics.collectBatchLatency(acc, resp, notificationSentTime) }
       .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform))
   }
 
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -45,7 +45,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
         deliverBatchNotificationStream(Stream.emits(chunkedTokens.toBatchNotificationToSends).covary[IO])
           .broadcastTo(
             reportBatchSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
-            reportBatchLatency(chunkedTokens, functionStartTime), // FIXME - we want the time that the MSS stack received the notification, not the function start time
+            reportBatchLatency(chunkedTokens, chunkedTokens.notificationAppReceivedTime.getOrElse(functionStartTime)), // FIXME: remove this fallback after initial deployment
             cleanupBatchFailures(chunkedTokens.notification.id),
             trackBatchProgress(chunkedTokens.notification.id))
     }.parJoin(maxConcurrency)
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -77,7 +77,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
     }
     input
       .fold(List.empty[Long]) { case (acc, resp) => LatencyMetrics.collectBatchLatency(acc, resp, notificationSentTime) }
-      .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform))
+      .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform, Reporting.notificationTypeForObservability(chunkedTokens.notification)))
   }
 
   def cleanupBatchFailures[C <: DeliveryClient](notificationId: UUID): Pipe[IO, Either[Throwable, BatchDeliverySuccess], Unit] = { input =>
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
@@ -52,7 +52,7 @@ trait SenderRequestHandler[C <: DeliveryClient] extends Logging {
     }
     input
       .fold(List.empty[Long]) { case (acc, resp) => LatencyMetrics.collectLatency(acc, resp, notificationSentTime) }
-      .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform))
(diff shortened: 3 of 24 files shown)


common/src/main/scala/models/ShardedNotification.scala
```diff
@@ -13,6 +15,7 @@ object ShardRange {
 case class ShardedNotification(
   notification: Notification,
   range: ShardRange,
+  notificationAppReceivedTime: Option[Instant]
```
jacobwinch, line 18: We are touching a lot of files here in order to wire this timestamp throughout the codebase. It might be easiest to review all of these via this commit: https://github.com/guardian/mobile-n10n/commit/a8d26d171c249e229a2240bf472d9d183866d3c2.

jacobwinch, line 18: This is optional for now so that any messages which are already in the queue (at deployment time) can still be processed; we'll be able to make this mandatory in a follow-up PR.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -47,6 +47,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
         deliverBatchNotificationStream(Stream.emits(chunkedTokens.toBatchNotificationToSends).covary[IO])
           .broadcastTo(
             reportBatchSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
+            reportBatchLatency(chunkedTokens, chunkedTokens.notificationAppReceivedTime.getOrElse(functionStartTime)), // FIXME: remove this fallback after initial deployment
```
jacobwinch, line 50: As mentioned before, this time is optional for now so that any messages which are already in the queue (at deployment time) can still be processed. 

We'll be able to make this mandatory in a follow-up PR, which will allow us to remove this hack.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
```diff
@@ -78,6 +87,7 @@ trait SenderRequestHandler[C <: DeliveryClient] extends Logging {
         deliverIndividualNotificationStream(individualNotifications)
           .broadcastTo(
             reportSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
+            reportLatency(chunkedTokens, chunkedTokens.notificationAppReceivedTime.getOrElse(functionStartTime)), // FIXME: remove this fallback after initial deployment
```
jacobwinch, line 90: As mentioned before, this time is optional for now so that any messages which are already in the queue (at deployment time) can still be processed. 

We'll be able to make this mandatory in a follow-up PR, which will allow us to remove this hack.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/tokens/TokenService.scala
```diff
@@ -9,12 +9,13 @@ import db.{HarvestedToken, RegistrationService, Topic}
...
 import play.api.libs.json.{Format, Json}
 
+import java.time.Instant
 import scala.concurrent.ExecutionContextExecutor
 
 case class IndividualNotification(notification: Notification, token: String)
 case class BatchNotification(notification: Notification, token: List[String])
 
-case class ChunkedTokens(notification: Notification, tokens: List[String], range: ShardRange) {
+case class ChunkedTokens(notification: Notification, tokens: List[String], range: ShardRange, notificationAppReceivedTime: Option[Instant]) {
```
jacobwinch, line 18: This is optional for now so that any messages which are already in the queue (at deployment time) can still be processed; we'll be able to make this mandatory in a follow-up PR.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
```diff
@@ -41,3 +44,45 @@ case class PerformanceMetrics(
   chunkTokenSize: Int,
 )
 
+case class LatencyMetricsForCloudWatch(uniqueValues: List[Long], orderedCounts: List[Int])
+
+object LatencyMetrics {
```
jacobwinch, line 49: Some of the logic here is a little complicated - it's unit tested [here](https://github.com/guardian/mobile-n10n/blob/6090beda81fe2a4cfd9634abda362f5d592b95b5/notificationworkerlambda/src/test/scala/com/gu/notifications/worker/models/SendingResultsSpec.scala).

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/Cloudwatch.scala
```diff
@@ -59,6 +67,23 @@ class CloudwatchImpl(val senderMetricNs: String) extends Cloudwatch {
...
+      val latencyMetrics = LatencyMetrics.aggregateForCloudWatch(deliveryTimes)
+      val dimension = new Dimension().withName("platform").withValue(platform.map(_.toString).getOrElse("unknown"))
+      val requests = latencyMetrics.map { valuesAndCounts =>
+        val cloudWatchMetric: MetricDatum = latencyDatum("TokenDeliveryLatency", valuesAndCounts.uniqueValues, valuesAndCounts.orderedCounts, dimension)
+        new PutMetricDataRequest()
+          .withNamespace(s"Notifications/$stage/$senderMetricNs")
+          .withMetricData(cloudWatchMetric)
+      }
+      if (shouldPushMetricsToAws && latencyMetrics.exists(_.uniqueValues.nonEmpty)) {
+        requests.map(req => cloudwatchClient.putMetricData(req))
```
jacobwinch, line 81: Whilst considering the risks we speculated that there was a possibility that we could:

> Introduce invocation failures (e.g. due to a CloudWatch API call failing)

At first I wondered if all of these calls to CloudWatch should be wrapped in a `Try`, but it looks like `IO.delay` is going to help us out here:

```scala
/**
 * Suspends a synchronous side effect in `IO`.
 *
 * Any exceptions thrown by the effect will be caught and sequenced
 * into the `IO`.
 */
def delay[A](body: => A): IO[A] = {
  val nextIo = Delay(() => body)
  if (isFullStackTracing) {
    IOTracing.decorated(nextIo)
  } else {
    nextIo
  }
}
```

So I think this is safe enough.

DavidLawes:
Now we're nearing the end of our experimentation, I think we're going to remove the ec2 senderworker from this project. The result will be that we were thinking of reverting to using the embedded format for sending metrics to aws. Is the embedded format still something you recommend? Or is it better/more convenient to use the put metric api?

Maybe our implementation has become a bit confusing:
- we previously used the put metric api only
- we added new metrics using the embedded format (plus sent the pre-existing metrics via put metric api)
- when developing the senderworker we sent all metrics via put metric api (because our logs for the ec2 worker bypass cloudwatch entirely)
- now we're considering tearing down the senderworker, we were considering moving back to the hybrid model

Maybe it's simpler to send all metrics in a consistent way? Unless we think the potential cost savings would be significant (something for mss to determine)

jacobwinch:
> Is the embedded format still something you recommend? 

Generally speaking I think that embedded format is likely to be cheaper and more performant within the Lambda runtime. However, looking at [the specification](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html) it's not so clear to me how we would send _many data points for the same metric name_ (which is what we're trying to do in this PR!). Perhaps I'm missing something, but it looks like it might involve making many separate logging calls? This is only really a problem for the new metric that this PR adds, but perhaps it's worth bearing in mind.

> Or is it better/more convenient to use the put metric api?

Although I would generally expect Embedded Metric Format to be more convenient (you can just log JSON rather than including a dependency and fighting to get Scala models into Java!), I recently ran into a strange problem where the Logstash Logback Encoder was not building the required JSON correctly, so we might run into difficulties with this again (we discussed this briefly via Chat but I never got to the bottom of it).

> Maybe it's simpler to send all metrics in a consistent way? 

I definitely think there is an argument for consistency here (this is the main reason I went for PutMetric over EMF in this first draft). Perhaps we are better off favouring consistency for now and migrating everything to EMF if cost/performance becomes a significant enough motivation in the future? Longer term, we are also interested in moving to [OpenTelemetry](https://opentelemetry.io/) as a standard within the department so we might want to avoid migrating twice!

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
DavidLawes: Thanks - the tests do make it clear what's going on :)

DavidLawes:
> I sent a single 'standard' notification (i.e. not a dry-run) to confirm that metrics were actually being pushed to CloudWatch. There were not many datapoints for this; I assume that we must have thousands of 'fake' tokens in CODE?

The CODE RDS proxy is currently pointing to the testRig database, which should have many records. The caveats to this are:
- The testrig database was initially populated with the records from the prod database. We 'sanitised' the tokens to prevent any notifications sent from the CODE environment accidentally making their way onto a user's device.
- If we try to send non-dryRun notifications in CODE when connected to the testRig database then we'll get an invalid token response from firebase/apns, triggering the cleaning of these invalid tokens.
- Over the past couple of weeks or so I think we've triggered non-dryRun notifications in CODE when connected to the testRig database, so I suspect that there's not as many records as we might think, and even less so with valid tokens (that will yield a successful response from firebase/apns).

In this case, I think the limited number of metrics we saw from the test is probably ok! If we wanted to double-check, we could:
- point the rds proxy to the code database
- send a content notification to the profile/francesca-hammond topic
--> we'd expect at least 3 or 4 devices to receive the notification

DavidLawes (approved):
🎉  This looks awesome, will be great to capture these metrics. Thanks so much for thinking about all this @michaelwmcnamara and @jacobwinch !

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/Cloudwatch.scala
```diff
@@ -59,6 +67,23 @@ class CloudwatchImpl(val senderMetricNs: String) extends Cloudwatch {
     }
   }
 
+  def sendLatencyMetrics(shouldPushMetricsToAws: Boolean, stage: String, platform: Option[Platform]): Pipe[IO, List[Long], Unit] = _.evalMap { deliveryTimes =>
+    IO.delay {
+      val latencyMetrics = LatencyMetrics.aggregateForCloudWatch(deliveryTimes)
+      val dimension = new Dimension().withName("platform").withValue(platform.map(_.toString).getOrElse("unknown"))
```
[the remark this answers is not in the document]
jacobwinch: Great point; I will add this!

jacobwinch:
> The CODE RDS proxy is currently pointing to the testRig database, which should have many records. The caveats to this are...

Thanks for this - it's really helpful to understand the current state of the system and it helps me to feel confident that the testing that I've already done is enough to confirm that the integration fundamentally works. 

That said, I guess some of the parts of this that are interesting (i.e. might go wrong!) that we haven't tested yet relate to pushing datapoints at scale (e.g. if we send 150 datapoints in a single batch[^1], or make 7 API calls in quick succession, will we run into problems?). 

I think it might be simplest to test this out by removing [the first part of this conditional](https://github.com/guardian/mobile-n10n/blob/6090beda81fe2a4cfd9634abda362f5d592b95b5/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/Cloudwatch.scala#L80) so that we push metrics for a dry-run too. In the logs for the dry-runs I could see that some invocations were delivering 1000 tokens, so hopefully this would let us do some kind of testing at scale. I will push that change up via a separate branch (since we'll never want to merge it) and do some testing before we push this to `PROD` (I will probably merge this after Christmas now anyway, just in case!).

[^1]: The [API docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html) and [quota limits](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch_limits.html) suggest that this will be fine, but I'd rather find out in `CODE` than `PROD`!

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/utils/Cloudwatch.scala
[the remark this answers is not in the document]
jacobwinch: https://github.com/guardian/mobile-n10n/pull/884/commits/df97dec32c3464805f1a8d855879f2fa6437764b

PR #886, opened 2022-12-20
LIVE-4790: rationalise logging

DavidLawes:
## What does this change?

Given we're reaching the end of our experiments this PR attempts to rationalise the logging we've been adding. The specific changes are:
- Remove the pushy metrics listener: the data this captured didn't help us to understand the performance of the system with better granularity. Now that we've confirmed we can successfully (and consistently) establish more than 1 concurrent connection I didn't think we needed this data anymore.
- Remove some superfluous log messages: there were some log events that I thought no longer add value for longer-term observability/maintenance of the project. Feel free to challenge any of the things I've removed though, I might have overlooked its usefulness.
- Remove the send method from the performance data metrics: previously this was either "batch" or "individual", but this isn't really needed anymore because both platforms are consistent in their sending approach, across all topics.
- Emit the function processing rate metric for the android worker. It might not give us the best overall sense of how well the android worker performs (because the worker processes multiple SQS messages at a time) but it will help us understand if something with the Firebase API changes significantly over time. 

A follow up task that I'd like to address is the logging associated with the latency of every request, for example [here](https://github.com/guardian/mobile-n10n/blob/main/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/apns/ApnsClient.scala#L76) and [here](https://github.com/guardian/mobile-n10n/blob/main/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala#L73). I think it might make sense to address this after we start collecting latency metrics (implemented in [this](https://github.com/guardian/mobile-n10n/pull/884) PR).

## How to test

From local testing I can see that the android worker emits the function processing rate:

Screenshot 2022-12-20 at 17 09 44

This will break some of the kibana dashboards, but currently only the MSS team are using these. I'll make sure to remove/update any visualisations that break.


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -61,9 +59,13 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
...
   def reportBatchSuccesses[C <: DeliveryClient](chunkedTokens: ChunkedTokens, sentTime: Long, functionStartTime: Instant, sqsMessageBatchSize: Int): Pipe[IO, Either[DeliveryException, BatchDeliverySuccess], Unit] = { input =>
     val notificationLog = s"(notification: ${chunkedTokens.notification.id} ${chunkedTokens.range})"
+    val enableAwsMetric = chunkedTokens.notification.dryRun match {
+      case Some(true) => false
+      case _ => true
+    }
     input
       .fold(SendingResults.empty) { case (acc, resp) => SendingResults.aggregateBatch(acc, chunkedTokens.tokens.size, resp) }
-      .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform, isIndividualNotificationSend = false, sqsMessageBatchSize), prefix = s"Results $notificationLog: "))
+      .evalTap(logInfoWithFields(logFields(env, chunkedTokens.notification, chunkedTokens.tokens.size, sentTime, functionStartTime, Configuration.platform, sqsMessageBatchSize), prefix = s"Results $notificationLog: ").andThen(_.map(cloudwatch.sendPerformanceMetrics(env.stage, enableAwsMetric))))
```
waisingyiu, line 68: Are we going to change back to embedded log format to send metric to AWS?  Is it part of this PR?

I have this question because you have not added the `_aws` attribute to the logging in `Logging.scala`.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
```diff
@@ -83,28 +80,16 @@ trait SenderRequestHandler[C <: DeliveryClient] extends Logging {
...
       .map { case (body, sentTimestamp) => (NotificationParser.parseChunkedTokenEvent(body), sentTimestamp, startTime, sqsMessageBatchSize) }
 
-    logger.info("Java version: " + System.getProperty("java.version"))
-
     deliverChunkedTokens(chunkedTokenStream)
       .compile
       .drain
       .unsafeRunSync()
-
-    logEndOfInvocation(sqsMessageBatchSize, totalTokensProcessed, startTime, registry)
```
waisingyiu, line 108: Do you think if it would be useful to have a start and end log with notification ID for each lambda invocation?  I think they may help us to understand when each lambda invocation starts and how long it takes.  They may give us some finer level of information on the concurrency of lambda execution.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
DavidLawes: I did wonder about this. I reverted back to embedded metrics but then I saw Jacob's recent PR and changed my mind: https://github.com/guardian/mobile-n10n/pull/884#issuecomment-1359197473

At the moment I wondered whether it might be better to favour consistency over any cost-benefit. It also looks like there might be a further department-wide migration of how we send metrics...

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
DavidLawes: Good point, maybe I've been too rash with what I've removed...
- in terms of lambda invocation times we get that info from the standard aws metrics
- however, given that I remove this, we would lose the trace in kibana

Looking at Jacob's PR I think we might have a solution for understanding overall latency metrics, but maybe it is good to understand latency on a per-notificationId basis too via kibana (for extra debugging)

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
DavidLawes: Ahhh great :) Thanks for raising this!

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/SenderRequestHandler.scala
DavidLawes: I've added the start and end log messages back in, I think the log trace looks sensible (for now):

Dry run:
Screenshot 2023-01-03 at 12 01 57
^ we can see the beginning and end of the processing by notification id

Not a dry run (sent to only a few subscribers):
Screenshot 2023-01-03 at 12 03 47
^ we continue to log for every api call, but I plan to address this after #884 is complete as it will change what metrics we're capturing and how.

Without filtering by notification id we also see when all sqs messages in a single event have been processed too:
Screenshot 2023-01-03 at 12 09 30

waisingyiu (approved):
Thank you for the work!

PR #903, opened 2023-01-09
LIVE-4829 Deploy worker lambda as zip archives

waisingyiu:
## What does this change?

Currently, the AWS SnapStart feature is enabled on lambdas which are deployed as ZIP archives using Java11 runtime only.   In order to enable SnapStart, we need to change the worker lambda such that it can be deployed as ZIP.

This PR removes some dependencies from the notification worker lambda so that the output archive meets the maximum file size requirement of lambda ZIP archive.  It also deploys the sender lambda using JAR archive.

Note: the PR does not enable SnapStart.  It just makes the lambda meet the requirement.  Since SnapStart also requires lambda to publish versioned lambda, we need riffraff changes to support this versioning deployment.

It makes the following changes
- create a new `commoncore` module
- move the classes in `common` which are needed by worker lambda to this new `commoncore` module
- change the worker lambda dependency from `common` to `commoncore`
- in worker lambda, remove the transitive dependency of firebase which we believe are not used (e.g. firestore, cloud-storage)
- in sender CDK, we create a new set of sender lambdas which are deployed as JAR archive.  Change the event source mapping to invoke the new sender lambdas.
- update riffraff yaml to upload the JAR archives to deployment bucket in S3
- 
This PR keeps both the container and JAR deployment of sender lambda.  If we need to include some large dependencies (for example, for security reason), we may fall back to the container deployment of sender lambda. 

## How to test

I deployed it to `CODE` and sent a notification successfully.  Both an iOS device and an Android emulator received the notification.

The sender lambda can be deployed as JAR and it runs properly.

Although it is the suggestion of the firebase support to remove transitive dependencies of firebase which we believe are not used, we are not completely certain that the firebase APIs which we call do not really invoke any classes / methods from the firestore / cloud-storage dependencies.

But we did test the function of sending notifications on CODE and it worked properly.  Since the logic of our sender is relatively straightforward (in terms of firebase API call), it suggests that the sender lambda is able to work properly with the firestore/cloud-storage dependencies removed.


notificationworkerlambda/cdk/lib/senderworker.ts
```diff
@@ -115,15 +117,35 @@ class SenderWorker extends cdkcore.Construct  {
       reservedConcurrentExecutions: props.reservedConcurrency
     })
 
+    const senderLambdaJar = new lambda.Function(this, `SenderLambdaJar`, {
```
waisingyiu, line 120: We are not allowed to change a lambda from container deployment to JAR deployment (or vice versa)

Here I create a new lambda which will be deployed as JAR, and just change the target lambda of the event source mapping below.   So it will be easy to fall back if we need to use container deployment again.

PR #924, opened 2023-01-24
Remove optionality from notificationAppReceivedTime in models

jacobwinch:
## What does this change?

Now that https://github.com/guardian/mobile-n10n/pull/884 has been merged for a while, we can be sure that all messages in our SQS queues will have the `notificationAppReceivedTime`. This means that we can simplify the models by making this property mandatory.

## How to test

I have deployed this to `CODE` and confirmed that a dry-run notification can still be processed.

We can simplify the models and remove a couple of fallbacks which don't really make sense.

Merging this will make it a little bit harder to rollback https://github.com/guardian/mobile-n10n/pull/884, so I'll wait for a while before merging this one.

### DIFF
--- a/common/src/main/scala/models/ShardedNotification.scala
+++ b/common/src/main/scala/models/ShardedNotification.scala
@@ -15,7 +15,7 @@ object ShardRange {
 case class ShardedNotification(
   notification: Notification,
   range: ShardRange,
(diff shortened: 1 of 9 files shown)


jacobwinch:
Closing this as it's superseded by https://github.com/guardian/mobile-n10n/pull/943.

PR #943, opened 2023-02-03
Categorise notifications based on their audience size to support SLI bucketing

jacobwinch:
cc @michaelwmcnamara (who also worked on this PR)

## What does this change?

This change is a follow-up to https://github.com/guardian/mobile-n10n/pull/884. It modifies the CloudWatch metrics sent from the notifications worker lambdas to include a dimension tracking the audience size for each notification. This replaces the `type` dimension (which captured whether the notification was `breakingNews` or `other`) in order to prevent us from creating too many unique metrics[^1].

The reason for this switch to audience size is:

1. IIUC there is no significant functional difference between the handling of Breaking News and other notifications.
2. The size of the audience for Breaking News notifications varies wildly between 10,000 and more than 1,000,000 (see the [Ophan notifications page](https://dashboard.ophan.co.uk/notifications) for examples).

The audience size of these notifications has been shown to have a very significant impact on the latency of the notification process and it is unrealistic expect the p90 latency for a notification sent to a small audience to be comparable with the p90 latency for a notification sent to a very large audience. This change allows us to bucket notifications by audience size instead, comparing like with like. 

This means that increases in latency can be attributed to performance problems, rather than notifications being sent to a different audience. There is more detail on the rationale for this [bucketing](https://cloud.google.com/stackdriver/docs/solutions/slo-monitoring/sli-metrics/overview#sli-bucket) approach [here](https://docs.google.com/document/d/1ao5-YGkAqN0NJ5x3EVOL0nv86QkMPSWSpw0JYMSSqUU/edit#heading=h.gphldjjod3y0).

## How to test
1. Using a copy of this branch, modify the worker to send metrics on a dry run ([example here](https://github.com/guardian/mobile-n10n/tree/jw-test-bucket-notifications))

2. Deploy this testing branch to code and use the instructions in the Readme docs to send a notification.
- note Make sure you set this notification up as a "dry-run" to prevent any accidental sendings.

- Once the notification has been sent wait 10 mins then, in Cloudwatch, examine the "Notifications/CODE/workers" metrics and confirm the expected Cloudwatch metrics and dimensions are visible.   

We are able to get meaningful and useful data on the internal latency of notifications.

We are able to establish an SLI (or SLIs) which accurately capture performance issues (unfortunately it is [difficult to do this with the current metrics](https://docs.google.com/document/d/1ao5-YGkAqN0NJ5x3EVOL0nv86QkMPSWSpw0JYMSSqUU/edit#heading=h.gphldjjod3y0)).

[^1]: We want to avoid creating too many unique metrics because we will need to use metric math to combine them for alerts/dashboard. We need to strike the balance between having granular data and keeping the number of metrics that we're working with manageable. Furthermore, each unique dimension creates a new custom metric in CloudWatch (and this costs money!).

### DIFF
--- a/common/src/main/scala/models/ShardedNotification.scala
+++ b/common/src/main/scala/models/ShardedNotification.scala
@@ -2,8 +2,6 @@ package models
 
 import play.api.libs.json.{Format, Json}
 
-import java.time.Instant
-
 case class ShardRange(start: Short, end: Short) {
   def range: Range = Range.inclusive(start, end)
 }
@@ -15,7 +13,7 @@ object ShardRange {
 case class ShardedNotification(
   notification: Notification,
   range: ShardRange,
-  notificationAppReceivedTime: Option[Instant]
+  metadata: Option[NotificationMetadata]
 )
 
 object ShardedNotification {
--- a/notification/app/notification/services/guardian/GuardianNotificationSender.scala
+++ b/notification/app/notification/services/guardian/GuardianNotificationSender.scala
@@ -112,7 +112,7 @@ class GuardianNotificationSender(
     }
 
     shard(countWithDefault).map { shard =>
-      val shardedNotification = ShardedNotification(notification, shard, Some(notificationReceivedTime))
+      val shardedNotification = ShardedNotification(notification, shard, Some(NotificationMetadata(notificationReceivedTime, registrationCount)))
       val payloadJson = Json.stringify(Json.toJson(shardedNotification))
       val messageId = s"${notification.id}-${shard.start}-${shard.end}"
       new SendMessageBatchRequestEntry(messageId, payloadJson)
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -45,7 +45,7 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
         deliverBatchNotificationStream(Stream.emits(chunkedTokens.toBatchNotificationToSends).covary[IO])
           .broadcastTo(
             reportBatchSuccesses(chunkedTokens, sentTime, functionStartTime, sqsMessageBatchSize),
-            reportBatchLatency(chunkedTokens, chunkedTokens.notificationAppReceivedTime.getOrElse(functionStartTime)), // FIXME: remove this fallback after initial deployment
+            reportBatchLatency(chunkedTokens, chunkedTokens.metadata.map(_.notificationAppReceivedTime).getOrElse(functionStartTime)), // FIXME: remove this fallback after initial deployment
             cleanupBatchFailures(chunkedTokens.notification.id),
             trackBatchProgress(chunkedTokens.notification.id))
     }.parJoin(maxConcurrency)
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -1,5 +1,6 @@
 package com.gu.notifications.worker
 
+import _root_.models.NotificationMetadata
 import cats.effect.{ContextShift, IO, Timer}
 import com.gu.notifications.worker.cleaning.CleaningClientImpl
 import com.gu.notifications.worker.delivery.DeliveryException.InvalidToken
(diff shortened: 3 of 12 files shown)


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
```diff
@@ -84,5 +84,14 @@ object LatencyMetrics {
...
   }
+
+  def audienceSizeBucket(audienceSize: Option[Int]): String = audienceSize match {
+    case None => "unknown"
+    case Some(audience) if audience < 100000 => "S"
+    case Some(audience) if audience < 500000 => "M"
+    case Some(audience) if audience < 1000000 => "L"
+    case _ => "XL"
+  }
+
```
michaelwmcnamara, line 95: Bucketing has been set initially as:
1 - 99,999 (captures most tag specific notifications)
100,000 - 499,999(captures au or us breaking news notifications)
500,000 - 999,999 (captures uk breaking news notifications)
1,000,000 + (captures combined region breaking notifications)

This was determined by looking at the [ophan-app-notifications page](https://dashboard.ophan.co.uk/notifications)

Numbers can be easily modified if we need to or new dimensions can be added if we need to add more distinctions. Be aware though that each new bucket we add will be adding 5 new dimensions (one each for "android, android-beta, android edition, ios and ios-edition) so do this with care.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -70,14 +71,14 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
...
+  def reportBatchLatency[C <: DeliveryClient](chunkedTokens: ChunkedTokens, metadata: Option[NotificationMetadata]): Pipe[IO, Either[DeliveryException, BatchDeliverySuccess], Unit] = { input =>
     val shouldPushMetricsToAws = chunkedTokens.notification.dryRun match {
       case Some(true) => false
       case _ => true
     }
     input
-      .fold(List.empty[Long]) { case (acc, resp) => LatencyMetrics.collectBatchLatency(acc, resp, notificationSentTime) }
-      .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform, Reporting.notificationTypeForObservability(chunkedTokens.notification)))
+      .fold(List.empty[Long]) { case (acc, resp) => LatencyMetrics.collectBatchLatency(acc, resp, metadata.map(_.notificationAppReceivedTime).getOrElse(Instant.now())) } // FIXME: remove this fallback after initial deployment
+      .through(cloudwatch.sendLatencyMetrics(shouldPushMetricsToAws, env.stage, Configuration.platform, metadata.flatMap(_.audienceSize)))
```
michaelwmcnamara, line 81: This change does 2 things. 
* It changes the source data format to the new NotificationMetadata case class;
* It replaces the notification-type parameter with one that tracks audience size. 
The reason for this switch to audience size is:
1. There is no functional difference between the handling of Breaking News and other notifications;
2. The size of Breaking News notifications varies wildly between 10,000 and more than 1,000,000 (see the [Ophan notifications page](https://dashboard.ophan.co.uk/notifications) for examples).
The size of these notifications has a very significant impact on the latency of the notification process and it is unrealistic to compare one of the small size notifications with the large notifications. This change allows us to bucket notifications by size instead, comparing like with like.

common/src/main/scala/models/ShardedNotification.scala
```diff
@@ -15,7 +13,7 @@ object ShardRange {
 case class ShardedNotification(
   notification: Notification,
   range: ShardRange,
-  notificationAppReceivedTime: Option[Instant]
+  metadata: Option[NotificationMetadata]
```
jacobwinch, line 16: We'll be able to make this mandatory in all of the models in a follow-up PR.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
jacobwinch: > * 1 - 99,999 (most tag specific notifications);

This should also capture breaking sport notifications.

notification/app/notification/services/guardian/GuardianNotificationSender.scala
```diff
@@ -112,7 +112,7 @@ class GuardianNotificationSender(
     }
 
     shard(countWithDefault).map { shard =>
-      val shardedNotification = ShardedNotification(notification, shard, Some(notificationReceivedTime))
+      val shardedNotification = ShardedNotification(notification, shard, Some(NotificationMetadata(notificationReceivedTime, registrationCount)))
```
waisingyiu, line 115: As an aside, the notification API does not query the database at real time to get this `registrationCount`, but retrieves this value from a S3 bucket which are updated by the `topicCounter` lambda regularly.  So strictly speaking, the `registrationCount` value may differ from the total number of devices the sender worker actually attempt to send, but I think it is more than good enough to be used to group notifications based on audience size in this context.

waisingyiu (approved):
Thank you @jacobwinch and @michaelwmcnamara for the great work!  
It will be very useful for us to be able to get metrics around token delivery time broken down by audience size group.

notification/app/notification/services/guardian/GuardianNotificationSender.scala
jacobwinch: Thanks for the extra context @waisingyiu. As you mentioned, I think this will be accurate enough for our purposes, but it's good to understand how this actually works!

PR #946, opened 2023-02-08
Make NotificationMetadata mandatory in models

jacobwinch:
## What does this change?

Now that https://github.com/guardian/mobile-n10n/pull/943 has been merged for a while, we can be sure that all messages in our SQS queues will have the `metadata` field with `NotificationMetadata`. This means that we can simplify the models by making this property mandatory.

## How to test

I have deployed this to `CODE` and confirmed that a dry-run notification can still be processed.

We can simplify the models and remove a couple of fallbacks which don't really make sense.

Merging this makes it slightly harder to rollback https://github.com/guardian/mobile-n10n/pull/943, but that change has been out for ~24 hours and it seems to be working fine, so I think that's OK.

### DIFF
--- a/common/src/main/scala/models/ShardedNotification.scala
+++ b/common/src/main/scala/models/ShardedNotification.scala
@@ -13,7 +13,7 @@ object ShardRange {
 case class ShardedNotification(
   notification: Notification,
   range: ShardRange,
-  metadata: Option[NotificationMetadata]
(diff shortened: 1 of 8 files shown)


DavidLawes (approved):
Thanks :) I've caught up with the [previous](https://github.com/guardian/mobile-n10n/pull/943) change and this looks like a great improvement to the metrics!

PR #947, opened 2023-02-08
Modify bucket sizes based on full topic counts

jacobwinch:
## What does this change?

In https://github.com/guardian/mobile-n10n/pull/943, we configured audience bucket sizes (for tracking purposes). We [based the initial sizing on the delivery data in Ophan](https://github.com/guardian/mobile-n10n/pull/943#discussion_r1095756893), but this seems to differ significantly from the audience sizes for each topic in the DB.[^1] 

Consequently this PR modifies the bucket sizes so that we treat different audiences as intended (for example, we need to increase the lower limit for the XL audience size to ensure that this only matches global breaking news notifications).

## How to test

I have deployed this to `CODE` and confirmed that a dry-run notification can still be processed. I've also added a unit test.

The bucket sizes should now line up with our original intentions, as explained [here]((https://github.com/guardian/mobile-n10n/pull/943#discussion_r1095756893)).

I think this change should be pretty low risk. We should exclude any data prior to this change when establishing baselines; I think it is simple enough to remember this as we'll be calculating the baselines manually.

[^1]: This disparity is slightly puzzling/worrying but I think that's beyond the scope of this PR!

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/models/SendingResults.scala
@@ -87,9 +87,9 @@ object LatencyMetrics {
 
   def audienceSizeBucket(audienceSize: Option[Int]): String = audienceSize match {
     case None => "unknown"
-    case Some(audience) if audience < 100000 => "S"
-    case Some(audience) if audience < 500000 => "M"
-    case Some(audience) if audience < 1000000 => "L"
+    case Some(audience) if audience < 250000 => "S"
+    case Some(audience) if audience < 1000000 => "M"
+    case Some(audience) if audience < 2000000 => "L"
     case _ => "XL"
   }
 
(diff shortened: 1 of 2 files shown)


DavidLawes:
We could query the registrations db to get a count of subscriptions by topic? Perhaps that's what you've done (or it'd be unnecessary) - but thought I'd make the suggestion anyway :)

DavidLawes (approved):
👍

jacobwinch:
> We could query the registrations db to get a count of subscriptions by topic? Perhaps that's what you've done (or it'd be unnecessary) - but thought I'd make the suggestion anyway :)

Thanks! @waisingyiu [pointed out](https://github.com/guardian/mobile-n10n/pull/943#discussion_r1097700769) that the topic counts are coming from [a file in S3](https://github.com/guardian/mobile-n10n/blob/a937f67d141a7774b7c628482975663c37b06dd5/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/TopicCounter.scala#L19-L28), so I downloaded that file and had a look! I think this achieves the same result as running the queries manually, so hopefully this version of the bucket sizes will be OK!

PR #978, opened 2023-03-07
Create github workflows to consolidate Scala Steward PR into one PR

waisingyiu:
## What does this change?

The PR creates github workflows so that the PR created by Scala Steward on `dependency-updates` branch are regularly consolidated into one PR.  We look at the PR to check if it is good to merge in our snyk session every week.

There are no changes to the n10n services.


DavidLawes (approved):
👍

PR #1035, opened 2023-03-24
Remove `MergeStrategy`s that break Scala Steward

rtyley:
These two custom `MergeStrategy`s are **no longer used**:

* `MergeLog4j2PluginCachesStrategy` : introduced with https://github.com/guardian/mobile-n10n/pull/161 and made obsolete with https://github.com/guardian/mobile-n10n/pull/682 . Note that, should you still need something like this elsewhere, https://github.com/idio/sbt-assembly-log4j2 also provides an implementation (but maybe better to just use logback?).
* `MergeMimeTypesStrategy` : introduced with https://github.com/guardian/mobile-n10n/pull/292 and made obsolete with https://github.com/guardian/mobile-n10n/pull/412

Both classes extend `sbtassembly.MergeStrategy`, which was possible up until that type became a _sealed_ trait that could not be extended as of https://github.com/sbt/sbt-assembly/pull/464 (see https://github.com/sbt/sbt-assembly/pull/464#issuecomment-1104467266).

## Scala Steward is broken by these MergeStrategy classes

While the existing build for this project specified an old version of `sbt-assembly`, it appears that the Scala Steward process brings in a newer version (>=v2.0.0) of `sbt-assembly`, that has the sealed-trait change introduced with https://github.com/sbt/sbt-assembly/pull/464. Consequently, **our Scala Steward job for this repo is [currently failing](https://github.com/guardian/scala-steward-public-repos/actions/runs/4513026623/jobs/7947284010#step:6:932)** with errors like:

```
/home/runner/scala-steward/workspace/repos/guardian/mobile-n10n/project/MergeLog4j2PluginCachesStrategy.scala:10:47: illegal inheritance from sealed trait MergeStrategy
class MergeLog4j2PluginCachesStrategy extends MergeStrategy {
```
image

Upgrading `mobile-n10n` to use the latest version of `sbt-assembly` makes this problem apparent with just a regular sbt build (as opposed to the special Scala Steward build). 

The easiest fix to the problem is just to remove the superfluous custom `MergeStrategy`s - **Delete The Obsolete!**


DavidLawes (approved):
Thanks Roberto - I suppose we hadn't done any kind of audit of our project files, so thanks for making sense of what wasn't needed. Had these merge strategies broken scala steward for some time?

PR #1043, opened 2023-03-28
LIVE-4879 remove EC2-based sender

waisingyiu:
## What does this change?

In our previous work on improving the performance of notification delivery pipeline, we created a new EC2-based stack to send notifications to APNS / Firebase API for experimentation.  We concluded that we were not going to take this approach.

This PR is to remove this EC2-based service from the repo and from the project setting.

After this PR, we will tear down the associated cloudformation stack via AWS cloudformation console. 

## How to test

We checked that there was no compiled archives for ec2-sender after teamcity build.  No CDK cloudformation was generated for ec2-sender.

We validated the notification delivery pipeline by sending a notification to debug app.


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
```diff
@@ -76,13 +74,7 @@ trait HarvesterRequestHandler extends Logging {
...
 
   def switchStack(topics: List[_root_.models.Topic]): SqsDeliveryStack = {
-    if (topics.forall(topic => allowedTopicsForEc2Sender.contains(topic.toString))) {
-      logger.info("Switch to EC2-based sender"); 
-      ec2ServiceSet
-    } else {
-      logger.info("Switch to lambda-based sender"); 
-      lambdaServiceSet
-    }
+    lambdaServiceSet
```
DavidLawes, line 77: Now that we only deliver notifications to 1 stack, do you think we should keep this `switchStack` method, or could we simplify further?

DavidLawes (approved):
👍  thanks Wai Sing!!

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/Harvester.scala
waisingyiu: Yes, I think we could remove the `switchStack` function altogether.

PR #1062, opened 2023-04-12
Use encrypted AMI

DavidLawes:
## What does this change?

It's good practice to use encrypted AMIs and this PR updates the riff raff configuration to consume encrypted ones. The n10n repo defines 3 EC2 services that will consume the encrypted recipe: `notification`, `registration` and `report`.

In amigo I cloned the existing recipe and then edited it to request an encrypted copy for the mobile aws account:

Screenshot 2023-04-12 at 15 23 16

I manually triggered a bake and could see that an encrypted copy had been created as part of the process:

Screenshot 2023-04-12 at 15 24 13

We can then find the corresponding encrypted AMI in the mobile account:

Screenshot 2023-04-12 at 15 25 53

## How to test

I updated the riff-raff configuration for the 3 EC2 services in this repo and deployed to CODE:

`notification`: I sent a test notification, the service returns an expected 201 response and I subsequently received the notification.

`registration`: I sent a POST request to register a sample device token. I received a 200 OK response and could subsequently retrieve the device token from the CODE database.

`report`: I was able to retrieve the report for the sent notification via a GET request to this service.

After this, and other associated PRs, are merged we should see that the existing non-encrypted AMI isn't being used, at which point I'll delete it from amigo.

### DIFF
--- a/notification/conf/riff-raff.yaml
+++ b/notification/conf/riff-raff.yaml
@@ -5,8 +5,9 @@ deployments:
     type: cloud-formation
     app: notification
     parameters:
+      amiEncrypted: true
       amiTags:
-        Recipe: jammy-mobile-java11-ARM
+        Recipe: mobile-java11-jammy-ARM
         AmigoStage: PROD
       templatePath: cfn.yaml
       templateStageParameters:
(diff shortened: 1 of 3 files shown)


waisingyiu (approved):
👍

PR #1066, opened 2023-04-14
Set mobile-notifications-api-models version to 1.0.17

silvacb:
### Updates version of mobile-notifications-api-models:

The mobile-notifications-api-model has changed, and the version was bumped in Maven:
Screenshot 2023-04-14 at 15 02 24
Now we need to update the version.sbt file to the next version, which is 1.0.17.

### Updates the readme file:
Screenshot 2023-04-17 at 15 31 13

## When to release version changes?

**If**: The model has changed and the latest version has been released in Maven

**Then**: This version in version.sbt has to be updated to the next available version

**But**: The main branch is protected and we cannot push directly from main, so we need to do this through a separate branch


DavidLawes (approved):
👍

PR #1118, opened 2023-05-23
Dependency updates 2023-05-23

DavidLawes:
## What does this change?

This bumps the sbt version to the latest as well as a number of dependencies. 

The [latest](https://github.com/guardian/pa-football-client/releases/tag/v7.0.7) version of the pa-client makes use of scala-xml 2.1.0 (to allow cross compilation across scala versions 2.12, 2.13 and 3) but this introduced binary incompatibility errors in n10n. From the scala-xml release notes there aren't actually breaking changes between 1.2.x and 2.1.x so we override these warnings using a dependency scheme rule. Note that these rules are required during both build and compilation (hence why it's repeated across both plugins.sbt and build.sbt).

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -26,3 +26,20 @@ addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.1.1")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.9.21")
 
 addSbtPlugin("net.virtual-void" % "sbt-dependency-graph" % "0.10.0-RC1")
+
+/*
+   Without setting VersionScheme.Always here on `scala-xml`, sbt 1.8.0 will raise fatal 'version conflict' errors when
+   used with sbt plugins like `sbt-native-packager`, which currently use sort-of-incompatible versions of the `scala-xml`
+   library. sbt 1.8.0 has upgraded to Scala 2.12.17, which has itself upgraded to `scala-xml` 2.1.0
+   (see https://github.com/sbt/sbt/releases/tag/v1.8.0), but `sbt-native-packager` is currently using `scala-xml` 1.1.1,
+    and the `scala-xml` library declares that it uses specifically 'early-semver' version compatibility (see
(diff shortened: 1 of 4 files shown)


project/plugins.sbt
```diff
@@ -23,6 +23,23 @@ addSbtPlugin("org.foundweekends" % "sbt-bintray" % "0.6.1")
...
+    and the `scala-xml` library declares that it uses specifically 'early-semver' version compatibility (see
+    https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html#versionscheme-librarydependencyschemes-and-sbt-150 ),
+    meaning that for version x.y.z, `x` & `y` *must match exactly* for versions to be considered compatible by sbt.
+
+    By setting VersionScheme.Always here on `scala-xml`, we're overriding its declared version-compatability scheme,
+    choosing to tolerate the risk of binary incompatibility. We consider this a safe operation because when set under
+    `projects/` (ie *not* in `build.sbt` itself) it only affects the compilation of build.sbt, not of the application
+    build itself. Once the build has succeeded, there is no further risk (ie of a runtime exception due to clashing
+    versions of `scala-xml`).
+ */
```
tkgnm, line 44: You've done a really great job of explaining quite a complex issue here, nice work

tkgnm (approved):

DavidLawes:
I think this change is ok after some testing the core services in CODE:

- fake breaking news notification can be triggered, and all expected logs can be seen in kibana (validated services `notification`, `harvester`, `ios` and `android`)
- can register (on the debug app) for new topics, when sending a notification to that topic I received the notification on my debug device (validated services `registration`, `notification`, `harvester`, `android`)

project/plugins.sbt
DavidLawes: thanks, although I can't take credit for this, I pinched this from an equivalent changed added to [mapi](https://github.com/guardian/mobile-apps-api/blob/main/project/plugins.sbt#L34)

PR #1127, opened 2023-06-13
Fix security vulnerabilities related to the netty library

DavidLawes:
## What does this change?

This PR addresses high and medium severity vulnerabilities related to the netty libraries. Although the netty libraries are transitive dependencies they were introducing security vulnerabilities via simple configuration core. We have a [ticket](https://theguardian.atlassian.net/browse/LIVE-5319) in the backlog to make the migration to the latest version of that library, but in the meantime this change addresses these security vulnerabilities.

## How to test

We validated locally (`snyk test`) that these changes address the netty-related security vulnerabilities (where patches are available).


waisingyiu (approved):
👍

PR #1133, opened 2023-06-22
 Correct typo in scala-steward.conf

DavidLawes:
## What does this change?

This corrects a small typo in the scala-steward configuration file that was causing the entire file to be thrown away, meaning some dependencies we'd explicitly pinned were attempting to be bumped. See #1129 for more details.

## How to test

After this change has been merged scala steward should not attempt to bump dependencies specified in the config file.


silvacb (approved):
Thanks for fixing this, such an easy thing to miss!

PR #1149, opened 2023-08-10
Migrate alarm configuration to slo-alerts repo

jacobwinch:
## What does this change?

These alarms (which were first introduced via https://github.com/guardian/mobile-n10n/pull/855) are going to be provisioned via a [different repository](https://github.com/guardian/slo-alerts).

There will be a few small changes to the alarms as part of this migration - most notably:

* We no longer provision these alarms for `CODE`, since having alarms for `CODE` generally creates noise that doesn't require an action (i.e. we expect to break `CODE` more frequently than `PROD` since 'real' users do not rely on it!). We could add a `CODE` alarm via `slo-alerts` if the team think it's really useful, but I don't think it's necessary now that we're no longer testing this new approach.
* We now use a [calculation for `valid events`](https://github.com/guardian/slo-alerts/blob/f715cb3c6cb48d5885d2ebe5d987f85acae4547b/cdk/lib/slo-alerts.ts#L69-L92), rather than relying on the `RequestCount` metric, as we found that the latter produced some unexpected/undesirable results (most notably, it wasn't incremented in certain 5XX error scenarios!)

See https://github.com/guardian/slo-alerts/pull/19 for more details.

## How to test

I've [deployed this to `CODE`](https://riffraff.gutools.co.uk/deployment/view/0e3ddaeb-acde-4a66-b6bd-33b2b9faf0bf) and confirmed that the only deletions are related to alarms (which is desirable in this case):

image
_N.B. launch config deletion at 10:51:49 relates to a previous deployment (AMI update)_

Our (`PROD`) alarm coverage will be the same once this PR and https://github.com/guardian/slo-alerts/pull/19 have been merged, but future updates/improvements (implemented by @guardian/devx-reliability) will be picked up automatically rather than requiring a GuCDK update.

There is a small risk that we've made a mistake during the migration, but I have double checked that the config provisioned via https://github.com/guardian/slo-alerts/pull/19 is correct now that it has been deployed.


cdk/lib/registration.ts
```diff
@@ -17,43 +14,8 @@ export class Registration extends GuStack {
 		// Until this project has been fully migrated to GuCDK you should update the 'old' infrastructure by modifying
 		// the YAML file and then re-running the snapshot tests to confirm that the changes are being pulled through by
 		// CDK
-		const cfnTemplate = new CfnInclude(this, 'YamlTemplate', {
+		new CfnInclude(this, 'YamlTemplate', {
 			templateFile: yamlTemplateFilePath,
 		});
-
-		const registrationLoadBalancer = cfnTemplate.getResource(
```
jacobwinch, line 24: This might look a little worrying, but the snapshots confirm that we're not actually deleting the load balancer 😅  

This is just a _reference_ to the [load balancer which is defined via YAML](https://github.com/guardian/mobile-n10n/blob/ff47b2b73d72bb99ec5935e02a8f372b4cdfb06c/cdk/lib/__snapshots__/registration.test.ts.snap#L1253-L1301).

cdk/lib/registration.ts
```diff
@@ -17,43 +14,8 @@ export class Registration extends GuStack {
 		// Until this project has been fully migrated to GuCDK you should update the 'old' infrastructure by modifying
 		// the YAML file and then re-running the snapshot tests to confirm that the changes are being pulled through by
 		// CDK
-		const cfnTemplate = new CfnInclude(this, 'YamlTemplate', {
+		new CfnInclude(this, 'YamlTemplate', {
```
jacobwinch, line 17: This seems a bit pointless now that we're not actually provisioning any resources via GuCDK, but I don't think it's worth rolling back this wiring (introduced via https://github.com/guardian/mobile-n10n/pull/831) as it will [slightly reduce the amount of work required when we decide to migrate this to a pattern](https://github.com/guardian/cdk/blob/main/docs/migration-guide.md#migration-to-cdk).

cdk/lib/registration.ts
```diff
@@ -17,43 +14,8 @@ export class Registration extends GuStack {
...
-		}
-
-		const applicationServerErrors = loadBalancerMetric('HTTPCode_Backend_5XX');
-		const loadBalancerErrors = loadBalancerMetric('HTTPCode_ELB_5XX');
-		const allRequests = loadBalancerMetric('RequestCount');
-
-		const budgetAlarm = new GuErrorBudgetAlarmExperimental(this, {
-			sloName: `RegistrationAvailability${this.stage}Slo`,
-			sloTarget: 0.995,
-			badEvents: new MathExpression({
```
jacobwinch, line 48: We can generalise the `badEvents` and `validEvents` needed for a 'typical' Availability SLI to work for every load balancer - so this now happens via some code that @guardian/devx-reliability will maintain: https://github.com/guardian/slo-alerts/blob/f715cb3c6cb48d5885d2ebe5d987f85acae4547b/cdk/lib/slo-alerts.ts#L28-L92

DavidLawes (approved):
Thanks Jacob - all looks very sensible!

cdk/lib/registration.ts
waisingyiu: Agree!

PR #1151, opened 2023-08-21
Migrate build to GitHub actions

DavidLawes:
## What does this change?

This PR adds GitHub workflows for the build of each of the apps inside in the n10n repo. Previously the builds were managed by TeamCity, but after this change is merged the corresponding TeamCity config will be deleted.

The current TeamCity configuration builds all projects, on every commit to every branch. The workflows in this PR change things slightly:
- Each project has its own workflow file
- Each project will be built on every commit, but only if the code changed in the commit matches one of the paths specified in the corresponding workflow
- Instead of using sbt-riffraff-artifact, we now upload assets to riff-raff using the Guardian's [action](https://github.com/guardian/actions-riff-raff/) instead.

Each workflow follows one of these patterns:
- Lambda assembly: event-consumer, fake-breaking-news, football, report-extractor, schedule-lambda, slo-monitor. This workflow typically just compiles and tests the project, then uses the sbt `assembly` command to generate a .jar. 
- Debian package: notification, registration, report. This workflow typically compiles and tests the project, then uses the `debian:packageBin` command (using plugin `sbt-native-packager`) to generate a .deb. 
- Docker image: notification-worker-lambdas. This workflow has to do something slightly more fiddly, it needs to upload images to ECR, which requires a valid authorisation token. There is some extra setup and steps in this type of workflow.

NB: not all projects have been converted to CDK, so the CDK synth step is not always required.

### Docker lambdas

The `notificationworkerlambda` project deploys to docker lambdas, the historical reason for this is because the size of the .zip exceeded AWS's maximum size whereas docker lambdas don't impose such a limit.

To use docker lambdas we need to upload images to a remote ECR repository. In order to push images to the repository we need to login to ECR. Previously we managed the docker auth/deploy using script, in the new workflow I've changed things slightly:
- I use an aws [action](https://github.com/aws-actions/amazon-ecr-login) to manage the login to ECR
- I pulled out the docker commands into the workflow step explicitly (I think this helps the readability of the workflow file)

For a login to be successful we need to supply a custom role. I followed a couple of guides, which I'll link inline below, but my starting point was [here](https://github.com/aws-actions/configure-aws-credentials#usage). The custom role has been configured as following:
- Create a new [identity provider](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html) in the `mobile` aws account for the GitHub token provider service
- Define a new role in the account that has a trust relationship with the GitHub identity provider, restricting access only to the n10n repo, following [this](https://github.com/aws-actions/configure-aws-credentials#sample-iam-oidc-cloudformation-template) guide.
- Give the new role permissions to get an ECR token and put an ECR image (a bit of trial and error to work out the correct permissions)
- Provide the role as a repository secret and use the role in the workflow.

I manually created the new role in the aws account, it's name is `mobile-n10n-deploy-role`. I believe I've applied the minimum permissions necessary for the workflow to succeed (including limiting the scope of any permissions applied to just the n10n ECR repository):
- To get the token: `ecr:GetAuthorizationToken`
- To publish an image: `ecr:CompleteLayerUpload`, `ecr:UploadLayerPart`, `ecr:InitiateLayerUpload`,  `ecr:BatchCheckLayerAvailability`, `ecr:PutImage`

Given login was successful, the workflow compiles and tests the code and then uses the docker `publishLocal` command to generate a local image.

A small change I had to make was to explicitly tag the local image using the expected naming convention (`$repository-url:$build-number`). I'm not quite sure how this was working in the TeamCity build, but this current solution does work. If we migrated to using a `docker build` command, I think we can explicitly set a tag value, but I couldn't see anything obvious/immediate in the packager [docs](https://www.scala-sbt.org/sbt-native-packager/formats/docker.html).

Given a local docker image exists with the expected tag we can then push the image to the remote ECR repository.

## Testing

| Service  | Testing |
| ------ | ------ |
| Event consumer | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/99a81944-34b1-4b53-b5a3-12c702deb3a9) to CODE, lambda continued to be triggered every 5 mins, no errors in logs, athena queries were successfully executed. |
| Fake breaking news lambda | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/e2aaeabc-db9c-4886-8e77-b4e61925dcd3) to CODE, lambda executed successfully after being triggered manually via the aws console, expected logs could be found in kibana. |
| Football | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/50073b56-5037-47e7-8832-0066e6b814ed) to CODE. |
| Notification | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/3b7d53ab-6eda-4021-82ad-d23b1f1dbc8e) to CODE, service successfully received a request to send a notification, no errors in logs |
| Notification worker lambda | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/5777db47-66bb-485b-9a00-3fdf042567ec) to CODE, after sending a test notification it was observed that the harvester successfully retrieved tokens from the db and then ios and android lambdas were invoked to deliver the messages (received on real device). For these lambdas the expected logs could be found in kibana. Other services tested: registration-cleaning-worker (continued to process/delete dodgy tokens from a queue), topic-counter (continued to be triggered on expected cron, continued to retrieve/count/store topic counts), expired-registration-cleaner (successfully triggered manually via the aws console) |
| Registration | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/dd802f55-0351-460a-9069-1413cd20a092) to CODE, request to register a new device was successful, corresponding token (with expected topic) could be retrieved from the database. |
| Report | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/8be9d8be-8290-4ea9-8deb-5ca5b5fa5275) to CODE,  api responded successfully to /healthcheck, /notifications and /notifications/:id requests. |
| Report extractor | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/14fffa70-aef3-4de6-b442-e51eeeeb3d22) to CODE, lambda executed successfully after triggering it manually via the aws console, expected logs could be found in kibana (no errors).  |
| Schedule | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/a26bb275-0306-44b8-9dfc-2ab2d904f337) to CODE, lambda continued to be triggered every minute, logs appeared normal (no errors). |
| SLO monitor | Successful [deploy](https://riffraff.gutools.co.uk/deployment/view/fa9b9a91-58bf-421e-8688-b81654e69e7f) to CODE, after sending a breaking news notification in CODE I could then see the lambda invoked successfully, along with expected logs (no unexpected errors reported). |


.github/workflows/event-consumer.yml
```diff
@@ -0,0 +1,53 @@
+name: Event consumer
+
+on:
+  push:
+    branches:
+      - "**"
```
Mark-McCracken, line 6: You can just leave this out
```suggestion
```

.github/workflows/event-consumer.yml
```diff
@@ -0,0 +1,53 @@
...
+        run: sbt "project eventconsumer" "compile" "test" "assembly"
+
+      - name: Upload to riff-raff
+        uses: guardian/actions-riff-raff@v2
+        with:
+          projectName: mobile-n10n:eventconsumer
+          configPath: eventconsumer/riff-raff.yaml
+          contentDirectories: |
+            eventconsumer:
+              - eventconsumer/target/scala-2.13/eventconsumer.jar
```
Mark-McCracken, line 51: Sometimes I've had differences between what ran locally and what came out in github actions, for the version of scala, ie. 2.12 locally vs. 2.13 in GHA, and those could change in theory.
Sometimes to fix this, I've added something like this:

```
      - name: copy jar to root dir
        run: cp eventconsumer/target/scala-*/eventconsumer.jar .
``` 
then in the riffraff action, provided the path at the root directory

.github/workflows/fake-breaking-news.yml
```diff
@@ -0,0 +1,55 @@
+name: Fake breaking news lambda
+
+on:
+  push:
+    branches:
+      - "**"
```
Mark-McCracken, line 6:
```suggestion
```

.github/workflows/football.yml
```diff
@@ -0,0 +1,53 @@
+name: Football
+
+on:
+  push:
+    branches:
+      - "**"
```
Mark-McCracken, line 6:
```suggestion
```

.github/workflows/registration.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          aws-region: eu-west-1
+          role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
+
+      - name: CDK synth
+        run: |
+          cd cdk
+          npm install
+          npm run lint
+          npm test
```
Mark-McCracken, line 55: Since you're writing tests, might be nice to add a test reporter step after this.
This is the one I've used:
https://github.com/dorny/test-reporter/tree/main

and here's an example:
https://github.com/guardian/janus/blob/e41b96a9221c0f2cf5d558990d81b3d8dd981bbc/.github/workflows/build.yml#L55-L64

.github/workflows/football.yml
Mark-McCracken: You can do this for the rest of the file independently

.github/workflows/notification-worker-lambdas.yml
```diff
@@ -0,0 +1,105 @@
...
+        run: |
+          cd notificationworkerlambda/cdk
+          yarn install
+          yarn test
+          yarn synth
+          cd ../../cdk
+          npm install
+          npm run lint
+          npm test
+          npm run synth
```
Mark-McCracken, line 79: just a stylistic thing but I'd probably split this into 2 separate steps with different names, and possibly use the working-directory parameter, but only if you want to.
work directory parameter I feel is slightly more explicit and intuitive, but just my opinion.
Splitting into 2 distinct steps will be more intuitive to future developers debugging an issue, where exactly this has failed. And you never know when Github might add a new feature to show you how long a step has run for historically, to suggest how long might be left, so this might be handy to have split, but neither of these are amazing reasons to do this if you disagree

Mark-McCracken (approved):
Excellent work! (assuming your builds all run successful, which is a big assumption).
All looks fine to me, I've left a few tiny optional comments, but you've definitely got this right.
Lovely code, I could read that all day

.github/workflows/notification-worker-lambdas.yml
DavidLawes: Thanks, I've split these into 2 separate steps.

.github/workflows/event-consumer.yml
DavidLawes: Great idea - I've added this extra step :)

.github/workflows/registration.yml
DavidLawes: I've been looking at the test reports - for the scala tests we use specs2, which output test data in the .stats format. I don't think this is supported by the action. I tested anyway and got an expected [error](https://github.com/guardian/mobile-n10n/actions/runs/5961506739/job/16170846330).

For the cdk tests, I can't see the output of the tests on my local machine, so not entirely sure what path to specify for the reporter. Will do a little more digging.

tkgnm (approved):
Lovely stuff! Thanks for doing this - this looks ginormous. Do we have a ticket to convert the non-cdk stuff in this project to cdk?

DavidLawes:
> Lovely stuff! Thanks for doing this - this looks ginormous. Do we have a ticket to convert the non-cdk stuff in this project to cdk?

Yep, it's a bit big, sorry! About the cdk migration there's a ticket in the backlog https://theguardian.atlassian.net/browse/LIVE-4170

DavidLawes:
After deploying this change:

- All checks passed

Screenshot 2023-08-24 at 11 39 59


- Continuous deployments to prod were successful

Screenshot 2023-08-24 at 11 44 34

DavidLawes:
NB: these workflows make use of some repository secrets, I've made copies of the values of these in SSM:
- AWS_ROLE_ARN: param name `/mobile-n10n/github-build-action/aws-role-arn`
- NOTIFICATION_LAMBDA_REPOSITORY_ID: param name `/mobile-n10n/github-build-action/repository-id`
- NOTIFICATION_LAMBDA_REPOSITORY_URL: param name `/mobile-n10n/github-build-action/repository-url`

PR #1181, opened 2023-11-09
Reconfigure notification sender worker alarms

tkgnm:
## Why

We have many alarms related to this stack in a permanent state of alarm. This is because they're configured to alarm if no invocations in 6 minutes. However....

- The editions app only sends one notification a day at 4am
- As for the live app, it's not uncommon for a notification not to be sent by editors for over 6 minutes. 
- We also have alarms going off in CODE - the CODE stack seems to send a test notification every hour, not every 6 minutes

## What does this change?

This change reconfigures our notification sender worker alarms.

- Editions apps will only alarm if no invocations for one day
- Other alarms will only alarm if no invocations for one hour
- Enables alarms being sent to our P&E/Apps/Server Alerts channel

These changes should mean our alarms are more useful and enable us to act on them in case of something going wrong by subscribing them to our alerts channel. 

## Testing

To test the alerts to our channel, I changed the period to 1 minute and deployed to CODE to ensure we would get alerts. 

image


notificationworkerlambda/cdk/lib/__snapshots__/senderworker.test.ts.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`The MobileAppsRendering stack matches the snapshot 1`] = `
+exports[`The Sender Worker stack matches the snapshot 1`] = `
 Object {
```
tkgnm, line 4: Wrong test name! Probably a copy/paste job.

waisingyiu (approved):
LGTM

notificationworkerlambda/cdk/lib/__snapshots__/senderworker.test.ts.snap
groakland: Good catch

groakland (approved):
LGTM

PR #1184, opened 2023-11-13
LIVE-5319 Bump simple-configuration to 1.5.7

waisingyiu:
## What does this change?

This PR bumps the [simple-configuration](https://github.com/guardian/simple-configuration) library to a newer version 1.5.7 from version 1.5.6

There are API breaking changes in `simple-configuration` between version 1.5.6 and version 1.5.7.  Please refer to guardian/mobile-apps-api#2797 for details.

This PR makes the following code changes for this new version of `simple-configuration`:
1.   In order to detect whether the service is running locally, I added a check on the environment variable "MOBILE_LOCAL_DEV" - if this environment variable is present, it assumes DevIdentity and bypassed the AppIdentity.whoAmI call (bypassing `AppIdentity.whoAmI` in local deployment was recommended in https://github.com/guardian/simple-configuration/pull/28). Otherwise it invoked AppIdentity.whoAmI and it is still able to return a DevIdentity if `whoAmI` fails to load the identity from EC2 tags / lambda environment variables.
2. Pass in the `region` to `SSMConfigurationLocation constructor`
 using the `region` from `AwsIdentity`
3. The simple-configuration library has to be bumped in mobile-logstash-encoder as well (https://github.com/guardian/mobile-logstash-encoder/pull/17). Otherwise, the microservice fails to start up because the mobile-logstash-encoder throws fatal exception due to the breaking API change.

## How to test
It was deployed to `CODE`, and we validated the service by registering a token and sending a notification to debug build.

Logs were written properly:

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -14,8 +14,8 @@ updates.pin = [
   { groupId = "org.scanamo", artifactId="scanamo-testkit", version = "1.0.0-M12-1" },
 
 # A separate task to bump simple-configuration
-  { groupId = "com.gu", artifactId = "simple-configuration-ssm", version = "1.5.6" },
-  { groupId = "com.gu", artifactId = "simple-configuration-core", version = "1.5.6" },
+  { groupId = "com.gu", artifactId = "simple-configuration-ssm", version = "1.5.7" },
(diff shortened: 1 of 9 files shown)


silvacb (approved):
👍

PR #1206, opened 2024-02-28
LIVE-6160 Upgrade firebase to v9.2.0

waisingyiu:
## What does this change?

Google is going to discontinue some of its legacy Firebase Cloud Messaging (FCM) APIs and some of them are being used by our notification service.

We do not invoke the APIs directly.  These APIs are REST APIs where clients have to make HTTP requests directly to the FCM server endpoints.  Instead, we interact with FCM servers [via the Firebase Admin SDK](https://firebase.google.com/docs/cloud-messaging/server),  which provides a programming interface over the REST API.

To migrate from the old APIs being discontinued, we need to upgrade the Firebase Admin SDK to version 9.2.0, and replace our deprecated API call with the new APIs.  ([Firebase Admin Java SDK Release Notes](https://firebase.google.com/support/release-notes/admin/java) ).  Specifically we may have to replace our `sendMulticastAsync` with the new `sendEachForMulticastAysnc` API.  The new API does not allow more than 500 devices to send in one invocation.

This PR upgrades the firebase to version 9.2.0

## How to test

The change was verified on `CODE` where we was able to send notifications successfully.

However, we cannot verify its performance on CODE.  The changes were in the API call which does the actual operation of sending notifications to real devices.  To check the performance impact, we can only run it as an experiment on PROD which has more than 1 million subscribed users.

We expect the performance to be comparable to that before the upgrade.

The new API may have negative impact on the performance.

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
@@ -99,7 +99,7 @@ class FcmClient (firebaseMessaging: FirebaseMessaging, firebaseApp: FirebaseApp,
       import FirebaseHelpers._
       val start = Instant.now
       firebaseMessaging
-        .sendMulticastAsync(message)
+        .sendEachForMulticastAsync(message)
         .asScala
         .onComplete { response =>
           val requestCompletionTime = Instant.now
--- a/build.sbt
+++ b/build.sbt
@@ -410,7 +410,7 @@ lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notifica
     dockerAlias := DockerAlias(registryHost = dockerRepository.value, username = None, name = (Docker / packageName).value, tag = buildNumber),
     libraryDependencies ++= Seq(
       "com.turo" % "pushy" % "0.13.10",
-      "com.google.firebase" % "firebase-admin" % "9.1.1",
+      "com.google.firebase" % "firebase-admin" % "9.2.0",
       "com.google.protobuf" % "protobuf-java" % "3.25.3",
       "com.amazonaws" % "aws-lambda-java-events" % "2.2.9",
       "com.amazonaws" % "aws-java-sdk-sqs" % awsSdkVersion,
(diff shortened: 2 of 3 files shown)


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
```diff
@@ -99,7 +99,7 @@ class FcmClient (firebaseMessaging: FirebaseMessaging, firebaseApp: FirebaseApp,
       import FirebaseHelpers._
       val start = Instant.now
       firebaseMessaging
-        .sendMulticastAsync(message)
+        .sendEachForMulticastAsync(message)
         .asScala
         .onComplete { response =>
```
lindseydew, line 104: It looks like we are only logging out in the case where the future completed successfully. If we're concerned about performance implications would it be worth logging an error on failure here to help track if we are experiencing errors? 
May be overkill, but could potentially introduce an alert if number of failures are of a certain level

lindseydew (commented):
LGTM, just a though about logging an error

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala
waisingyiu: yeah, we do.  The input argument `response` for the `onComplete` call is actually a `Try[BatchResponse]` object.  We handle errors from the API call [here](https://github.com/guardian/mobile-n10n/blob/73ba5e726cc63ec5355f6ce5479dff5b4a9fb13f/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala#L147).  When the API succeeds but only some of the tokens fail, we handle the errors [here](https://github.com/guardian/mobile-n10n/blob/73ba5e726cc63ec5355f6ce5479dff5b4a9fb13f/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala#L120)

lindseydew: Ah gotcha. Looks good

lindseydew (approved):
LGTM 👍

PR #1215, opened 2024-04-10
Run tests in common and api models as part of CI

waisingyiu:
## What does this change?

It was reported that the tests in the common projects such as `common` and `api-models` were not executed as part of CI.

This PR changes the GHA workflows to run tests in the common projects on which the target project depends.

## How to test

The Github action works successfully as expected.

No changes to the source code or build configuration were made.

### DIFF
--- a/.github/workflows/event-consumer.yml
+++ b/.github/workflows/event-consumer.yml
@@ -38,7 +38,7 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
 
       - name: Compile, test and assembly
-        run: sbt "project eventconsumer" "compile" "test" "assembly"
+        run: sbt "project commoneventconsumer" "compile" "test" "project eventconsumer" "compile" "test" "assembly"
 
       - name: Copy jar to root
         run: cp eventconsumer/target/scala-*/eventconsumer.jar .
--- a/.github/workflows/notification.yml
+++ b/.github/workflows/notification.yml
@@ -40,7 +40,7 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
 
       - name: Compile, test and package
-        run: sbt "project notification" "compile" "test" "debian:packageBin"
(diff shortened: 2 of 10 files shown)


.github/workflows/event-consumer.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
 
       - name: Compile, test and assembly
-        run: sbt "project eventconsumer" "compile" "test" "assembly"
+        run: sbt "project commoneventconsumer" "compile" "test" "project eventconsumer" "compile" "test" "assembly"
```
lindseydew, line 41: Do you mean common here? Or is `commoneventconsumer` a shared app?

.github/workflows/notification.yml
```diff
@@ -40,7 +40,7 @@ jobs:
           role-to-assume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
 
       - name: Compile, test and package
-        run: sbt "project notification" "compile" "test" "debian:packageBin"
+        run: sbt "project common" "compile" "test" "project commonscheduledynamodb" "compile" "test" "project notification" "compile" "test" "debian:packageBin"
```
lindseydew, line 43: is `commonscheduledynamodb` a shared app?

lindseydew (approved):
LGTM, just a question about a couple of the shared apps names, but it's been used in a couple of places so it looks deliberate

.github/workflows/event-consumer.yml
waisingyiu: Yeah, the `commoneventconsumer` is the shared app which `eventconsumer` depends on - https://github.com/guardian/mobile-n10n/blob/main/build.sbt#L328

.github/workflows/notification.yml
waisingyiu: Yes, `notification` depends on both `common` and `commonscheduledynamodb` - https://github.com/guardian/mobile-n10n/blob/main/build.sbt#L163

PR #1216, opened 2024-04-11
LIVE-6383 send requests to Firebase API directly

waisingyiu:
## What does this change?

The Firebase cloud message lagacy HTTP endpoint is going to be discontinued this June, and there is no replacement for the multicast API in the HTTP API v1.  We have to explore alternatives to improve the throughput of our notification service in order to achieve the 90in2 SLA we are seeing today with the soo-to-be-expired multicast API.

This PR implements direct firebase HTTP calls over multiplexed HTTP/2 protocol.  As [this Firebase FAQ](https://firebase.google.com/support/faq#fcm-depr-multiple-tokens) suggests, 

The HTTP v1 API over HTTP/2 performs similarly for 99.9% of multicast requests (sending < 100 tokens). For outlier use cases (sending 1000 tokens), it achieves up to a third of the throughput rate, so additional concurrency is needed to optimize for this atypical use case.

We are going to validate this option over small notifications and apply it to larger notifications if we have more confidence on its throughput.

We have made the following changes:
1. implemented client class to directly invoke Firebase's individual send API using JDK's HttpClient which supports multiplexed HTTP/2
2. changed the FcmClient class to support both the multicast API we are using now as well as the individual send API which we have implemented in this PR
3. added a configuration item on the topics to which we send notification using the individual send 
4. added a configuration item to control the concurrency when sending via individual send
5. added test cases

## How to test

The changes were tested on CODE and notification can be sent to my simulator.  However, we only have a very small number of device token on CODE database (compared to PROD) so we need to run this change as an experiment on PROD.

After the changes are deployed to PROD, we may first enable the individual send on content notification first (by setting `tag/` and `contributor/` in the selected topics).

When it is shown to send notifications correctly with reasonable throughput, we may extend the topics to breaking sport notification (such as 'breaking/international-sport`) and monitor its throughput all along.  As we gain confidence, we may gradually extend it to cover other breaking news topic.

Some parameter tuning may be needed during the experiment.

Notifications can be sent with throughput comparable to what we have achieved with multicast API.

It may hit some resources problem or have some latency issues.  That's the reason we started with small notification.


aracho1 (approved):
Thank you for taking the time to pair review this with me @waisingyiu!

PR #1218, opened 2024-04-18
LIVE-6468 Reuse http client across invocations

waisingyiu:
## What does this change?
The PR #1217 creates a `HttpClient` instance for the processing of each SQS client so that we don't send too many requests over the same connection.

However, the AWS lambda service reuses execution environment when it needs to invoke a lambda shortly after one has finished.  So it is possible that within the same Java runtime, the `AndroidSender` instance is used to process SQS messages many times and, as a result, creates too many `HttpClient` instances and the connections that are left idle.  The connections consume the limit on the file descriptors, which is [1024](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html#function-configuration-deployment-and-execution) for Java lambda runtime.  It is likely to happen when we send a huge notification (more than 100,000).

What we really want is that we have dedicated `HttpClient` instances for each SQS message, but we want to reuse these instances in the subsequent invocation if the same runtime environment is reused.

This PR creates 20 `HttpClient` instances upfront (we process at most 20 SQS messages at a time) and uses them when processing the stream of SQS messages.

This PR also creates a new configuration to control the max connection pool size to avoid "too-many-opened-files" exception.

## How to test

We invoked the Android sender locally with two batches of SQS messages (with 3 messages in each batch).  We printed the instance in the log message [here](https://github.com/guardian/mobile-n10n/blob/LIVE-6468-reuse-http-client-across-invocations/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/FcmClient.scala#L79) and we saw from the logs that the three `FcmClient` instances were reused across invocation:

First batch of SQS messages:
```Individual send request completed - com.gu.notifications.worker.delivery.fcm.FcmClient@19d807f9```

```Individual send request completed - com.gu.notifications.worker.delivery.fcm.FcmClient@591354dd```

```Individual send request completed - com.gu.notifications.worker.delivery.fcm.FcmClient@13fa8ac```

Second batch of SQS messages:

```Individual send request completed - com.gu.notifications.worker.delivery.fcm.FcmClient@19d807f9```

```Individual send request completed - com.gu.notifications.worker.delivery.fcm.FcmClient@591354dd```

```Individual send request completed - com.gu.notifications.worker.delivery.fcm.FcmClient@13fa8ac```

The service was also tested on `CODE` and notification was sent successfully to android app on emulator.

### DIFF
--- a/notificationworkerlambda/cdk/lib/senderworker.ts
+++ b/notificationworkerlambda/cdk/lib/senderworker.ts
@@ -108,7 +108,6 @@ class SenderWorker extends cdkcore.Construct  {
         Stack: scope.stack,
         App: id,
         Platform: props.platform,
-        JAVA_TOOL_OPTIONS: "-Djdk.httpclient.connectionPoolSize=2"
       },
       memorySize: 10240,
       description: `sends notifications for ${id}`,
--- a/notificationworkerlambda/cdk/lib/senderworker.ts
+++ b/notificationworkerlambda/cdk/lib/senderworker.ts
@@ -107,7 +107,8 @@ class SenderWorker extends cdkcore.Construct  {
         Stage: scope.stage,
         Stack: scope.stack,
         App: id,
-        Platform: props.platform
+        Platform: props.platform,
+        JAVA_TOOL_OPTIONS: "-Djdk.httpclient.connectionPoolSize=2"
       },
       memorySize: 10240,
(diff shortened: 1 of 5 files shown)


lindseydew (approved):
LGTM 👍 Great work 🔥

PR #1221, opened 2024-04-30
LIVE-6506 parameter for concurrent message processing

waisingyiu:
## What does this change?

We are evaluating the throughput of our new implementation with Firebase individual send operation on PROD.

When we tested it on fairly large notifications (with more than 40,000 subscribers), we saw that the Android sender lambda function threw an exception when it attempted to create threads.  In these cases, the lambda was invoked with nearly 20 SQS messages and when it processed the 17th or 18th messages, it threw the exceptions and stopped working.  It was terminated by AWS lambda service when it went over the 3 minute timeout.  It appears that the lambda runtime may be have sufficient resources / file descriptor space to support 20 HttpClient instances.

This PR addresses this problem by -
1. make it configurable to change the number of HttpClient instances so we can adjust this number in our experiments
2. make it configurable to change the number of messages to process concurrently
3. spread out our requests evenly over the pool of HttpClient instances within a message
4. display Java runtime version and maximum heap size in log
5. add AWS request ID to some logs which have throughput data, which make us easier to compare the throughput

## Accessibility

### DIFF
--- a/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
+++ b/notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
@@ -35,13 +35,15 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
   logger.info(s"Using thread pool size: ${config.threadPoolSize}")
   logger.info(s"Topics for individual send: ${config.allowedTopicsForIndividualSend.mkString(",")}")
   logger.info(s"Concurrency for individual send: ${config.concurrencyForIndividualSend}")
+  logger.info(s"Concurrency for message processing: ${config.concurrencyForMessages}")
+  logger.info(s"HttpClient object pool: ${config.httpClientPoolSize}")
 
   override implicit val ioContextShift: ContextShift[IO] = IO.contextShift(ec)
   override implicit val timer: Timer[IO] = IO.timer(ec)
 
   val fcmFirebase: Try[FcmFirebase] = FcmFirebase(config.fcmConfig, firebaseAppName)
 
-  val fcmClients: Seq[Try[FcmClient]] = Seq.fill(20)(fcmFirebase).map(firebase => firebase.map(FcmClient(_)))
+  val fcmClients: Seq[Try[FcmClient]] = Seq.fill(config.httpClientPoolSize)(fcmFirebase).map(firebase => firebase.map(FcmClient(_)))
 
   val deliveryServiceStream: Stream[IO, Fcm[IO]] = 
     Stream.emits(fcmClients).covary[IO].flatMap(_.fold(e => Stream.raiseError[IO](e), c => Stream.eval[IO, Fcm[IO]]( IO.pure(new Fcm(c)))))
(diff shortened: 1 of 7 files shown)


notificationworkerlambda/src/main/scala/com/gu/notifications/worker/NotificationWorkerLocalRun.scala
```diff
@@ -44,13 +48,26 @@ object NotificationWorkerLocalRun extends App {
     val sqsMessage = new SQSMessage()
     sqsMessage.setBody(Json.stringify(Json.toJson(tokens)))
     sqsMessage.setAttributes((Map("SentTimestamp" -> s"${Instant.now.toEpochMilli}").asJava))
-    event.setRecords(List(sqsMessage).asJava)
+    event.setRecords(List.fill(4)(sqsMessage).asJava)
     event
   }
+  val localTestContext: Context = new Context{
```
lindseydew, line 54: I wonder whether a mock might be more appropriate here if we only need to have one of the variables defined?

lindseydew (approved):
Excellent work 🔥

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
```diff
@@ -35,16 +35,18 @@ class AndroidSender(val config: FcmWorkerConfiguration, val firebaseAppName: Opt
...
   override implicit val timer: Timer[IO] = IO.timer(ec)
 
   val fcmFirebase: Try[FcmFirebase] = FcmFirebase(config.fcmConfig, firebaseAppName)
 
-  val fcmClients: Seq[Try[FcmClient]] = Seq.fill(20)(fcmFirebase).map(firebase => firebase.map(FcmClient(_)))
+  val fcmClients: Seq[Try[FcmClient]] = Seq.fill(config.httpClientPoolSize)(fcmFirebase).map(firebase => firebase.map(FcmClient(_)))
 
   val deliveryServiceStream: Stream[IO, Fcm[IO]] = 
-    Stream.emits(fcmClients).covary[IO].flatMap(_.fold(e => Stream.raiseError[IO](e), c => Stream.eval[IO, Fcm[IO]]( IO.delay(new Fcm(c)))))
+    Stream.emits(fcmClients).covary[IO].flatMap(_.fold(e => Stream.raiseError[IO](e), c => Stream.eval[IO, Fcm[IO]]( IO.pure(new Fcm(c)))))
```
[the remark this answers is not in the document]
waisingyiu: If we create the `IO` with `delay` method, the `new Fcm` will be executed every time we evaluate this `IO` object.  

Now that we want to evaluate this IO object to get the `Fcm` for every device token, it may be good not to create this `Fcm` instance for every device?  So I use `pure` here to create the `Fcm` instance and put this instance into this `IO` object.

The `Fcm` does not have any states actually, so it should be safe to reuse the same instance.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/NotificationWorkerLocalRun.scala
waisingyiu: If we want to use mock objects, we may need to use the `spec2` and extend the class to `Mockito` (correct me if I was wrong), but this `NotificationWorkerLocalRun` is not a test case.  It is more of a command line Java entry class for us to run the Android sender lambda code locally.

lindseydew: Oh I'm sorry, my mistake, I thought this was a test file. Yeah makes sense for a real invocation

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/AndroidSender.scala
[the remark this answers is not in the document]
lindseydew: Thanks for the explanation

PR #1223, opened 2024-05-01
Adopt GHA Scala Library Release Workflow

aracho1:
This replaces the old release process which had developers manually running sbt release on their own laptops - each developer had to obtain their own PGP key and Sonatype credentials, which was an elaborate & fiddly process.

This PR adopts [gha-scala-library-release-workflow](https://github.com/guardian/gha-scala-library-release-workflow), which provides an automated release process that uses single shared set of release credentials, available through GitHub Organisation Secrets (like we already do with NPM) - this workflow allows you to [release a new library version](https://github.com/guardian/gha-scala-library-release-workflow/blob/main/docs/making-a-release.md) with the click of a button!

Most of the changes made here are documented in gha-scala-library-release-workflow's [configuration.md](https://github.com/guardian/gha-scala-library-release-workflow/blob/add-configuration-guidence/docs/configuration.md).

We've granted access to release credentials via [this PR](https://github.com/guardian/github-secret-access/pull/52).

PR #1225, opened 2024-05-01
remove bintray

aracho1:
remove bintray


rtyley (approved):
This is a step we missed in #1223 !

PR #1228, opened 2024-05-07
LIVE-6530 Change Registration API to use a larger instance type t4g.small

waisingyiu:
## What does this change?

We found that the EC2 instances of registration API failed health check from time to time:
Screenshot 2024-05-01 at 12 44 48

It also happened on `CODE`.  I examined the system logs `syslog` on an EC2 instance which failed the health check and saw that the OS process of our Play application was killed because of out of memory, which suggested the instance type not large enough to support the application (and the kinesis agent).  Previously we were using `t4g.micro`, which had 1GB memory.

Screenshot 2024-05-02 at 11 42 34
Screenshot 2024-05-02 at 11 42 56

This PR increases the EC2 instance type to `t4g.small`.

## How to test

Since this PR was deployed to `CODE` five days ago, no EC2 instances failed health check again.  Previously, it happened around once in one or two days on average.


lindseydew (approved):
Good spot 👍

PR #1233, opened 2024-05-16
Add timeout to HTTP requests to Firebase API in Android sender lambda

waisingyiu:
## What does this change?

We notice that occasionally the Firebase API returns an internal error of service unavailable and then our Android sender lambda appears to stop working until it is terminated due to timeout:

Screenshot 2024-05-16 at 13 38 45

Screenshot 2024-05-16 at 13 39 37

It appears that the lambda may be waiting for responses from Firebase API who may have become _unavailable_.

At the moment, we do not set any timeout configuration for requests we send to Firebase API.  According to [API doc](https://docs.oracle.com/en%2Fjava%2Fjavase%2F11%2Fdocs%2Fapi%2F%2F/java.net.http/java/net/http/HttpRequest.Builder.html#timeout(java.time.Duration)), the HTTP client waits indefinitely for responses by default.

This PR adds configurable parameters (`fcm.connectTimeout` and `fcm.requestTimeout`) to set the timeout (in seconds) of making HTTP connections and waiting for responses.

We did a sanity test on `CODE` and we were able to receive notification via Android simulator.


lindseydew (approved):
LGTM 👍 Good spot

PR #1235, opened 2024-05-23
Remove discontinued Firebase multicast API call and bump Firebase Admin SDK to 9.2.0

waisingyiu:
## What does this change?

We have been testing our new implementation of making HTTP requests to the Firebase v1 API on all contents notifications and all breaking notifications, and we are able to achieve the level of throughput we are seeing with our multicast API.  It means we are ready to remove the multicast API calls left in the codebase.

This PR removes 
- all the multicast API calls from the Android senders
- configuration that was used in our experiment and is no longer relevant
- all test cases on multicast API

It also bumps the Firebase Admin SDK to the latest version.

## How to test

I deployed the branch to `CODE`.  I send a test notification and I could receive it on my Android emulator successfully.  No error messages were found on the application log.


aracho1 (approved):
Amazing work 🎉

PR #1250, opened 2024-06-20
Use token auth for Sonatype

rtyley:
Use token auth for Sonatype - see https://github.com/guardian/gha-scala-library-release-workflow/pull/23


waisingyiu (approved):
👍

PR #1284, opened 2024-08-20
Update README.md

aracho1:
## What does this change?
The DB cleaner is no longer written in Rust. It was re-written in Scala some years ago ([PR](https://github.com/guardian/mobile-n10n/pull/491)). It also cleans up token every 300 days instead of 120.


lindseydew (approved):
Good spot 🔍

PR #1285, opened 2024-08-23
Add breaking news notification type for US elections

aracho1:
## What does this change?
Add editionalised breaking news notification type for US elections

- Global (all)
- UK
- US
- Aus
- Europe
- International (ie rest of world)

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #971, opened 2023-03-01
Support breaking/europe notifications alerts

silvacb:
## What does this change?

With the introduction of the Europe edition, it has become necessary to enable the delivery of notifications for _breaking/europe_ and _breaking/europe-sport*_ topics.

*NB: currently the _europe-sport_ notifications in Android are deverted to the _international-sport_ id (as per

-  https://github.com/guardian/android-news-app/pull/8024)

## Covid-19 topics
We have also received confirmation from editorial to remove all of the covid-19-related topics as these are no longer required. This work was done as part of this PR to avoid the need to update the  mobile-notifications-api-models (and following that the facia-tool) twice.

### Unit tests
We have inspected the existing unit tests, and they do not specifically cover adding / removing of topics. Adding a new topic / edition does not break the current unit tests.
We have added a ticket to our backlog to address this in the future as this work is out of scope for this PR: https://theguardian.atlassian.net/browse/LIVE-5084

### Testing in CODE
This has been tested in CODE in several stages:
#### Making sure the old code works
We sent several notifications to an existing topic (specifically _breaking/uk_) to make sure our change didn't break the existing code. As part of this, we registered a new device for uk breaking news, for which we confirmed:

- that the new device token exists in the database of registered tokens for _breaking/uk_ topic
- that the device successfully received the notification
- that the logs successfully recorded the notification sent

#### Making sure the new code works
We then proceeded to test the new change using the same device by changing the Edition in the debug app to Europe, and registered it to receive breaking news notifications. We confirmed:

- that the new device token exists in the database of registered tokens for _breaking/europe_ topic
- that the device successfully received the notification
- that the logs successfully recorded the notification sent

#### Changes to the Breaking News Tool
In order to support sending Europe notifications, we had to make changes to the Breaking News Tool that are described in the PR:
- https://github.com/guardian/facia-tool/pull/1497

####  mobile-notifications-api-models
After the deploy of this PR, we'll release a new version of the mobile-notifications-api-models, which will then be imported in the facia-tool.

## Images
| europe | europe-sport |
| --- | --- |
| IMG_20230315_172919 | IMG_20230315_172938 |

PR #1288, opened 2024-08-29
Add tests for adding/removing a breaking news notification topic

aracho1:
## What does this change?
Add tests for adding or removing a breaking news topic, in relation to #971 and #1285

### DIFF
--- a/notification/test/notification/controllers/MainSpec.scala
+++ b/notification/test/notification/controllers/MainSpec.scala
@@ -60,6 +60,30 @@ class MainSpec(implicit ec: ExecutionEnv) extends PlaySpecification with Mockito
     }
   }
 
+  "adding a new notification topic" should {
+    "successfully add a new topic and send a notification to it" in new MainScope {
+      val newTopic = Topic(Breaking, "new-topic")
+      val topicsWithNew = validTopics :+ newTopic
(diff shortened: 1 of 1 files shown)


notification/test/notification/controllers/MainSpec.scala
```diff
@@ -60,6 +60,30 @@ class MainSpec(implicit ec: ExecutionEnv) extends PlaySpecification with Mockito
...
+      val response = main.pushTopics()(request)
+
+      status(response) should equalTo(CREATED)
+      pushSent must beSome.which(_.topic.toSet must beEqualTo(topicsWithNew.toSet))
+    }
+  }
+
+  "removing a new notification topic" should {
+    "successfully remove a topic and stop sending notifications to it" in new MainScope {
+      val topicToRemove = validTopics.head
```
lindseydew, line 77: Small point, but is this variable being used anywhere? If not let's remove it

lindseydew (approved):
Looks great 🥳 


Just a question about whether a variable is needed

notification/test/notification/controllers/MainSpec.scala
aracho1: ah yeah we're not using it anywhere I will remove it, thanks for spotting

PR #1319, opened 2024-10-08
Update architecture diagram

aracho1:
## What does this change?
Update the architecture diagram to make upstream services more clear (such as mobile content notifications) and also add more granular information about how notifications are updated and delivered via various AWS services and Fastly


lindseydew (approved):
Nice work 👍

PR #1446, opened 2025-04-23
Remove netty from common

lindseydew:
## What does this change?

This removes the netty library from notification worker lambda because an [upgrade](https://github.com/guardian/mobile-n10n/pull/1445) caused our service to break. Since netty is pulled in by the pushy library, we will use their version of netty so to avoid getting version conflicts

## How to test

Deployed to CODE and tested that notifications send on iOS Devices

## Accessibility

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -94,7 +94,6 @@ lazy val commontest = project
     ),
   ))
 
-
 lazy val common = project
   .dependsOn(commoneventconsumer)
   .settings(LocalDynamoDBCommon.settings)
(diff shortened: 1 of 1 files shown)


lindseydew:
> A quick dig back through history suggests these dependencies were moved out the notification worker lambda and into common a ways back:
> 
> [Commit that moves the netty dependencies](https://github.com/guardian/mobile-n10n/commit/4db4e685f6b33504a321ced6da4ac448f3921848)
> 
> This might mean the notification worker lambda project still requires netty to run properly, and after this change it may no longer be available. Of course, the notification worker lambda sub-project may not need netty at all in which case no worries.
> 
> No doubt you're already well aware!

Yep, it gets pulled in by a different 3rd party lib. This is to ensure those versions don't conflict 👍

PR #1729, opened 2026-03-16
Add DynamoDB LiveActivityChannelRepository

vlbee:
## What does this change?

Adds a Repository to read, write and delete live activity channel mappings, using the existing AsyncDynamoClient in Common. 

- This can probably move to Common especially if used by another server for the autostart feature. 
- We should make a call if we want one table per live activity type or one table for all ios live activities (football, cricket, liveblog, elections...) IF the latter we might want to prefix uuids from the diffierent live activity sources eg. "football-[matchId]"

Not tested yet, to be incorporated into channel creation step. 

## Accessibility

### DIFF
--- a/liveactivities/src/main/scala/com/gu/liveactivities/LiveActivityChannelRepository.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/LiveActivityChannelRepository.scala
@@ -12,44 +12,46 @@ import tracking.RepositoryError
 import scala.concurrent.{ExecutionContext, Future}
 import scala.jdk.CollectionConverters._
 
-
 // MODELS /////////////////////////////////////////////
 
 sealed trait LiveActivityData
 case class FootballLiveActivity(
-                                 homeTeam: String,
-                                 awayTeam: String,
-                                 articleUrl: String
-                               ) extends LiveActivityData
+    homeTeam: String,
+    awayTeam: String,
+    articleUrl: String
+) extends LiveActivityData
 
 object FootballLiveActivity {
-  implicit val format: OFormat[FootballLiveActivity] = Json.format[FootballLiveActivity]
+  implicit val format: OFormat[FootballLiveActivity] =
+    Json.format[FootballLiveActivity]
 }
 
 object LiveActivityData {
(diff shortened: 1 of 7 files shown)


waisingyiu (commented):
What value do you think we should use for the ID of the live activity channel repository?

liveactivities/src/main/scala/com/gu/liveactivities/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,98 @@
...
+    def reads(json: JsValue): JsResult[LiveActivityData] =
+      (json \ "type").validate[String].flatMap {
+        case "football" => FootballLiveActivity.format.reads(json)
+        case other      => JsError(s"Unknown LiveActivityData type: $other")
+      }
+  }
+}
+
+case class LiveActivityMapping(
+                                liveActivityId: UUID,
```
waisingyiu: Following our discussion, do you think it would be clearer to use `matchId` for the name of this field?  And we may want to use `String` instead of the `UUID` data type?

liveactivities/src/main/scala/com/gu/liveactivities/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,97 @@
+package com.gu.liveactivities
+
+import aws.AsyncDynamo
+import aws.DynamoJsonConversions.{fromAttributeMap, toAttributeMap}
+import cats.syntax.all._
+import com.amazonaws.services.dynamodbv2.model._
+import org.slf4j.{Logger, LoggerFactory}
+import play.api.libs.json._
+import tracking.Repository.RepositoryResult
+import tracking.RepositoryError
```
vlbee, line 10: Should we move these models out of common/tracking or just duplicate for live activities?

liveactivities/src/main/scala/com/gu/liveactivities/LiveActivityChannelRepository.scala
vlbee: I called it liveActivityId on the repository at least.

waisingyiu (approved):
💯

PR #1730, opened 2026-03-16
Build a CODE stack for live activities with a dynamo table

waisingyiu:
## What does this change?

Add a cloudformation stack for live activities CODE with a dynamo table.


vlbee (approved):
Thanks. 

Is it work making a code stack deployable with riff raff at this stage?

waisingyiu:
> Thanks.
> 
> Is it work making a code stack deployable with riff raff at this stage?

I think we will build the stack incrementally so there will be frequent changes.

Therefore, yes, it would make it easier for us to test if it can be deployed to CODE via riffraff.  I am going to add this in the next PR.

PR #1732, opened 2026-03-18
[Live activities] Add lambda functions

waisingyiu:
## What does this change?

This pull request refactors the code structure to
- improve the organisation of service / model classes
- change the lambda classes to meet the requirement of AWS lambda

It also provisions AWS lambda functions for the channel manager and broadcast functions and updates the CI / riffraff configuration accordingly.

The functions work properly on local machines.  However, more work is needed to get it running on CODE -
1. get the API signing key, probably in the same way as how n10n's sender lambdas are doing
2. serialisation of the input object of the lambda function handlers

It may be good to merge it into our live-activities-prototype first to unblock other work on the prototype.


liveactivities/src/main/scala/com/gu/liveactivities/service/Authentication.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import java.util.Date
+import java.util.concurrent.atomic._
+import com.turo.pushy.apns.auth.ApnsSigningKey
+import com.turo.pushy.apns.auth.AuthenticationToken
+import org.checkerframework.checker.units.qual.A
+
+object Authentication {
+
+  private val authenticationToken2 : AtomicReference[Option[AuthenticationToken]] = new AtomicReference[Option[AuthenticationToken]](None)
```
vlbee, line 11: Is this being used?

cdk/lib/liveactivities.ts
vlbee, line 21: I used 'liveActivityId' on the repository PR that just when in but probably makes sense to use a more generic "Id" key so will amend that.

vlbee (approved):

cdk/lib/liveactivities.ts
vlbee: I've fixed this here: https://github.com/guardian/mobile-n10n/pull/1733
Feel free to merge if approved.

liveactivities/src/main/scala/com/gu/liveactivities/service/Authentication.scala
waisingyiu: Good spot!  I'll see if we can use this atomic reference thing in the next PR.

PR #1739, opened 2026-03-24
Manage match events in dynamo

waisingyiu:
## What does this change?

This pull request adds the functionality of managing live activities in the dynamo table to the Channel and Broadcast lambdas.

The dynamo table has the following columns:

| Column name | Description |
| --- | --- |
| id | Primary key.  The match ID in the football context. |
| channelId | the channel ID that has been created via Apple API for this live activity |
| isChannelActive | indicate whether the channel is still active or has been closed. |
| isLive | indicate whether the live activity is live i.e. the football match is taking place. |
| data | data specific to the live activity |
| competitionId | competition ID of the match |
| lastEventId | event ID of the last message to the live activity |
| lastEventAt | timestamp of the event of the last message |
| createdAt | timestamp when the record was inserted.  Not directly used by the application. |
| lastModifiedAt | timestamp when the record was last modified.  Not directly used by the application. |

The dynamo table repository class has been extended / revised to provide the following operations
- check if the record for a match exists.  The Channel lambda creates a channel only if there is no records for the match.
- create a record for a match with channel ID.  The Channel lambda creates a record after it creates a channel successfully.
- update the lastEvent ID and lastEventAt of a record.  The Broadcast lambda checks for duplicated events / out-of-order events before sending an event to Broadcast API, and update them afterwards.
- update the isChannelActive and isLive of a record.  The Channel lambda updates these two fields after it closes the channel.

I think we may make use of the TTL feature of dynamo table to physically delete records after expiry date.


liveactivities/src/main/scala/com/gu/liveactivities/models/LiveActivities.scala
```diff
@@ -0,0 +1,73 @@
...
+		channelId: String,
+    isChannelActive: Boolean,
+    isLive: Boolean,
+		data: Option[LiveActivityData],
+    competitionId: Option[String],
+    lastEventId: Option[String],
+    lastEventAt: Option[ZonedDateTime]
+)
+object LiveActivityMapping {
+  implicit val reads: Reads[LiveActivityMapping] = (
```
davidfurey: Why can this not be derived automatically?

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,236 @@
...
+    id: String,
+		channelId: String,
+		eventData: Option[LiveActivityData],
+    competitionId: Option[String]): Future[Unit]
+
+  def getMappingById(id: String): Future[LiveActivityMapping]
+  
+  def updateMappingActiveChannel(id: String, isActive: Boolean): Future[Unit]
+
+  def updateMappingSetLive(id: String, isLive: Boolean): Future[Unit]
```
davidfurey:
```suggestion
  def updateMappingLive(id: String, isLive: Boolean): Future[Unit]
```

liveactivities/src/main/scala/com/gu/liveactivities/models/LiveActivities.scala
davidfurey: Why can the automatic serialiser not be used?

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,236 @@
...
+  def updateMappingLastEvent(id: String, lastEventId: Option[String], lastEventUpdate: Option[ZonedDateTime]): Future[Unit]
+  
+  def deleteMappingById(id: String): Future[Unit]
+}
+
+class LiveActivityChannelRepository(client: DynamoDbAsyncClient, tableName: String)(
+    implicit ec: ExecutionContext
+) extends ChannelMappingsRepository with Logging {
+
+  private val idField = "id"
```
davidfurey: What is the difference between a field and a keyName?

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,236 @@
...
+          throw new RepositoryException(ex, "Error checking if live activity mapping exists")
+        }
+      )
+  }
+
+  override def createMapping(
+		id: String,
+		channelId: String,
+		eventData: Option[LiveActivityData],
+    competitionId: Option[String],
```
davidfurey, line 73: Is there a strange mix of tabs and spaces in this PR?

liveactivities/src/main/scala/com/gu/liveactivities/util/DynamoJsonConversions.scala
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Base64
+import scala.jdk.CollectionConverters._
+import play.api.libs.json._
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue
+import software.amazon.awssdk.core.BytesWrapper
+
+import java.nio.ByteBuffer
+import software.amazon.awssdk.core.SdkBytes
+
+object DynamoJsonConversions {
```
davidfurey: Have you considered using https://github.com/scanamo/scanamo instead of having to do this all yourself?

liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
```diff
@@ -8,52 +8,94 @@ import scala.util.Success
...
+    eventTime: Option[ZonedDateTime])
 
 object BroadcastRequest {
-	implicit val jf: Format[BroadcastRequest] = Json.format[BroadcastRequest]
-}
-
-object BroadcastLambda extends RequestStreamHandler{
-
-  implicit val ec: scala.concurrent.ExecutionContext = scala.concurrent.ExecutionContext.global
+  implicit val reads: Reads[BroadcastRequest] = (
```
davidfurey: Why do you need to create a custom serialiser here?

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
```diff
@@ -14,71 +16,76 @@ import play.api.libs.json.Format
 import play.api.libs.json.JsSuccess
 import play.api.libs.json.JsError
 import java.nio.charset.StandardCharsets
+import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
+import software.amazon.awssdk.regions.Region.EU_WEST_1
+import com.gu.liveactivities.util.Logging
+import scala.concurrent.Future
+import com.gu.liveactivities.models.LiveActivityInvalidStateException
 
 // TODO - we should get the channel ID by looking up the match ID in the datastore
```
davidfurey: Is this todo done?

liveactivities/src/main/scala/com/gu/liveactivities/util/DynamoJsonConversions.scala
waisingyiu: Ah this sounds like a good idea!  That `DynamoJsonConversions` was not entirely new.  It came from the `DynamoJsonConversion` helper in the common module but was changed to use AWS SDKv3.  But it feels the `scanamo` may be a better option.

liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
waisingyiu: It's mainly because of the serialisation of the `ZonedDateTime` value in the `eventTime` property.

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
waisingyiu: Yes I will remove this comment.

liveactivities/src/main/scala/com/gu/liveactivities/models/LiveActivities.scala
waisingyiu: It's mainly because of the serialisation of the `ZonedDateTime` value in the `lastEventAt` property.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
waisingyiu: They are the same thing.  I'll align the naming.

liveactivities/src/main/scala/com/gu/liveactivities/models/LiveActivities.scala
waisingyiu: I think we don't need the JSON serialisation at all if we switch to Scanamo.

davidfurey: I think you can probably have a `Json.Formatter[ZoneDateTime]` in scope to resolve this but as you say, might be unnecessary if you move to Scanamo. Perhaps discuss with Roberto to see if Scanamo is a good option.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
waisingyiu: Thank you for spotting these.  I think @vlbee and I have been using different tabs/spaces for indentation.  I've replaced them with whitespaces to make it in line with other subprojects in the codebase.

liveactivities/src/main/scala/com/gu/liveactivities/models/LiveActivities.scala
waisingyiu: Mainly due to the data type `ZonedDateTime`, but now I've create a JSON Format for this class so it can be derived automatically.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,186 @@
...
+        ).flatten
+      val result = scanamo.exec {
+        NonEmptyList.fromList(updates).map(ups =>
+          table.update(idKeyName === id, ups.reduce[UpdateExpression](_ and _))
+        ).sequence
+      }
+      result.flatMap(_ match {
+        case Some(Right(_)) => Future.successful(())
+        case None => {
+          val errorMsg = s"Live activity mapping for id $id not found for update"
```
davidfurey: I think the `None` branch actually occurs of NonEmptyList is empty (which in practice will never happen), rather than when the live activity mapping for id is not found.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -0,0 +1,186 @@
...
+          logger.error(errorMsg)
+          Future.failed(new LiveActivityDataException(id, errorMsg))
+        }
+        case Some(Left(ex)) => {
+          val errorMsg = s"Error updating live activity mapping for id $id - ${DynamoReadError.describe(ex)}"
+          logger.error(errorMsg)
+          Future.failed(new RepositoryException(errorMsg))
+        }
+      })
+    }
```
davidfurey:
```suggestion
  val updates: List[UpdateExpression] = List(
      isChannelActive.map(c => set("isChannelActive", c)),
      isLive.map(l => set("isLive", l)),
      lastEventId.map(eventId => set("lastEventId", eventId)),
      lastEventAt.map(eventAt => set("lastEventAt", dateTimeToString(eventAt))),
    ).flatten

    NonEmptyList.fromList(updates) match {
      case None =>
        logger.warn(s"No fields to update for live activity mapping with id $id")
        Future.successful(())
      case Some(ups) =>
        val result = scanamo.exec {
          val modifiedAt = ZonedDateTime.now()
          val upsWithLastModified = set("lastModifiedAt", dateTimeToString(modifiedAt)) :: ups
          table.update(idKeyName === id, upsWithLastModified.reduce[UpdateExpression](_ and _))
        }

        result.flatMap(_ match {
          case Right(_) => Future.successful(())
          case Left(ex) =>
            val errorMsg = s"Error updating live activity mapping for id $id - ${DynamoReadError.describe(ex)}"
            logger.error(errorMsg)
            Future.failed(new RepositoryException(errorMsg))
        })
    }
```

liveactivities/src/main/scala/com/gu/liveactivities/util/DynamoJsonConversions.scala
davidfurey: Do you still need this?

davidfurey (approved):
Approved with a minor suggestion

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
waisingyiu: A great spot!

liveactivities/src/main/scala/com/gu/liveactivities/util/DynamoJsonConversions.scala
waisingyiu: No, it is not needed.  I'll delete it.

PR #1740, opened 2026-03-25
Ship PA match events from football lambda to live activities test q. 

vlbee:
## What does this change?

Adds a liveActivityPusher to football lambda that sends footbal MatchDataWithArticle to the live activties eventbus. 

Eventbus has a rule to sent to  a test Q to interrogate messages. 

To test in code:

- Set football time machine CODE to a recent match dateTime. 
- Deploy football CODE lambda to code and enable it to poll in Eventbridge Rules. 
- Ensure the _Eventbus_ rule to route messages to the testing Q is alsp enabled. 

## Accessibility


football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
```diff
@@ -87,11 +89,14 @@ object Lambda extends Logging {
...
-      .flatMap(articleSearcher.tryToMatchWithCapiArticle)
-      .map(_.flatMap(eventConsumer.eventsToNotifications))
-      .flatMap(eventFilter.filterNotifications)
-      .flatMap(notificationSender.sendNotifications)
+    val processing = for {
+      footballDataResult <- footballData.pollFootballData(getZonedDateTime())
+      articleMatches <- articleSearcher.tryToMatchWithCapiArticle(footballDataResult)
+      _ = liveActivityPusher.pushToEventbus(articleMatches)
+      notifications = articleMatches.flatMap(eventConsumer.eventsToNotifications)
+      filteredNotifications <- eventFilter.filterNotifications(notifications) // todo what filtering is happening.
```
vlbee, line 97: I believe this is filtering out duplicate notifications, because the lamba will poll pa and create a notification even when there are no new match events so it does not actually need to send that notification. 

We could create the broadcast notification here to replicate this infrastructure in the broadcast lambda I imagine.

waisingyiu: Yes, I think the football lambda [generates the notification ID from the event ID](https://github.com/guardian/mobile-n10n/blob/main/football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusNotificationBuilder.scala#L53) and [keeps a list of which notification IDs](https://github.com/guardian/mobile-n10n/blob/main/football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala#L23) have been sent.  So it could remove duplicates between sending the notification again.  (Even if it sends duplicated notification to the API downstream, it will be removed anyway because of duplicated notification ID).

For live activities, I plan to store the last event ID and last event time in the table so that the broadcast lambda broadcasts messages only when (1) event ID is different and (2) the event is more recent than the last event.

vlbee: There may be events coming in that we do not want to broadcast a change for (Eg. bookings for yellow cards) so we need to filter these out too even though it's a new event.

waisingyiu: I think it may be good to have very similar de-duplication here just to avoid triggering too many duplicated events, but the broadcast lambda still checks for duplicated event before broadcasting messages.

and agree, it would be good to filter events here

waisingyiu (approved):
Just some suggestion, but they can be left to subsequent pull requests.

PR #1750, opened 2026-04-14
Remove Play code from `common` and simplify lambda deps

rtyley:


project/plugins.sbt
```diff
@@ -3,8 +3,6 @@ libraryDependencies ++= Seq(
   "com.github.docker-java" % "docker-java-transport-httpclient5" % "3.5.3"
 )
 
-resolvers += "Typesafe repository" at "https://repo.typesafe.com/typesafe/releases/"
```
rtyley, line 6: * https://github.com/guardian/maintaining-scala-projects/issues/11

PR #1752, opened 2026-04-16
[Live Activities] synthetic events

vlbee:
Replaced by #1764 to sort commit history conflicts

## What does this change?

The goal of this PR is to generate a synthetic match event for CreateChannel  (2hrs before KO) and StartLiveActivity (20min before KO) and generate a LiveActivityPayload via the EventConsumer correctly. 

Next steps in new PRs: 
- push those payloads to the eventbus and route the CreateChannel payload to the channelManager lambda to see a channel created - the event bus model located in a [shared directory here](https://github.com/guardian/mobile-n10n/tree/live-activities-prototype/api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites) needs to match the model consumed by the lambda. 
- determine exactly when to generate an EndActivity event to end and delete channel.  (FT status plus a match result?)

Notes:
The StartActivity payload will not be needed until we iterate push to start functionality later. 
We should place any models shared between services here: https://github.com/guardian/mobile-n10n/tree/live-activities-prototype/api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites

## Accessibility


football/src/main/scala/com/gu/mobile/notifications/football/lib/EventConsumer.scala
```diff
@@ -36,7 +48,8 @@ class EventConsumer(
       event: MatchEvent,
       articleId: Option[String]
   ): List[NotificationPayload] = {
-    logger.debug(s"Processing event $event for match ${matchDay.id}")
+
+//  logger.debug(s"Processing event $event for match ${matchDay.id}")
```
JamieB-gu, line 52: Why are the debug log statements in this file commented out?

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventConsumer.scala
```diff
@@ -20,10 +20,22 @@ class EventConsumer(
   def eventsToNotifications(
       matchData: MatchDataWithArticle
   ): List[NotificationPayload] = {
-    matchData.allEvents.flatMap { event =>
+
+    /** We have added synthetic events for live activities which we don't want
+      * to generate push notifications for, so we filter those out here.
+      */
+    val liveActivityEventTypes =
+      List("create-channel", "start-live-activity", "end-live-activity")
```
JamieB-gu, line 28: Is there any way we can couple this list to the events we're creating in `SyntheticMatchEventGenerator`? It occurs to me that we have some hardcoded strings in a couple of different places and we're relying on them to always remain in sync? If one or other of them changes we may end up accidentally sending notifications for live activity events?

It's possible we could do this with something like a union type containing members that have `toString` methods? This could be one to capture and look at in a follow-up.

football/src/test/scala/com/gu/mobile/notifications/football/lib/SyntheticMatchEventGeneratorSpec.scala
```diff
@@ -71,4 +71,27 @@ class SyntheticMatchEventGeneratorSpec extends Specification {
...
+    }
+
+    "Add a startLiveActivity event if match kick off is within 20min" in new TestScope {
+      val generator = new SyntheticMatchEventGenerator()
+      generator.generate(List(), "match-id", matchInfo.copy(date = ZonedDateTime.now.plusMinutes(15))).head.eventType mustEqual "start-live-activity"
+    }
+
+    "Add id to first timeline event" in new TestScope {
+      val generator = new SyntheticMatchEventGenerator()
+      println(generator.generate(List(timelineEvent), "match-id", matchInfo.copy(date = ZonedDateTime.now())))
```
JamieB-gu, line 88: Do you need this `println`?

JamieB-gu (approved):
Looks good, some questions, non-blocking.

PR #1754, opened 2026-04-17
Add dynamodb for live activity events

marjisound:
## What does this change?
Add a DynamoDB table to store football live activity events, enabling event deduplication


marjisound:
This work was done in https://github.com/guardian/mobile-n10n/pull/1766

PR #1766, opened 2026-04-22
Deduplicate liveactivities payload events

marjisound:
## What does this change?
Adds a DynamoDB table to store football live activity events, enabling reliable event deduplication.

This PR does the followings:

**Infrastructure**
- Introduces a new DynamoDB table for storing live activities payload events
- Configures the table with on-demand (PAY_PER_REQUEST) billing
This is chosen because the Lambda only interacts with the table during live matches, making provisioned capacity inefficient for sporadic usage
- Adds CloudWatch alarms for read/write throttling on the table
- Adds metrics to track successful and failed event processing

**Implementation**
- Refactors DynamoDistinctCheck to be generic, allowing reuse across both notifications and live activity events
- Introduces a deterministic ID for deduplicating match events (based on matchId + eventId)

**Notes / Considerations**
- The deduplication strategy ensures idempotent processing of live activity events
- The on-demand billing model aligns cost with actual usage patterns during live matches

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
@@ -63,11 +63,6 @@ object Lambda extends Logging {
 
   lazy val liveActivityHandler = new LiveActivityHandler(configuration, dynamoDBClient, liveActivitiesTableName)
 
-  // live activities //
-  lazy val liveActivityPusher = new LiveActivityPusher()
-  lazy val matchStatusLiveActivityPayloadBuilder = new MatchStatusLiveActivityPayloadBuilder(configuration.mapiHost)
-  lazy val liveActivityEventConsumer = new LiveActivityEventConsumer(matchStatusLiveActivityPayloadBuilder)
-
   def getZonedDateTime(): ZonedDateTime = {
     val zonedDateTime = if (configuration.stage == "CODE") {
       val is = URI.create("https://hdjq4n85yi.execute-api.eu-west-1.amazonaws.com/Prod/getTime").toURL.openStream()
--- a/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
@@ -6,7 +6,7 @@ import java.util.concurrent.TimeUnit
 import com.amazonaws.regions.Regions
 import com.amazonaws.services.dynamodbv2.{AmazonDynamoDBAsync, AmazonDynamoDBAsyncClientBuilder}
 import com.gu.contentapi.client.GuardianContentClient
-import com.gu.mobile.notifications.football.lib.{ArticleSearcher, DynamoDistinctCheck, EventConsumer, LiveActivityEventConsumer, EventFilter, FootballData, LiveActivityPusher, NotificationHttpProvider, NotificationSender, NotificationsApiClient, PaFootballClient, SyntheticMatchEventGenerator}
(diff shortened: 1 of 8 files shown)


football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
```diff
@@ -58,12 +61,7 @@ object Lambda extends Logging {
 
   lazy val notificationHandler = new NotificationHandler(configuration, apiClient, dynamoDBClient, tableName)
 
-  lazy val liveActivityHandler = new LiveActivityHandler(configuration, dynamoDBClient, tableName)
-
-  // live activities //
-  lazy val liveActivityPusher = new LiveActivityPusher()
```
vlbee, line 64: good spot!

football/src/main/scala/com/gu/mobile/notifications/football/lib/DynamoDistinctCheck.scala
vlbee, line 20: Should this be narrowed to NotificationPayload?

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala
```diff
@@ -19,24 +19,24 @@ class EventFilter[A <: Payload](distinctCheck: DynamoDistinctCheck[A]) extends L
...
+          Some(item)
         case Duplicate =>
-          cache(notification.id)
+          cache(item.id)
           None
         case _ => None
       }
     } else {
-      logger.debug(s"Event ${notification.id} already exists in local cache or does not have an id - discarding")
+      logger.debug(s"Event ${item.id} already exists in local cache or does not have an id - discarding")
```
vlbee: Shoule we specify in the comment which handler (notifications or liveActivities) this filtering has failed?

vlbee (approved):
This looks good. Some small comments.

football/src/main/scala/com/gu/mobile/notifications/football/lib/DynamoDistinctCheck.scala
marjisound: The DynamoMatchNotification is the payload that's currently in use in production, and is specific to football (hence the field matchId), and changing the dynamo table schema for that (to rename the matchId table field) would be a bit more involved. Maybe that could be a nice to do thing in the future?

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala
marjisound: This isn’t a failure case, it happens when an item is already present in the local cache (in-memory), so we skip adding it to DynamoDB. That’s why it’s logged at debug level, and since our log level is set to info, this won’t be emitted in practice.

In here we also don’t have visibility on which handler the event came from. It would be possible to pass that as a parameter to EventFilter, but that would make the interface a bit messy.

For now, I’ve updated the log to include the DynamoDB table the event would have been written to.

football/src/main/scala/com/gu/mobile/notifications/football/lib/DynamoDistinctCheck.scala
marjisound: Just for visibility, Vanessa's point was about narrowing this function def apply[A <: Payload] to depend on `NotificationPayload` rather than `Payload`. This change was done as part of this commit https://github.com/guardian/mobile-n10n/pull/1766/changes/b367225ba3120dcc25015c26967c9f582202105f

PR #1767, opened 2026-04-23
Add event bus rule to route payloads

marjisound:
## What does this change?

This PR introduces EventBridge rules to route football live activity events to the appropriate Lambda consumers based on their event type.

Two rules are added:

- **ChannelRule**
Routes channel-related events (channel-create, channel-delete) to channelLambda.
- **BroadcastRule**
Routes broadcast-related events (broadcast-start, broadcast-update, broadcast-end) to broadcastLambda.

Both rules listen to events emitted from the football-lambda source and use detail-type to determine routing.

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/lib/LiveActivityPusher.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/lib/LiveActivityPusher.scala
@@ -56,7 +56,6 @@ class LiveActivityPusher extends Logging {
         case UpdateLiveActivity => "broadcast-update"
         case EndLiveActivity    => "broadcast-end"
         case DeleteChannel      => "channel-delete"
-        case _ => "unknown"
       }
 
       val result = Try {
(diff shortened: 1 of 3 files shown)


football/src/main/scala/com/gu/mobile/notifications/football/lib/LiveActivityPusher.scala
```diff
@@ -56,7 +56,6 @@ class LiveActivityPusher extends Logging {
         case UpdateLiveActivity => "broadcast-update"
         case EndLiveActivity    => "broadcast-end"
         case DeleteChannel      => "channel-delete"
-        case _ => "unknown"
```
marjisound, line 59: Removed the fallback case _ => "unknown" to make the match exhaustive. Since LiveActivityEventType is sealed and we fail on non-exhaustive matches, any new event type will now cause a compile error instead of being silently mapped to "unknown" and not matching any Amazon EventBridge rules.

vlbee (approved):

PR #1770, opened 2026-04-28
Impeove channel lambda error handling

marjisound:
## What does this change?

**Refined Timeout:**
App-level Timeout (160s): Calculated based on the maximum cumulative latency of downstream calls ($10s \text{ (Read)} + 90s \text{ (Create Channel)} + 10s \text{ (Write)}$) plus a 50s buffer to account for cold starts, authentication token fetches, potential network congestion, JSON parsing, etc.

Lambda Global Timeout (180s): Set to 3 minutes to provide a 20s grace period over the application logic. This ensures the application stays in control and can successfully log specific error context to CloudWatch before a hard termination by the AWS runtime.

**DynamoDB Client Timeout:**
Configured request timeouts (apiCallAttemptTimeout and apiCallTimeout) to prevent the client from hanging indefinitely. This ensures the system fails fast and recovers quickly if DynamoDB is slow or the network is congested.

**Repository Error Handling:**
Implemented custom error recovery in the repository layer to wrap SDK exceptions with domain-specific context (e.g., including the record ID in logs).

**Graceful State Handling**
Updated the create and close flows to check for "Already Exists" or "Already Inactive" states.

Instead of throwing an error, the system now logs the state and exits successfully. This prevents unnecessary retries and prevents DLQ "pollution" when the system has already achieved the desired outcome.
Asynchronous Reliability:

**SQS Destination (On Failure):**
Added an SQS Destination (On Failure) to capture failed events.

Unlike a standard DLQ, this provides the full responsePayload (stack trace and error message) directly in the SQS message for easier debugging.

## Test
Fully tested this locally to check the timeouts and error handling by enforcing delay or error. 

Also tested this in CODE to check the message that goes to the dead letter queue. The message looks like this:
```
{
  "version": "1.0",
  "timestamp": "2026-04-28T14:16:14.107Z",
  "requestContext": {
    "requestId": "someId",
    "functionArn": "<arn>",
    "condition": "RetriesExhausted",
    "approximateInvokeCount": 3
  },
  "requestPayload": {
    "matchId": "testId",
    "competitionId": "competiton-001",
    "toCreate": false
  },
  "responseContext": {
    "statusCode":<status code>,
    "executedVersion": "<version>",
    "functionError": "Unhandled"
  },
  "responsePayload": {
    "errorMessage": "<error message>",
    "errorType": "<error type>",
    "stackTrace": [<stack trace> ]
  }
}
```

### DIFF
--- a/liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
@@ -28,7 +28,7 @@ object ChannelManagerLambda extends RequestStreamHandler with Lambda with Loggin
     maybeChannelId.flatMap {
       case Some(id) =>
         logger.error(s"Channel mapping already exists for match ID $matchId")
-        // id channel id already exists for a match we don't want to fail the lambda to avoid the lambda retry on this event
+        // if channel id already exists for a match we don't want to fail the lambda to avoid the lambda retry on this event
         Future.successful(id)
       case None =>
         for {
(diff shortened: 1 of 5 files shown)


liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
```diff
@@ -33,31 +22,37 @@ object ChannelManagerLambda extends RequestStreamHandler with Lambda with Loggin
...
-        } else Future.successful(())
-      channelId <- channelApiClient.createChannel()
-      _ <- repository.createMapping(matchId, channelId, eventData, competitionId)
-      _ = logger.info(s"Channel created with channel ID ${channelId} for match ID ${matchId}")
-    } yield channelId
+    logger.info(s"Received request to create channel for match ID $matchId")
+
+    val maybeChannelId = repository.containMapping(matchId)
+    maybeChannelId.flatMap {
+      case Some(id) =>
```
vlbee, line 29: I thought using for comps here all the way through was slightly clearer to read but up to you.

vlbee (approved):
Very thorough. ✨

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
marjisound: I see your point on consistency with for-comprehensions. I moved containMapping out of the for because we now branch based on whether a mapping exists and may short-circuit with success early instead of failing. 

We could have done it with for comprehension too: 
```

    for {
      maybeChannelId <- repository.containMapping(matchId)
      result <- maybeChannelId match {
        case Some(id) =>
          logger.error(s"Channel mapping already exists for match ID $matchId")
          Future.successful(id)

        case None =>
          for {
            channelId <- channelApiClient.createChannel()
            _ <- repository.createMapping(matchId, channelId, eventData, competitionId)
            _ = logger.info(s"Channel created with channel ID $channelId for match ID $matchId")
          } yield channelId
      }
    } yield result
```

But this way we'll end up with a for with another for. Which one do you think is more readable?

marjisound: I’ll merge this for now, but happy to revisit if you think a for-comprehension version would be clearer

PR #1773, opened 2026-04-29
Add Live Activity topics

JamieB-gu:
When a user is subscribed to a team, we will offer them a live activity whenever that team is playing a match, as an alternative to push notifications throughout the match. This live activity will be provided via a single notification when the match starts, and users with app versions that can support live activities will receive it by subscribing to topics of type `FootballTeamLiveActivity`. Users with app versions that do not support live activities will instead be subscribing to topics of type `FootballTeam`, which will deliver notifications throughout the match.

A user may visit a match info page before a live activity for that match is available, as we only create it a couple of hours before the match starts. We still want them to be able to subscribe and get that live activity when it's ready, so we'll allow them to sign up to this notification topic instead. The live activity will then be provided via a single notification when the match starts. Users with app versions that do not support live activities will instead be subscribing to topics of type `FootballMatch`, which will deliver notifications throughout the match.

This change adds two new `TopicType`s, `FootballTeamLiveActivity` and `FootballMatchLiveActivity`, and includes them in the notification payload for kick-off events only.


guardian-ben:
@JamieB-gu we'll need something similar for "football-match-live-activity" too.

vlbee:
Is releasing this blocked on updating the Push Notification copy for kick off notifications or any native changes being release? Presumably the notification service just won't send for these new topics if no device has subscribed to them.

JamieB-gu:
> @JamieB-gu we'll need something similar for "football-match-live-activity" too.

Good point, added that too.

> Is releasing this blocked on updating the Push Notification copy for kick off notifications or any native changes being release? Presumably the notification service just won't send for these new topics if no device has subscribed to them.

We're not changing the copy, so it'll just be the normal kick-off notification payload. I don't believe it's blocked on any native changes because, as you say, it's a new topic so there'll be no devices subscribed to it until @guardian-ben makes changes on iOS.

marjisound (approved):
Looks good 👍

PR #1774, opened 2026-04-30
Process channel events

marjisound:
## What does this change?
This PR ensures that the channel manager successfully receives and processes the vent bridge events. The following changes were implemented:
- Created types for `source` & `detail-type` within the `EventBridgeEvent` to avoid using String for them

- Shared the existing LiveActivityEventType type between `LiveActivityPayload.eventType` & `EventBridgeEvent.detail-type`. This is done to avoid consusion and duplication of types that were very similar
- Parameterized the date in SyntheticMatchEventGenerator to keep it in sync with PA polling, ensuring reliable event generation when using historical data in the CODE environment.
https://github.com/guardian/mobile-n10n/blob/main/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala#L66
- Modified the Channel Manager to accept the structured EventBridgeEvent as the primary input.
- Included competitionId in the content state to improve downstream event context.

## Test
Fully tested this locally and also in CODE. Set the time machine to about 1 hour before the start of match, and noticed the channel manager create a new channel. Then when the match started, the broadcast events were successfully processed by the broadcast manager.

### DIFF
--- a/api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/BroadcastContentState.scala
+++ b/api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/BroadcastContentState.scala
@@ -60,6 +60,7 @@ object MatchStatus {
 }
 
 case class Competition(
+    id: String,
     name: String,
     round: Option[String] = None // World Cup Group name
 )
--- a/api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/LiveActivityPayloads.scala
+++ b/api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/LiveActivityPayloads.scala
@@ -11,11 +11,63 @@ import com.gu.mobile.notifications.client.models.Payload
  * Channel Manager and Broadcast lambdas.
  **/
 
+sealed trait EventSource
+case object FootballLambda extends EventSource
+
+object EventSource {
+  implicit val format: Format[EventSource] = new Format[EventSource] {
+    def writes(source: EventSource): JsValue = source match {
+      case FootballLambda => JsString("football-lambda")
+    }
+
+    def reads(json: JsValue): JsResult[EventSource] = json match {
+      case JsString("football-lambda") => JsSuccess(FootballLambda)
+      case JsString(other)             => JsError(s"Invalid EventSource: $other")
+      case _ => JsError("EventSource must be a string")
+    }
+  }
+}
+
+sealed trait LiveActivityEventType {
+  def asString: String
+}
+
+case object CreateChannelEvent extends LiveActivityEventType { val asString = "channel-create" }
+case object StartLiveActivityEvent extends LiveActivityEventType { val asString = "broadcast-start" }
+case object UpdateLiveActivityEvent extends LiveActivityEventType { val asString = "broadcast-update" }
+case object EndLiveActivityEvent extends LiveActivityEventType { val asString = "broadcast-end" }
+case object DeleteChannelEvent extends LiveActivityEventType { val asString = "channel-delete" }
+
+object LiveActivityEventType {
+  val values: Seq[LiveActivityEventType] = Seq(
+    CreateChannelEvent,
+    StartLiveActivityEvent,
+    UpdateLiveActivityEvent,
+    EndLiveActivityEvent,
+    DeleteChannelEvent
+  )
+
+  implicit val format: Format[LiveActivityEventType] = new Format[LiveActivityEventType] {
+    override def reads(json: JsValue): JsResult[LiveActivityEventType] = json match {
+      case JsString("channel-create")    => JsSuccess(CreateChannelEvent)
+      case JsString("broadcast-start")   => JsSuccess(StartLiveActivityEvent)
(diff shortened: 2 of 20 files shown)


api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/BroadcastContentState.scala
```diff
@@ -60,6 +60,7 @@ object MatchStatus {
 }
 
 case class Competition(
+    id: String,
```
vlbee, line 63: Hi. This is not included in the iOS app native model so need to confirm adding here does not impact broadcast updates?

api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/LiveActivityPayloads.scala
```diff
@@ -11,11 +11,63 @@ import com.gu.mobile.notifications.client.models.Payload
...
+      case _ => JsError("EventSource must be a string")
+    }
+  }
+}
+
+sealed trait LiveActivityEventType {
+  def asString: String
+}
+
+case object CreateChannelEvent extends LiveActivityEventType { val asString = "channel-create" }
```
vlbee, line 35: Great, thank you - this was needed!

api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/LiveActivityPayloads.scala
vlbee: Was this not going to be moved?

football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
```diff
@@ -69,7 +69,7 @@ class FootballData(
 
   def matchIdsInProgress(dateTime: ZonedDateTime): Future[List[MatchDay]] = {
     def inProgress(m: MatchDay): Boolean =
-      m.date.minusMinutes(5).isBefore(dateTime) && m.date.plusHours(4).isAfter(dateTime)
+      m.date.minusHours(2).isBefore(dateTime) && m.date.plusHours(4).isAfter(dateTime)
```
vlbee, line 72: What is the impact if any of this on the existing push notification service? Presumably nothing, as the first event sent for pushes is kick off?

football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
```diff
@@ -49,7 +49,7 @@ object Lambda extends Logging {
 
   lazy val capiClient = GuardianContentClient(configuration.capiApiKey)
 
-  lazy val syntheticMatchEventGenerator = new SyntheticMatchEventGenerator()
+  lazy val syntheticMatchEventGenerator = new SyntheticMatchEventGenerator(getZonedDateTime())
```
vlbee, line 52: Is this just for the time machine or was this impacting actual matches as well? good catch.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -58,16 +57,14 @@ class LiveActivityChannelRepository(client: DynamoDbAsyncClient, tableName: Stri
...
   ): Future[Unit] = {
     val createdAt = ZonedDateTime.now()
     val newItem = new LiveActivityMapping(
       id = id, 
       channelId = channelId, 
       isChannelActive = true, 
-      isLive = true, 
-      data = eventData, 
-      competitionId = competitionId,
+      isLive = true,
```
vlbee, line 66: Presumably we will be amending this to isFalse and then only isLive once the broadcasts lamba is triggered to send first broadcast for placeholder content state to be available?

vlbee (approved):

api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/LiveActivityPayloads.scala
marjisound: Good spot, this type `dynamoData` is not used anywhere, just deleted it.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
marjisound: Yes exactly, this will be done as part of the work for fixing the LiveActivity lifecycle ticket

api-models/src/main/scala/com/gu/mobile.notifications.client/models/liveActitivites/BroadcastContentState.scala
marjisound: Tested this together

football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
marjisound: This shouldn't have any impact on notifications. But we tested this together and we received notifications when the match started.

PR #1779, opened 2026-05-05
Support new Android football notification fields

aracho1:
## What does this change?
Extend the Android football notification to support more fields and also adds a new penalty shootout alert type.

Key changes
**1. Add new fields in `FootballMatchAlert`** 
- kickOffTimestamp (epoch seconds)
- lineupsAvailable (boolean) 
- detailedMatchStatus (detailed match phase strings which match what's used in the iOS activities so that the client can use this info to change the notification background colour etc)
- homeTeamRedCards / awayTeamRedCards: red card count for each team

**2. Create a new `FootballShootoutPenalty`**
- We decided to create a new type so that old Android builds which don't expect the penalty shootout notification types silently drop them rather than creating empty notifications. Only clients that can explicitly handle this new type will receive shootout notifications.

The new type includes all fields from the standard match alert, plus:
- homeTeamPenaltiesScored/MissedSaved
- awayTeamPenaltiesScored/Missed/Saved

Tested on CODE
Screenshot_20260519_185347


common/src/main/scala/models/TopicType.scala
aracho1, line 1: adding the new topic types here which were omitted in https://github.com/guardian/mobile-n10n/pull/1773 to enable the local run

api-models/src/main/scala/com/gu/mobile.notifications.client/models/Payloads.scala
```diff
@@ -209,12 +221,88 @@ case class FootballMatchStatusPayload(
...
+    "importance" -> o.importance,
+    "topic" -> o.topic,
+    "matchStatus" -> o.matchStatus,
+    "eventId" -> o.eventId,
+    "kickOffTimestamp" -> o.kickOffTimestamp,
+    "lineupsAvailable" -> o.lineupsAvailable,
+    "detailedMatchStatus" -> o.detailedMatchStatus,
+    "debug" -> o.debug
+  )
+}
```
marjisound, line 268: As we discussed, it’s understandable why we need a separate type for FootballPenaltyShootout, but it may be worth thinking about ways to reduce the duplication here in future work

common/src/main/scala/models/TopicType.scala
```diff
@@ -14,7 +14,9 @@ object TopicTypes {
   object TagSeries extends TopicType("tag-series", 1)
   object TagBlog extends TopicType("tag-blog", 1)
   object FootballTeam extends TopicType("football-team", 1)
+  object FootballTeamLiveActivity extends TopicType("football-team-live-activity", 1)
   object FootballMatch extends TopicType("football-match", 1)
+  object FootballMatchLiveActivity extends TopicType("football-match-live-activity", 1)
```
marjisound, line 19: As discussed, let’s make sure this is necessary here and clarify how it differs from the topic types defined in https://github.com/guardian/mobile-n10n/blob/main/api-models/src/main/scala/com/gu/mobile.notifications.client/models/Topic.scala#L46

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/models/payload/FcmPayloadBuilder.scala
```diff
@@ -123,7 +124,52 @@ object FcmPayloadBuilder {
...
+          Keys.HomeTeamPenaltiesScored -> p.scored.toString,
+          Keys.HomeTeamPenaltiesMissed -> p.missed.toString,
+          Keys.HomeTeamPenaltiesSaved -> p.saved.toString
+        )).getOrElse(Map.empty)
+        ++ n.awayTeamPenalties.map(p => Map(
+          Keys.AwayTeamPenaltiesScored -> p.scored.toString,
+          Keys.AwayTeamPenaltiesMissed -> p.missed.toString,
+          Keys.AwayTeamPenaltiesSaved -> p.saved.toString
+        )).getOrElse(Map.empty),
       ttl = FootballMatchStatusTtl
```
marjisound, line 173: Can we reduce duplication in here. Most of the field in data are the same in footballMatchStatusAndroidNotification & footballPenaltyShootoutAndroidNotification so it'd be good to share them where possible. Though happy if you want to tackle that in a follow-up improvement PR

build.sbt
```diff
@@ -412,7 +412,7 @@ lazy val ecrRepositorySettings =
   }
 
 lazy val notificationworkerlambda = lambda("notificationworkerlambda", "notificationworkerlambda", Some("com.gu.notifications.worker.TopicCounterLocalRun"))
-  .dependsOn(common)
+  .dependsOn(common, football % "compile->compile;test->compile")
```
marjisound: Adding football as a dependency means the entire football package will be included in the worker package, which increases its size. 
Since this is only needed for the local run, could we remove the local run changes from this PR for now so we can avoid adding the dependency here? We can then look into adding the local run file later in a way that doesn’t require pulling in the whole football package as a dependency.

marjisound (approved):
Great work Ara 🎉 and thanks for walking me through the changes. As we already discussed in the call, I left a few comments but feel free to decide whether to address them in this PR or in a follow-up.

notificationworkerlambda/src/main/scala/com/gu/notifications/worker/delivery/fcm/models/payload/FcmPayloadBuilder.scala
aracho1: There is definitely room for reducing duplication but I'll raise it as a follow up PR

PR #1782, opened 2026-05-06
Ensure current time is evaluated on every invocation

marjisound:
This PR fixes an issue related to the current time used in the football lambda. 

The current time was passed as a parameter to the `SyntheticMatchEventGenerator` class, but because `SyntheticMatchEventGenerator` was instansiated as `lazy val`. Lazy val is evaluated only once per JVM/container lifecycle (typically at Lambda cold start), so the timestamp was effectively frozen and did not reflect the actual current time during subsequent Lambda invocations.

This caused time-based conditions on `create-channel` event to behave incorrectly during warm Lambda executions, as comparisons were made against a stale reference time.

The change ensures that the current time is evaluated dynamically at execution time rather than at object initialisation, preventing stale time state and ensuring correct behaviour across all Lambda invocations.

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/Lambda.scala
@@ -49,7 +49,7 @@ object Lambda extends Logging {
 
   lazy val capiClient = GuardianContentClient(configuration.capiApiKey)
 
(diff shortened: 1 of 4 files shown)


vlbee (approved):
great catch!

JamieB-gu (approved):

PR #1787, opened 2026-05-07
Complete live activity life cycle

marjisound:
## What does this change?
This PR is correcting the LiveActivity lifecycle management: 
- Create commonEventBusPusher sbt module in order to share the implementation between football lambda and Channel manager lambda
- Throw error when there's any failedEntryCount on event bus pusher response
- Set `isLive` to false initially when channel is created in Channel manager
- Push a `UpdateLiveActivityEvent` event to event bus straight after channel is created in channel manager
- Set `isLive` to true when a UpdateLiveActivityEvent is received 
- Set `isLive` to false when EndLiveActivityEvent is received 
- Update channel manager policy to allow it to push messages to the event bus
- Support start live activity event in broadcast lambda (this is just a safety measurement since the create channel event also send a broadcast update message. We can remove this support in the future if we find it un-necessary)

### DIFF
--- a/commonEventBusPusher/src/main/scala/com/gu/mobile/liveactivities/event/bus/LiveActivityPusher.scala
+++ b/commonEventBusPusher/src/main/scala/com/gu/mobile/liveactivities/event/bus/LiveActivityPusher.scala
@@ -1,4 +1,4 @@
-package aws
+package com.gu.mobile.liveactivities.event.bus
 
 import com.gu.mobile.notifications.client.models.liveActitivites._
 import org.slf4j.Logger
--- a/liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
@@ -94,11 +94,10 @@ object ChannelManagerLambda extends RequestStreamHandler with Lambda with Loggin
   }
 
   private def pushUpdateLiveActivityEvent(matchId: String, broadcastContentStateData: Option[ContentState]): Future[Unit] = {
-    val triggeringEventId = Some(UUID.nameUUIDFromBytes(s"football-match/$matchId/live-activity-initial-data".getBytes).toString)
-    val derivedId = s"football-match-status/$matchId/$triggeringEventId"
+    val triggeringEventId = s"football-match-status/$matchId/live-activity-initial-data"
     liveActivityPusher.pushToEventbus(
       LiveActivityPayload(
-        id = UUID.nameUUIDFromBytes(derivedId.getBytes),
+        id = UUID.nameUUIDFromBytes(triggeringEventId.getBytes),
         eventType = UpdateLiveActivityEvent,
         liveActivityType = FootballLiveActivity,
         liveActivityID = matchId,
--- a/liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
@@ -4,25 +4,20 @@ import com.amazonaws.services.lambda.runtime.{Context, RequestStreamHandler}
 import com.gu.liveactivities.models.LiveActivityData
 import com.gu.liveactivities.service.ChannelApiClient
 import com.gu.liveactivities.util.Logging
-import com.gu.mobile.notifications.client.models.liveActitivites.{CreateChannelEvent, DeleteChannelEvent, EventBridgeEvent, LiveActivityPayload}
-import play.api.libs.json.{Format, JsError, JsSuccess, Json}
+import com.gu.mobile.notifications.client.models.liveActitivites._
+import play.api.libs.json.{JsError, JsSuccess, Json}
 
 import java.io.{InputStream, OutputStream}
+import java.util.{Date, UUID}
 import scala.concurrent.duration.DurationInt
 import scala.concurrent.{Await, Future}
 import scala.util.{Failure, Success, Try}
-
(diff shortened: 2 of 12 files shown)


commonEventBusPusher/src/main/scala/com/gu/mobile/liveactivities/event/bus/LiveActivityPusher.scala
```diff
@@ -41,13 +28,6 @@ class LiveActivityPusher extends Logging {
     logger.info(
       s"Eventbus pusher: Processing event with id ${payload.id}"
     )
-    val jsonDetail = Json.toJson(payload).toString()
-    if (jsonDetail.isEmpty || jsonDetail == "{}") {
-      logger.info(
-        s"Eventbus pusher: Skipping empty event for ${payload.id}"
-      )
-      Future.successful(())
-    } else {
```
marjisound, line 50: This code is redundant. Since we have `LiveActivityPayload` type for payload, the payload can never be an empty object, so the checking in here is un-necessary

commonEventBusPusher/src/main/scala/com/gu/mobile/liveactivities/event/bus/LiveActivityPusher.scala
```diff
@@ -64,26 +44,27 @@ class LiveActivityPusher extends Logging {
...
-          s"Eventbus pusher: Event published with event ${payload.eventType} is ${payload.id}. Failed entry count: ${response.failedEntryCount()}"
-        )
+        if (response.failedEntryCount() > 0) {
+          val error = response.entries().get(0).errorMessage()
+          throw new RuntimeException(s"EventBridge rejected entry: $error")
+        } else {
+          logger.info(
+            s"Eventbus pusher: Event published with event ${payload.eventType} is ${payload.id}"
+          )
+        }
```
marjisound, line 54: Because this operation handles only a single event, failedEntryCount will effectively be binary (0 or 1). 
Also, we should throw an exception on any failure to ensure the caller is notified and to prevent silent data loss, as a non-zero count indicates the event bus rejected the message.

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
```diff
@@ -95,4 +92,20 @@ object ChannelManagerLambda extends RequestStreamHandler with Lambda with Loggin
         throw exception
     }
   }
+
+  private def pushUpdateLiveActivityEvent(matchId: String, broadcastContentStateData: Option[ContentState]): Future[Unit] = {
+    val triggeringEventId = Some(UUID.nameUUIDFromBytes(s"football-match/$matchId/live-activity-initial-data".getBytes).toString)
```
JamieB-gu: Why does this have to be wrapped in a `Some`?

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
```diff
@@ -95,4 +92,20 @@ object ChannelManagerLambda extends RequestStreamHandler with Lambda with Loggin
...
+    val derivedId = s"football-match-status/$matchId/$triggeringEventId"
```
JamieB-gu: I'm having a bit of trouble understanding the meaning of the different IDs here, sorry! Why do you create two UUIDs and derive the second one from the first?

JamieB-gu (approved):
Looks good. A couple of questions, non-blocking.

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
marjisound: I actually have this question too. The reason I did this was just to keep it consistent with how it's done in the `SyntheticMatchEventGenerator` but I was also wondering why we need to have 2 UUID. Currently there's UUID wrapping the synthetic event ids (e.g. [fullTime](https://github.com/guardian/mobile-n10n/blob/main/football/src/main/scala/com/gu/mobile/notifications/football/lib/SyntheticMatchEventGenerator.scala#L20)) and then another one in the [live activity payload builder ](https://github.com/guardian/mobile-n10n/blob/main/football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusLiveActivityPayloadBuilder.scala#L77)

But now that I think, this must be accidental, and probably also because the normal PA event ids are not UUID, so this is making sure that all id's are ending up as UUID. 

But in the case of this PR, we can certainly remove one of them. Thanks for pointing it out. I also simplified the id and just left it as `s"football-match-status/$matchId/live-activity-initial-data"`

liveactivities/src/main/scala/com/gu/liveactivities/ChannelManagerLambda.scala
marjisound: Good point, this happened accidentally by copy pasting the pattern from synthetic event generator. Fixed now

commonEventBusPusher/src/main/scala/com/gu/mobile/liveactivities/event/bus/LiveActivityPusher.scala
vlbee, line 34: I think this needs to be amended so the source reflects it's coming from football lambda or channel manager, then the equivalent rule updated in the live activities cdk. 

It's working, but might be confusing if eventbusPusher is now a common utility.

vlbee: Or we should change the rules to be from source "liveactivity-stack" and rely on the detail to diff between fooball lambda and other sources.

PR #1795, opened 2026-05-12
fix liveactivities life cycle after end event

marjisound:
## What does this change?
Updated the broadcast logic to gracefully handle update events for matches that have already ended. Instead of failing with a LiveActivityInvalidStateException, the system now treats these as expected "no-ops" to avoid unnecessary Lambda retries and lambda dead letter log noise.

**Changes:**
- Added a check to identify terminal states (!isLive and lastEventId.isDefined).
- If an update is received for an ended match, the Lambda now returns a successful mapping.channelId without calling the Broadcast API or updating the database.
- Moving the broadcast logic into a `BroadcastService` class 
- Write unit tests for BroadcastService covering most of the scenarios up until the `sendToChannel` 

**Why:**
The upstream source may continue to send update events after a Live Activity has technically ended. This change acknowledges this as valid post-match traffic rather than an error scenario.

The main states for live activity:
- **Initial State:** isLive: false, lastEventId: None (Allows first update)

- **Live State:** isLive: true (Allows updates)

- **Ended State:** isLive: false, lastEventId: Some (Blocks further updates)

### DIFF
--- a/liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
@@ -79,33 +79,45 @@ object BroadcastLambda extends RequestStreamHandler with Lambda with Logging {
 
     }
 
-    val broadcastFuture = for {
-      mapping <- repository.getMappingById(matchId)
-      _ <- if (!mapping.isChannelActive) {
-          logger.error(s"Channel not active for match ID $matchId")
-          Future.failed(new LiveActivityInvalidStateException(matchId, "Channel not active"))
-        } else Future.successful(())
+    val broadcastFuture = repository.getMappingById(matchId).flatMap { mapping =>
+      // If the mapping is not live, but it HAS a lastEventId, it means it was Live once and has now Ended.
+      // We want to skip any further updates in this terminal state.
+      // But we don't want to fail the lambda because it's expected that the source may continue
+      // to send update events after a session has ended. Therefore, we treat these as expected no-ops.
+      val broadcastNotAllowed = !shouldEndBroadcast && !mapping.isLive && mapping.lastEventId.isDefined
 
-      _ <- if (mapping.lastEventId.contains(eventId)) {
-          logger.warn(s"Duplicate event ID $eventId for match ID $matchId")
-          Future.failed(new LiveActivityInvalidStateException(matchId, "Duplicate event ID"))
-        } else Future.successful(())
+      if (broadcastNotAllowed) {
+        logger.warn(s"${requestPayload.eventType.asString} event ID $eventId not allowed after ${EndLiveActivityEvent.asString} for match ID $matchId")
+        Future.successful(mapping.channelId)
+      } else {
+        for {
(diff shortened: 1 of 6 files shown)


liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
```diff
@@ -79,44 +79,55 @@ object BroadcastLambda extends RequestStreamHandler with Lambda with Logging {
...
-
-      _ = logger.info(s"Sending broadcast for match ID $matchId to channel ID ${mapping.channelId}")
-      broadcastPayload = BroadcastBody(contentState, shouldEndBroadcast)
-      _ <- broadcastApiClient.sendToChannel(mapping.channelId, None, None, broadcastPayload)
-      _ = logger.info(s"Broadcast ${requestPayload.eventType.asString} sent successfully for match ID $matchId to channel ID ${mapping.channelId}")
-
-      _ <- repository.updateMappingLiveAndLastEvent(matchId, isLive = !shouldEndBroadcast, Some(eventId), Some(eventTime))
-      _ = logger.info(s"Record updated successfully for match ID $matchId")
-    } yield mapping.channelId
+    val broadcastFuture = repository.getMappingById(matchId).flatMap { mapping =>
```
vlbee: There is quite a lot of logic here. I am just wondering if its worth setting up a test?

marjisound: Good point. I added unit tests for all the logic the broadcast lambda up until `sendToChannel`. Happy to do the rest of them in an upcoming PR

liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
```diff
@@ -41,6 +41,8 @@ trait Lambda extends Logging{
       .build()
 
   val repository = new LiveActivityChannelRepository(dynamoDbClient, s"mobile-notifications-liveactivities-${config.stage}")
+  val broadcastApiClient = new BroadcastApiClient(authentication, config.bundleId, config.sendingToProdServer)
```
vlbee: This will bundle in the broadcast service into the channelManager lambda as well via the trait - is there another way to keep cleaner separation? 

I think you might be able to pass into BroadcastLambda as an implicit parameter instead of via the lambda trait which is shared.

vlbee (approved):
Thanks. Tests are great. 

Just a comment about keeping the service out of the shared lambda trait so we don't bloat or blur the lines between the two lambdas.

liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
marjisound: Great point, I had moved them to Lambda without even thinking about channel manager. I now moved both `BroadcastApiClient` & `BroadcastService` to the BroadcastLambda.

vlbee: I've also moved the clean up service out of the clean up lambda with this pattern by the way.

PR #1796, opened 2026-05-12
Add LiveActivity Clean Up scheduled lambda to delete channels

vlbee:
## What does this change?

We need to delete channels in APNS once a broadcast has ended.

This PR adds a scheduled  Lambda to clean up channels for live activities. The primary clean up step which is added here is to delete channels in APNS that are no longer live and created more than 24hours ago. This will cover all activities that have ended, but prevent us from deleting channels that are in the process of being created but have not had a first broadcast yet. 

The lambda is set to run at 3pm via event bridge schedule 

Schedule tested in code:
- [x] deployed and invocation observed on schedule in CODE
- [x] mappings in CODE table marked inactive
- [x] channels in APNS debug sandbox deleted. 

Additional clean up steps that need to be assessed and potentially handled:

```
 deletedChannels <- cleanUpService.deleteChannelsForEndedBroadcasts()

 // orphanChannels <- deleteOrphanChannelsInAPNS()
 /////// channels that exist in APNS but corresponding mapping is missing in Dynamo

 // staleMappings <- markInactiveStaleMappings 
 /////// mapping still "LIVE" that missed an end activity update and need to be deleted - Mapi may return channel id otherwise. 

 // orphanMappings <- markInactiveOrphanMappings
 /////// mapping still "active" but channel has been deleted.
```

## Accessibility


liveactivities/src/main/scala/com/gu/liveactivities/service/ChannelApiClient.scala
```diff
@@ -88,6 +83,42 @@ class ChannelApiClient(authentication: Authentication, bundleId: String, sending
       }
     })
     p.future
-  } 
+  }
+
+  /////////////////////////////////////////////////////////////////
```
vlbee, line 88: This wasn't needed at this stage so can remove if preferred.

JamieB-gu: Do you mean because the only method calling this is `deleteOrphanChannelsInAPNS`, and that's not called itself yet?

liveactivities/src/main/scala/com/gu/liveactivities/ChannelCleanUpLambda.scala
```diff
@@ -0,0 +1,111 @@
...
+        isChannelActive = true,
+        isLive = false,
+        hasLastEvent = true,
+      )
+      _ = logger.info(s"Channels identified for possible deletion: ${channelsForDeletion.size}")
+      _ <- Future.sequence(
+        channelsForDeletion.map { mapping =>
+          {
+            if (mapping.createdAt.isAfter(ZonedDateTime.now().minusHours(24))) {
+              logger.warn(s"Channel with ID ${mapping.channelId} for match ID ${mapping.id} is " +
```
JamieB-gu: Is this a warning? Would we expect this to be a fairly common occurrence, not something that we'd need to look into fixing, and therefore perhaps just an "info" level log?

liveactivities/src/main/scala/com/gu/liveactivities/ChannelCleanUpLambda.scala
```diff
@@ -0,0 +1,111 @@
...
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
+
+class ChannelCleanUpService(
+    repository: ChannelMappingsRepository,
+    channelApiClient: ChannelApiClient,
+)(implicit ec: ExecutionContext) extends Logging {
+
+  def deleteChannelsForEndedBroadcasts(): Future[List[LiveActivityMapping]] = {
```
JamieB-gu: If I've understood correctly, this creates a list of channels that _might_ be deleted (`channelsForDeletion`). Some of them will be deleted, some won't be because they're less than 24 hours old, and some others might not be if there's an exception (which we're recovering from).

Therefore, of that original list, only some of the channels may have been deleted, but it's then returned and assigned to a variable called `deletedChannels`? If we're intending to use that variable somewhere that assumes everything in it has been deleted, could that be an issue? Alternatively, if we're not intending to use the variable, do we need this method to return the list at all? If it's just a side-effect, could it be `Future[Unit]`?

liveactivities/src/test/scala/com/gu/liveactivities/DynamodbSpecification.scala
```diff
@@ -31,13 +21,18 @@ trait DynamodbSpecification extends Specification with BeforeAfterAll {
...
-  override def afterAll(): Unit = {
+  override def before: Unit = {
     LocalDynamoDB.deleteTable(awsClient)(TableName)
+    LocalDynamoDB.createTable(awsClient)(TableName)(targetTableAttributes: _*)
   }
+  override def after: Unit = ()
 
+  override def afterAll(): Unit = {
+    LocalDynamoDB.deleteTable(awsClient)(TableName)
+  }
```
JamieB-gu: Do you want the table created before each test, and deleted after each test ready for the next one? If so, is this enough?

```scala
override def before: Unit = {
  LocalDynamoDB.createTable(awsClient)(TableName)(targetTableAttributes: _*)
}

override def after: Unit = {
  LocalDynamoDB.deleteTable(awsClient)(TableName)
}
```

and you no longer need `beforeAll` and `afterAll`? Or have I missed something, perhaps about the way specs2 parallelises tests?

liveactivities/src/test/scala/com/gu/liveactivities/LiveActivityChannelRepositorySpec.scala
```diff
@@ -72,6 +72,84 @@ class LiveActivityChannelRepositoryTest(implicit ev: ExecutionEnv)
...
+
+      val setup = for {
+        _ <- repository.createMapping(activeNotLiveWithEvent.id, activeNotLiveWithEvent.channelId, activeNotLiveWithEvent.data)
+        _ <- repository.createMapping(activeNotLiveNoEvent.id, activeNotLiveNoEvent.channelId, activeNotLiveNoEvent.data)
+        _ <- repository.createMapping(activeLiveWithEvent.id, activeLiveWithEvent.channelId, activeLiveWithEvent.data)
+        _ <- repository.createMapping(inactiveNotLiveWithEvent.id, inactiveNotLiveWithEvent.channelId, inactiveNotLiveWithEvent.data)
+        _ <- repository.updateMappingLastEvent(activeNotLiveWithEvent.id, Some("event-1"), Some(ZonedDateTime.now()))
+        _ <- repository.updateMappingLiveAndLastEvent(activeLiveWithEvent.id, isLive = true, Some("event-3"), Some(ZonedDateTime.now()))
+        _ <- repository.updateMappingLastEvent(inactiveNotLiveWithEvent.id, Some("event-4"), Some(ZonedDateTime.now()))
+        _ <- repository.updateMappingActiveChannel(inactiveNotLiveWithEvent.id, isActive = false)
```
JamieB-gu, line 91: Why update these mappings with values identical to those that they already contain?

JamieB-gu (commented):
Looks good! A few comments and questions.

liveactivities/src/main/scala/com/gu/liveactivities/ChannelCleanUpLambda.scala
vlbee: I think we should probably take this out completely? it will be very common with the 24hrs so will def change to info.

liveactivities/src/test/scala/com/gu/liveactivities/LiveActivityChannelRepositorySpec.scala
vlbee: because the createMapping defaults to certain boolean values, so they need to be created than updated to the state we want to test them at. 

I might just create a single table at the start then verify all the queries return the correct number of results in one go, rather than set up a table each time.  This will make testing clearer.

liveactivities/src/test/scala/com/gu/liveactivities/DynamodbSpecification.scala
vlbee: I will fix this.

liveactivities/src/main/scala/com/gu/liveactivities/service/ChannelApiClient.scala
vlbee: It's not tested yet, I will set it up in a follow up PR. We believe we could accumulate orphan channels on apply if a channel was created but there was a failure to subsequently write the mapping the Dynamo

liveactivities/src/main/scala/com/gu/liveactivities/ChannelCleanUpLambda.scala
vlbee: I have fixed this so it returns a list of the channels that were actually deleted

liveactivities/src/main/scala/com/gu/liveactivities/ChannelCleanUpLambda.scala
```diff
@@ -0,0 +1,50 @@
...
+      case Success(_)         => Future.successful(())
+      case Failure(exception) =>
+        logger.error(s"Failed to process: ${exception.getMessage}")
+        throw exception
+    }
+  }
+
+  def handleRequest(input: ScheduledEvent, context: Context): Unit = {
+    logger.info("Channel clean up lambda triggered")
+    // todo assess timeout
```
vlbee: @marjisound Could increase to 5min. I can't see this ever taking that long though.

marjisound (approved):
Well done, it's great work. And thanks for walking me through the work and also demoing  it. We went through all possible scenarios in the demo and all is working as expected 👍

PR #1797, opened 2026-05-13
filter matches based on the competitions data in S3

marjisound:
## What does this change?
This PR introduces a filtering mechanism to ensure only matches from supported competitions are processed. The list of supported competitions is now fetched from a JSON configuration file stored in S3.

Tested this in CODE

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
@@ -1,11 +1,11 @@
 package com.gu.mobile.notifications.football.lib
 
-import java.time.ZonedDateTime
-
-import com.gu.mobile.notifications.football.Logging
+import java.time.{LocalDate, ZonedDateTime}
+import com.gu.mobile.notifications.football.{Configuration, Logging}
 import com.gu.mobile.notifications.football.models.RawMatchData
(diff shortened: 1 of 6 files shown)


football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
```diff
@@ -92,13 +107,29 @@ class FootballData(
...
-      _.filter(inProgress)
-       .filter(paProvideAlerts)
-       .filterNot(isMidnight)
-    )
+    for {
+       matches <- paClient.aroundToday(dateTime)
+       competitions <- competitionsDataStore
+         .fetch(s"${stage}/competition/competitions.json")
+         .recover { case exception =>
+           // We don't want to fail the whole process if we can't retrieve the list of competitions,
```
vlbee, line 123: agree

vlbee (approved):
Will this have any impact on push notifications? Were people somehow receiving push notifications for these competitions or were we sending them even though no one was following/receiving them?

marjisound:
> Will this have any impact on push notifications? Were people somehow receiving push notifications for these competitions or were we sending them even though no one was following/receiving them?

Yes that's right. It will affect the notifications. But users couldn't sign up for those notifications as there are no match page for them in the app. So the notification service would send these notifications but no one receives them.

PR #1799, opened 2026-05-15
Add team id to football content state

marjisound:
## What does this change?
Fixing crest in live activity. The device needs to get the team id in the payload in order to load the correct crest for the team in live activity

Tested in CODE


vlbee (approved):
🚀

PR #1805, opened 2026-05-19
Live activities clean up - Fix Dynamo AttemptCalloutException

vlbee:
## What does this change?

We observed a number of live activity mappings failing to complete -> channel was sucessfully deleted on Apple, but attempt to update the `isChannelActive` attribute on Dynamo resulted in time out failures:

```
Unable to execute HTTP request: software.amazon.awssdk.core.exception.ApiCallAttemptTimeoutException: HTTP request execution did not complete before the specified timeout configuration: 2000 millis (SDK Attempt Count: 2)
```

This PR increases the dynamo client `attemptTimeout` and `totalTimeout` used for the Cleaner Lambda dynamo. repository

It also increases the overall lambda timeout to 10min, and the cdk timeout to 11min. 
It also fixes logging in the deletion service that was running syncronously in error. 

We anticipate anywhere from 1 to 75 games per day to delete. The last few days with ~50 games took 22-31seconds overall. 

1/ LiveActivities cleaner lambda deployed to CODE and triggerd manually. 
2/ 10 channels marked inactive and deleted successfully. 
3/ No timeouts observed (though none have been observed in CODE to date, only in PROD) 

## Accessibility

### DIFF
--- a/liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
@@ -24,23 +24,28 @@ trait Lambda extends Logging{
     DefaultCredentialsProvider.builder.build(),
   )
 
-  val dynamoDbClient =
+  private def buildDbClient(attemptTimeout: Int = 2, timeout: Int = 10): DynamoDbAsyncClient = {
     DynamoDbAsyncClient
       .builder()
       .credentialsProvider(credentialsv2)
       .region(Region.of(config.region))
       .overrideConfiguration(
         ClientOverrideConfiguration.builder()
-          .apiCallAttemptTimeout(Duration.ofSeconds(2)) // limit for one single try
-          .apiCallTimeout(Duration.ofSeconds(10)) // limit for the entire operation
+          .apiCallAttemptTimeout(Duration.ofSeconds(attemptTimeout)) // limit for one single try
+          .apiCallTimeout(Duration.ofSeconds(timeout)) // limit for the entire operation
           .retryStrategy(DefaultRetryStrategy.standardStrategyBuilder()
             .maxAttempts(3)
             .build())
           .build()
       )
       .build()
+  }
 
-  val repository = new LiveActivityChannelRepository(dynamoDbClient, s"mobile-notifications-liveactivities-${config.stage}")
+  private val dynamoTableName = s"mobile-notifications-liveactivities-${config.stage}"
+  val repository = new LiveActivityChannelRepository(buildDbClient(), dynamoTableName)
+  def getRepositoryWithCustomTimeouts(attemptTimeout: Int, timeout: Int): LiveActivityChannelRepository = {
+    new LiveActivityChannelRepository(buildDbClient(attemptTimeout, timeout), dynamoTableName)
+  }
 
   private val eventBusName = s"liveactivities-eventbus-${config.stage}"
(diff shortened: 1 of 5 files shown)


liveactivities/src/main/scala/com/gu/liveactivities/service/ChannelCleanUpService.scala
```diff
@@ -36,14 +36,19 @@ class ChannelCleanUpService(
               .closeChannel(mapping.channelId)
               .flatMap { _ =>
                 val f = repository.updateMappingActiveChannel(mapping.id, isActive = false)
-                channelsDeleted ++= List(mapping)
-                logger.info(s"Channel successfully marked inactive with channel ID: ${mapping.channelId} for match ID ${mapping.id}")
```
vlbee, line 40: This was running synchronously and logging by mistake.

marjisound: Good catch

liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
```diff
@@ -24,23 +24,28 @@ trait Lambda extends Logging{
...
     DynamoDbAsyncClient
       .builder()
       .credentialsProvider(credentialsv2)
       .region(Region.of(config.region))
       .overrideConfiguration(
         ClientOverrideConfiguration.builder()
-          .apiCallAttemptTimeout(Duration.ofSeconds(2)) // limit for one single try
-          .apiCallTimeout(Duration.ofSeconds(10)) // limit for the entire operation
+          .apiCallAttemptTimeout(Duration.ofSeconds(attemptTimeout)) // limit for one single try
+          .apiCallTimeout(Duration.ofSeconds(timeout)) // limit for the entire operation
```
marjisound, line 35: Nice 👍

liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
```diff
@@ -24,23 +24,28 @@ trait Lambda extends Logging{
...
       )
       .build()
+  }
 
-  val repository = new LiveActivityChannelRepository(dynamoDbClient, s"mobile-notifications-liveactivities-${config.stage}")
+  private val dynamoTableName = s"mobile-notifications-liveactivities-${config.stage}"
+  val repository = new LiveActivityChannelRepository(buildDbClient(), dynamoTableName)
+  def getRepositoryWithCustomTimeouts(attemptTimeout: Int, timeout: Int): LiveActivityChannelRepository = {
+    new LiveActivityChannelRepository(buildDbClient(attemptTimeout, timeout), dynamoTableName)
+  }
```
marjisound, line 48: I'm just wondering how come we're getting timeout errors when clean up lambda is querying the table but not when the other 2 lambdas query the table?! Could it be because the clean up lambda is retrieving a list of records while the other 2 are just getting 1 mapping?

marjisound (approved):
Looks fine to me, just left a question to keep an eye on this in the future if other lambdas might experience this timeout and weather we may need to increase the timeout for them too.

liveactivities/src/main/scala/com/gu/liveactivities/Lambda.scala
vlbee: The clean up lamba performs a scan of everything with a filter. It's a small table 100 items. But may grow to 2-400 with the 2 week ttl. I doubt it will ever get bigger than that.

PR #1807, opened 2026-05-20
Add Dynamo optimistic locking strategy to updateMappingById

vlbee:
## What does this change?

In production we observed multiple cases where an End-Broadcast setting `isLive = false` was processed simultaneously as a normal update setting  the field back `isLive = true` - this meant that MAPI would return channel IDs for ended matches and that the clean up lambda would not select these matches for channel deletion. 

As a this appears caused by concurrent lambdas overwriting the dynamo record, this PR introduces a strategy of Dynamo 'optimistic locking' whereby _before_ any updated is made, the update method checks the `lastModifiedAt` value and only writes the update if that has not changed.

If the lastModifiedAt condition fails, it indicates another process has since modified the record and the write will fail to be tried again. 

TBC.

## Accessibility


liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -88,14 +88,16 @@ class LiveActivityChannelRepository(client: DynamoDbAsyncClient, tableName: Stri
...
 
-  private def updateMappingById(
+  private[liveactivities] def updateMappingById(
       id: String,
       isChannelActive: Option[Boolean] = None,
       isLive: Option[Boolean] = None,
       lastEventId: Option[String] = None,
-      lastEventAt: Option[ZonedDateTime] = None
+      lastEventAt: Option[ZonedDateTime] = None,
+      expectedLastModifiedAt: Option[ZonedDateTime] = None, // ONLY for testing
```
vlbee, line 97: I am not sure if this is good idea, or how to test this othersie.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -107,25 +109,39 @@ class LiveActivityChannelRepository(client: DynamoDbAsyncClient, tableName: Stri
...
+        val result = for {
+          lastModifiedAt <- expectedLastModifiedAt
+            .map(Future.successful)
+            .getOrElse(getMappingById(id).map(_.lastModifiedAt))
+          result <- scanamo.exec {
+            val modifiedAt = ZonedDateTime.now()
+            val upsWithLastModified = set("lastModifiedAt", dateTimeToString(modifiedAt)) :: ups
+            val lastModifiedAtCondition = Condition("lastModifiedAt" === dateTimeToString(lastModifiedAt))
+            table
+              .when(lastModifiedAtCondition)
```
vlbee, line 122: I think perhaps we could move fetching the lastModifiedAt value to the exposed method (eg. updateMappingLive, updateMappingActiveChannel) insted of the the base method updateMappingById.

liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -107,25 +109,39 @@ class LiveActivityChannelRepository(client: DynamoDbAsyncClient, tableName: Stri
...
-              }"
-            )
-            Future.successful(())
-          case Left(ex) =>
-            val errorMsg = s"Error updating live activity mapping for id $id - ${DynamoReadError.describe(ex)}"
-            logger.error(errorMsg)
-            Future.failed(new RepositoryException(errorMsg))
-        }.recover(handleErrors("updating mapping in DynamoDB", id))
+        val result = for {
+          lastModifiedAt <- expectedLastModifiedAt
```
vlbee, line 114: Should we increase the lambda Timeouts for this?

PR #1808, opened 2026-05-20
Remove unused batch sending code

JamieB-gu:
To send Android notifications, the worker lambda sends requests to Firebase. A few years ago, Firebase introduced a multicast/batching API, allowing consumers to group multiple notifications into a single request. The team experimented to see if it would improve performance:

- https://github.com/guardian/mobile-n10n/pull/747

Some time later, Firebase dropped support for this API, and recommended using HTTP/2 instead. The team therefore switched back to sending requests individually, and removed much of the batching code:

- https://github.com/guardian/mobile-n10n/pull/1216
- https://github.com/guardian/mobile-n10n/pull/1235

There is still some unused code left over, so this change removes most of what remains.


vlbee (approved):
delete delete delete!

PR #1809, opened 2026-05-21
Filter out "League" from competition round string.

vlbee:
## What does this change?

We don't want "League" appearing twice but we do want world cup grouping and general Round info (semi final, fina, play off, etc) appearing. This PR filters out this field if it's just a string matching "League"

image
image

## Accessibility

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusLiveActivityPayloadBuilder.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusLiveActivityPayloadBuilder.scala
@@ -57,7 +57,7 @@ class MatchStatusLiveActivityPayloadBuilder {
       competition = Competition(
         id = matchInfo.competition.map(_.id).getOrElse(""),
         name = matchInfo.competition.map(_.name).getOrElse(""),
-        round = matchInfo.round.name // World Cup Group
+        round = matchInfo.round.name.filter(_ != "League") // Round includes world cup grouping, "semi-final" etc.
       ),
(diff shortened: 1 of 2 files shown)


football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusLiveActivityPayloadBuilder.scala
```diff
@@ -57,7 +57,7 @@ class MatchStatusLiveActivityPayloadBuilder {
       competition = Competition(
         id = matchInfo.competition.map(_.id).getOrElse(""),
         name = matchInfo.competition.map(_.name).getOrElse(""),
-        round = matchInfo.round.name // World Cup Group
+        round = matchInfo.round.name.filter(_ != "League") // Round includes world cup grouping, "semi-final" etc.
```
markjamesbutler, line 60: Do you know how the group stage for the world cup would appear here? I assume it would not come through with the league name. Although some people may refer to it this way the correct terminology would be group stage likely with the group letter. I.e. Group A

vlbee: We can already see the match info for the world cup matches and the Group comes in as the Round:

vlbee: The round says "semi-final", "final" etc for the knockout matches.

ippily (approved):

PR #1812, opened 2026-05-21
Stop End-Broadcasts and Update-Broadcasts being sent together. 

vlbee:
## What was the problem? 

End live activity synthetic event and full-time synthetic event were both generated when the PA `matchInfo.result` becomes `true` so we were sending both MatchPhaseEvents to live activities in the same polling cycle. The full-time event was potentially overwriting the end-broadcast event  in the Dynamo causing isLive to return to true even though the match had ended. 

Furthermore, certain edge cases, eg. goal scored in the last minute of game time or during a penalty shootout, there can be a triggering update event (goal) and an end event generated in the same polling cycle that can cause race conditions and are not processed sequentially. We need to sequence updates and end-events for the same match to prevent this. 

These issues are due to the fact that 1/ we use polling which processes events for the same match in batches, 2/ we use eventbridge bus routing to deliver to different lambdas but does not guarantee order, particularly observed for events that occur in the same cycle.  Another solution could potentially be to migrate to sqs and eventbridge pipes. 

## What has changed?

This PR:
- Stops `"full-time"` synthetic event from being processed for Live Activites, relying ONLY on 'end-live-activity' synthetic event.
- Make `"end-live-activitiy"` synthetic event generation more robust by basing this on `result=true` AND `isLive=false`
- Stops an end activity event from being processed at all if it is not received alone in a single polling cycle - we do this by filtering out non-isolate end events during duplicate event filtering.  

To achieve this last goal, we need to make an additional Dynamo get call to check if an event already exists, but these should not be too bad as we first dedupe from an in-memory cache. However, the PushNotification table is provisioned so we should keep an eye on this with this change - Perhaps worth changing this to on demand as well? 

 

`EventFilter` unit tests added. 

Tested in CODE with match 4551558 (penalties) and normal match (4551093):

- [x] end-broadcast synthetic event is now generated when a match has a result AND is no longer live - this seems like a safer check as we observed in 15/5 match 4551953 that "ended" before penalties were played - I believe this was a random PA discrepancy/error. 
- [x] "full-time" synthetic event is filtered out so that we don't get duplicate FullTime payloads 
- [x] ifLive=false is not being overwritten to true in either penalty shootout or 90min type of match
- [x] for Penalty shootout match (4551558), we see full-time update and end-broadcast payloads sent sequentially.
- [x] for average 90min match (4551093), we see sequentially last end-broadcast payload with a full-time status - UI is updated with final state, channel mapping is successfully marked isLive=false.

## Accessibility


cdk/lib/footballnotificationslambda.ts
```diff
@@ -106,21 +106,16 @@ export class FootballNotificationsLambda extends GuStack {
...
+				actions: [
+					'dynamodb:PutItem',
+					'dynamodb:UpdateItem',
+					'dynamodb:Query',
+					'dynamodb:GetItem',
 				],
-			}),
-		);
-
-		footballnotificationslambda.addToRolePolicy(
```
vlbee, line 118: This appears to be a redundant policy covered by the permissions in the policy above.

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala
```diff
@@ -36,7 +42,67 @@ class EventFilter[A <: Payload, D](distinctCheck: DynamoDistinctCheck[A, D]) ext
...
+        * have been processed (dispatched via eventbridge). This only affects Live Activities.
+        */
+      eventsToProcess = filterOutEndEventsNotReceivedInIsolation(newEvents)
+      processedEvents <- Future.traverse(eventsToProcess) { item =>
+        distinctCheck.insertEvent(ev(item)).map {
+          case Distinct => Some(item)
+          case _        => None
+        }
+      }
+      _ = logger.debug(
```
vlbee: to be deleted

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala
```diff
@@ -36,7 +42,65 @@ class EventFilter[A <: Payload, D](distinctCheck: DynamoDistinctCheck[A, D]) ext
...
+    for {
+      newEvents <- filterAsync(dynamoEvents)(item => isDuplicateRecord(ev(item)).map(!_))
+      /** Because we poll once a minute, we might end up with a triggering update event (eg. very late goal) along with
+        * an end event in the same polling cycle, but we only ever want to process the end event alone after all updates
+        * have been processed (dispatched via eventbridge). This only affects Live Activities.
+        */
+      eventsToProcess = filterOutEndEventsNotReceivedInIsolation(newEvents)
+      processedEvents <- Future.traverse(eventsToProcess) { item =>
+        distinctCheck.insertEvent(ev(item)).map {
+          case Distinct => Some(item)
```
marjisound: Can we not do the local caching at this point `cache(item.id)` so that we only cache when we add an item to dynamo and we don't need to decache any more?

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala
```diff
@@ -36,7 +42,65 @@ class EventFilter[A <: Payload, D](distinctCheck: DynamoDistinctCheck[A, D]) ext
...
+        if (!isDup) {
+          cache(item.id)
+        }
+        isDup
+      }
+    } else {
+      logger.debug(s"Event ${item.id} already exists in local cache or does not have an id - discarding (dynamo table: ${distinctCheck.tableName})")
+      Future.successful(true)
+    }
+  }
```
marjisound, line 46: Although it completely makes sence to check if it's duplicated, but another approach here could be to check if it's unique, so that the function returns true when an item is not in dynamo already

football/src/main/scala/com/gu/mobile/notifications/football/lib/EventFilter.scala
```diff
@@ -36,7 +42,65 @@ class EventFilter[A <: Payload, D](distinctCheck: DynamoDistinctCheck[A, D]) ext
...
+
+    isolatedEndEvents ++ updateEvents
+  }
+
+  // we need to be able to access liveActivityId so the Payload trait must be narrowed. "An instance of A <:< B witnesses that A is a subtype of B."
+  def filterDynamoEventsForLiveActivities(
+      dynamoEvents: List[LiveActivityPayload],
+  )(implicit ec: ExecutionContext, ev: LiveActivityPayload <:< A): Future[List[LiveActivityPayload]] = {
+    for {
+      newEvents <- filterAsync(dynamoEvents)(item => isDuplicateRecord(ev(item)).map(!_))
```
marjisound: If we change `isDuplicateRecord` into `isUniqueRecord`, we should be able to avoid the `map(!_)`

marjisound (approved):
Looks good, left a few none-blocking comments, the most important one is if we could change where we do the caching in order to avoid de-caching

vlbee:
Thank you for your feedback @marjisound 

I have changed the logic to use `isUnique` and also moved the decaching. 

I have deployed football and tested again on CODE:

Normal Match: 4552042 / isLive=false
image

Penalties Match: 4551800 / isLive=false
image

PR #1813, opened 2026-05-22
Log PA football `matchDay.result` and `matchStatus`

aracho1:
## What does this change?
To test the suspicion that the `matchDay.result` boolean field in the PA football matchDay data gets set to true before the match is fully over (including the penalties), add logging for:
- matchStatus
- result
- liveWatch


vlbee:
Could we also log the isLive boolean? 
It would is interesting to log the match status, result and isLive boolean all together in this log line. And do log again when Result = True AND isLive = false.

vlbee (approved):

PR #1817, opened 2026-05-26
Generate pre-match football push notifications

aracho1:
## What does this change?
- Generate a push notification 20 minutes before a match kick off using the existing `koWithin20Minutes` helper
- Android builds which have the latest live updates beta will display the kick off information
- Old Android builds with the old notifications UI will display "Pre" as the match status

Deployed to football CODE and used football time machine and tested that around 20 minutes before a match kick off, a push notification is received to indicate the kick off starting soon
- [x] Android latest beta
- [x] iOS push notifications
- [x] Android older builds

prematch push notifications are sent correctly

- Tested

## Images
Android with latest live updates UI (displays kick off time in the prematch specific UI)
Screenshot_20260609_162644

Android with old UI (shows "Pre" as the match status since it doesn't have prematch specific UI)
Screenshot_20260609_165932

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/lib/SyntheticMatchEventGenerator.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/lib/SyntheticMatchEventGenerator.scala
@@ -18,6 +18,15 @@ class SyntheticMatchEventGenerator(getCurrentTime: () => ZonedDateTime) {
 
   private type MatchEventGenerator = (MatchDay, List[pa.MatchEvent]) => Option[MatchEvent]
 
+  // generate a push notification 20 minutes before the scheduled kick off
+  private val preMatch: MatchEventGenerator = { (matchDay: MatchDay, _) =>
+    if (koWithin20Minutes(matchDay.date.toEpochSecond)) Some(emptyMatchEvent.copy(
+      id = Some(UUID.nameUUIDFromBytes(s"football-match/${matchDay.id}/pre-match".getBytes).toString),
+      eventType = "pre-match"
+    ))
+    else None
+  }
+
   private val fullTime: MatchEventGenerator = { (matchDay: MatchDay, matchEvents: List[pa.MatchEvent]) =>
     if (matchDay.result) Some(emptyMatchEvent.copy(
       id = Some(UUID.nameUUIDFromBytes(s"football-match/${matchDay.id}/full-time".getBytes).toString),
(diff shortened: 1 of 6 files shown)


football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusNotificationBuilder.scala
```diff
@@ -193,7 +193,8 @@ class MatchStatusNotificationBuilder(mapiHost: String) {
     case SecondHalf(_) => "Second-half start"
     case FullTime(_) => "Full-Time"
     case _:Dismissal => "Red card"
-    case _:PenaltyShootoutKick  => "Penalty Kick" // needed for spec. We are filtering these out in the EventConsumer for now.
```
aracho1, line 202: cleaned up the comment since we're not filtering out the penalty kicks in push notifications since implementing [this](https://github.com/guardian/mobile-n10n/commit/e5e80628f21d21cd1d06bec337eab2c34e279362)

football/src/main/scala/com/gu/mobile/notifications/football/lib/SyntheticMatchEventGenerator.scala
```diff
@@ -157,6 +157,14 @@ class SyntheticMatchEventGenerator(getCurrentTime: () => ZonedDateTime) {
     else None
   }
 
+  private val preMatch: MatchEventGenerator = { (matchDay: MatchDay, _) =>
```
vlbee: Can we add a comment why this exists? And also move it to before create-channel in all places these are listed?

football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusNotificationBuilder.scala
```diff
@@ -41,7 +41,10 @@ class MatchStatusNotificationBuilder(mapiHost: String) {
     val penaltyShootoutKicks = allEvents.collect { case psr: PenaltyShootoutKick => psr }
     val penaltyShootoutScore = PenaltyShootoutScore.fromPenaltyShootoutKicks(matchInfo.homeTeam, matchInfo.awayTeam, penaltyShootoutKicks)
 
-    val status = statuses.getOrElse(matchInfo.matchStatus, matchInfo.matchStatus)
+    val status = triggeringEvent match {
```
vlbee: Why can't we use the status map below for this?

```
 ("Fixture", "Pre"),
 ("-", "Pre"), 
```

football/src/test/scala/com/gu/mobile/notifications/football/lib/SyntheticMatchEventGeneratorSpec.scala
```diff
@@ -142,7 +142,7 @@ class SyntheticMatchEventGeneratorSpec extends Specification {
 
     "Add a startLiveActivity event if match kick off is within 20min" in new TestScope {
       val generator = new SyntheticMatchEventGenerator(currentTime)
-      generator.generate(List(), "match-id", matchInfo.copy(date = ZonedDateTime.now.plusMinutes(15))).head.eventType mustEqual "start-live-activity"
+      generator.generate(List(), "match-id", matchInfo.copy(date = ZonedDateTime.now.plusMinutes(15))).map(_.eventType) must contain("start-live-activity")
```
aracho1, line 145: The generator now also fires when kick-off is within 20 minutes, and it's now first in the generators list - so `.head` returns pre-match instead of start-live-activity.  Update the test to check that the event includes 'start-live-activity' rather than it being the first one on the list.

michaelbjacobson (approved):
Looks great to me!

vlbee (approved):

PR #1821, opened 2026-06-01
Simplify creation of penalty score push notification payloads in Football Lamba

vlbee:
## What does this change?

https://github.com/guardian/mobile-n10n/pull/1779/ recently was merged allowing us to send penalty push notifications to some Android devices. This functionality depends on pattern matching on the Notification`type` [here](https://github.com/guardian/mobile-n10n/blob/b88736a1d26486ff09850aadea0f4825ff7ac5c9/common/src/main/scala/models/Notification.scala#L43)

However, in the football lambda we were creating two different Payload classes rather than just differentiating the payload type so this PR attempts to make the Football `MatchStatusNotificationBuilder` more DRY so it can be more easily iterated on. 

- [x] Deployed football lambda to CODE and rabn recent match with penalities - observed expected notifications on Android emulator. 

image
image
image

## Accessibility

### DIFF
--- a/api-models/src/main/scala/com/gu/mobile.notifications.client/models/Payloads.scala
+++ b/api-models/src/main/scala/com/gu/mobile.notifications.client/models/Payloads.scala
@@ -101,7 +101,6 @@ object NotificationPayload {
       case n: BreakingNewsPayload => BreakingNewsPayload.jf.writes(n)
       case n: ContentAlertPayload => ContentAlertPayload.jf.writes(n)
       case n: FootballMatchStatusPayload => FootballMatchStatusPayload.jf.writes(n)
-      case n: FootballPenaltyShootoutPayload => FootballPenaltyShootoutPayload.jf.writes(n)
     }
   }
 }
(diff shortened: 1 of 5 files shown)


aracho1 (dismissed):
Looks great, would be good to test in CODE before merging

api-models/src/main/scala/com/gu/mobile.notifications.client/models/Payloads.scala
```diff
@@ -300,11 +231,12 @@ case class FootballPenaltyShootoutPayload(
   debug: Boolean,
   dryRun: Option[Boolean]
 ) extends NotificationPayload with derivedId {
-  val `type` = FootballPenaltyShootout
-  override val derivedId = s"football-penalty-shootout/$matchId/$eventId"
+  private val hasPenalties = homeTeamPenalties.isDefined || awayTeamPenalties.isDefined
+  val `type` = if (hasPenalties) FootballPenaltyShootout else FootballMatchStatus
+  override val derivedId = if (hasPenalties) s"football-penalty-shootout/$matchId/$eventId" else s"football-match-status/$matchId/$eventId"
```
aracho1, line 236: If we derive type and detailedMatchStatus derives from penaltyShootoutScore, once a match has any penalty kick, when an FT event fires, the type and detailedMatchStatus will stay as penalties. 

To get around this, the type/detailedMatchStatus should depend on whether triggeringEvent is itself a PenaltyShootoutKick (existing behaviour). 

The downside of this is that the final FT notification doesn't show any penalty shoot scores, since homeTeamPenalties/awayTeamPenalties aren't part of FootballMatchStatusNotification currently but arguably getting the correct matchStatus is more important.

Since this is a refactoring PR, I'd suggest keeping the existing behaviour and leaving the penalty score gap in FT as a follow up, but let me know what you think.

PR #1827, opened 2026-06-09
Add isEnabled flag

vlbee:
Set ssm parameter for prod before merging. 

## What does this change?

Adds a config flag that stops both the channel manager and broadcast lambdas from processing events.  Clean up lambda will still run. 

New channels will not be created. End events to delete channels will not be processed so activites already live will not receive an end broadcast and go stale. Apples auto dismissal policy is max 8hr per for each live activity on the dynamic island, and 4hrs on the lock screent (worst case 12hrs total from start) 

Tested in CODE:

1. Switching channel mapping to isLive=false for single match
- [x] Broadcast updates stopped being sent.
- [x] Logs observed: `"broadcast-update event ID 81da91bc-ac3a-36f3-83c2-8c1f4e937901 not allowed after broadcast-end for match ID 4550484"` 
- [x] Football live activity payload matching UUID recorded in deduping table.

2. Switching config isEnabled flag to false
- [x] logs observed: `"Received channel-create event for match ID 4550484 but channel creation is disabled by config"`
- [x] new match channels not created. 
- [x] logs observed: `"Broadcast lambda is currently disabled by config. Skipping broadcast with id 224309f8-0dde-334e-8f30-1eca6b7fe305 for match ID 4550484"`
- [x] broadcasts not sent for a live match
- [x] Football live activity payloads recorded for deduping. 

## Accessibility

### DIFF
--- a/liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
+++ b/liveactivities/src/main/scala/com/gu/liveactivities/BroadcastLambda.scala
@@ -7,7 +7,7 @@ import com.gu.mobile.notifications.client.models.liveActitivites._
 import play.api.libs.json._
 
 import java.io.{InputStream, OutputStream}
-import scala.concurrent.Await
+import scala.concurrent.{Await, Future}
 import scala.concurrent.duration.DurationInt
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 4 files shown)


marjisound (approved):
Looks good 👍

PR #1828, opened 2026-06-12
Revert "Log PA football `matchDay.result` and `matchStatus`"

aracho1:
Remove the PA data logging which was added in guardian/mobile-n10n#1813 which is no longer needed and which may be contributing to our AWS bill


michaelbjacobson (approved):
👍🏼

PR #1833, opened 2026-06-17
Logs and comment clean up

vlbee:
## What does this change?

Deletes out of date todo comments and logs originally added for debugging. 

## Accessibility


liveactivities/src/main/scala/com/gu/liveactivities/service/LiveActivityChannelRepository.scala
```diff
@@ -115,11 +115,6 @@ class LiveActivityChannelRepository(client: DynamoDbAsyncClient, tableName: Stri
 
         result.flatMap {
           case Right(_) =>
-            logger.info(
```
vlbee, line 118: This was added to debug why ended matches were staying "Live". It just logs the dynamo writes for each request so we could understand the race condition that was happening.

aracho1 (approved):

PR #1836, opened 2026-06-18
Filter out invalid topics from registration requests

aracho1:
## What does this change?
At the moment, Registration rejects any requests that contain even a single invalid topic. Instead, filter out the invalid ones and subscribe to the valid ones. 

1. Deploy this branch to registrations CODE
2. Send a curl request to the CODE registrations with a topics list containing an invalid topic
```
curl -X POST https://notifications.code.dev-guardianapis.com/device/register \
    -H "Content-Type: application/json" \
    -H "User-Agent: ac-test" \
    -d '{
      "deviceToken": "ac-test-2026-06-18",
      "platform": "android",
      "topics": [
        {"type": "breaking", "name": "uk"},
        {"type": "breaking", "name": "us"},
        {"type": "breaking", "name": "international"},
        {"type": "tag", "name": "sport/live/2026/jun/06/england-v-new-zealand-first-mens-cricket-test-day-three-live"}
      ]
    }'
```
3. Tunnel into the registrations DB and check that only the valid topics have been subscribed
image

Reduce the number of 4xx requests

This could increase the number of writes to the registrations DB since we're no longer rejecting requests that contain an invalid topic. This however should be remediated when the Android team will release a change to stop sending these invalid topic types.


common/src/main/scala/models/Registration.scala
```diff
@@ -21,17 +21,17 @@ object Registration {
...
         }
-        val all = (json \ "topics").asOpt[JsArray].toSeq.flatMap(_.value.map(topicSummary))
-        val invalid = (json \ "topics").asOpt[JsArray].toSeq.flatMap { arr =>
-          arr.value.flatMap { v =>
-            if (v.validate[Topic].isError) Some(topicSummary(v)) else None
-          }
-        }
-        if (invalid.nonEmpty) {
-          MDC.put("invalidTopics", invalid.mkString(", "))
-          if (all.nonEmpty) MDC.put("allTopics", all.mkString(", "))
```
vlbee, line 32: Do we know why the whole request was rejected before? Any old PR?

vlbee (commented):
Have we checked the DB read write provisioned capacity that this change won't  cause any concerns there? How many requests are we rejecting right now? I am assuming it's quite minimal.

aracho1:
> Have we checked the DB read write provisioned capacity that this change won't cause any concerns there? How many requests are we rejecting right now? I am assuming it's quite minimal.

Since the registrations DB is an RDS, there isn't a single capacity unit but I've created a [Cloudwatch dashboard](https://201359054765-ghidelkn.eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#dashboards/dashboard/n10n-registrations?start=PT672H&end=null) with some key metrics and they seem to indicate that we're nowhere near the max (e.g. CPU utilisation averages around 6%). I will keep an eye out when we release this.

vlbee (approved):

PR #1838, opened 2026-06-19
Add LiveActivities metrics

vlbee:
## What does this change?

Adds a metrics class with liveactivity lambda namespaces. 

Adds metrics for:
APNS Responses to track 2xx/5xx/4xx and network erros
Our various "do not process" scenarios - these messages will still end up in the DLQ for investigation but we can see how frequently they occur over time with metrics. 

This will allow us to easily spot problems talking to APNS, as well as the frequency overtime of outlier behaviour we are handling.

Ran in CODE. 
Metrics observed in Cloudwatch. 

## Accessibility


liveactivities/src/main/scala/com/gu/liveactivities/util/Metrics.scala
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+object Metrics {
+  // Lambda names used to scope the CloudWatch namespace to a specific lambda
+  val BroadcastLambdaName = "broadcast"
+  val ChannelManagerLambdaName = "channel-manager"
+  val ChannelCleanUpLambdaName = "channel-cleanup"
+
+  val BroadcastProcessed = "BroadcastProcessed"
```
vlbee: Add more data to these metrics

liveactivities/src/main/scala/com/gu/liveactivities/service/ChannelApiClient.scala
```diff
@@ -42,15 +41,19 @@ class ChannelApiClient(authentication: Authentication, bundleId: String, sending
     val p = Promise[String]()
     httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofString()).whenComplete((response, err) => {
       if (response == null) {
+        logger.error(s"Failed to create channel due to error ${err.getMessage}")
+        metrics.recordApnsNetworkError()
```
aracho1, line 45: Even though sendAsync is used for the APNS call, I think `CloudWatchClient.putMetricData` is synchronous and blocks the `whenComplete` thread for handling the APNS response until CloudWatch responds. Have you considered Consider using [CloudWatchAsyncClient](https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/cloudwatch/CloudWatchAsyncClient.html)

aracho1 (approved):

liveactivities/src/main/scala/com/gu/liveactivities/service/ChannelApiClient.scala
vlbee: Metrics are fire once after the broadcast to APNS goes out so there no  impact on the service if metrics are fired synchronously.  

For the channel creation its a very low volume metric count so would only fire once per invocation and not during a live event so again the extra time would be negligible. 

For channel deletions, this is a chron job unrelated to the service performance so it doesn't need to be the fastest. 

Overall it's a very low volume service and there is also a risk that the main lambda process terminates before the async cloud watch thread is finished and we loose metrics on a separate thread. 

For these reasons I think firing synchronously is fine.

PR #1852, opened 2026-07-13
Generate push notifications for additional match phases including extra time and penalties

aracho1:
## What does this change?
PA doesn't send an explicit event when a goal is overturned (e.g. by VAR) - the score correction only becomes visible to us via the next event we happen to generate a notification for. We saw this cause ncorrect scores to be shown to users for extended periods during recent World Cup matches, since these phase-transition events (extra time starting, penalties starting, etc.) were previously filtered out entirely and generated no notifications.

This PR removes the filtering on these phase events so that we get more frequent "checkpoints" during a match (extra time, penalties) where a score correction can reach users, reducing howc long a stale/incorrect score can persist.

**Impact on old clients**: 
Android builds before May 20th 2026 (when the new live updates UI was released) cannot display the 'detailedMatchStatus' field (the field used for the additional phase events) and can only display the pre-existing 'matchStatus' field. 

This means that while they will receive notifications for the new match statuses e.g. extra time to be played / extra time 1st, 2nd half, their notification can only display "ET" or "PT".

Deployed to CODE and tested against recent World Cup matches

## Images
| New match status | Latest Android builds* | Old Android builds** |
|------------------|------------------------|----------------------|
| Extra time to be played |  |  |
| Extra time 1st half |  |  |
| Extra time HT |  |  |
| Extra time 2nd half |  |  |
| Penalties to be played | I didn't manage to screenshot it in time but confirmed with CODE DB record  |  |
| Penalties |  |  |

\* Builds after May 20th 2026 when the new live updates UI was released.  
\** Builds before May 20th 2026.

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/lib/EventConsumer.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/lib/EventConsumer.scala
@@ -30,13 +30,6 @@ class EventConsumer(
         // additional synthetic live activity life cycle events
         "create-channel",
         "end-live-activity",
-        // additional synthetic match phase events for live activities
-        "extra-time-to-be-played",
-        "extra-time-first-half",
-        "extra-time-half-time",
-        "extra-time-second-half",
-        "penalties-to-be-played",
-        "penalties",
         "suspended",
         "resumed",
         "abandoned",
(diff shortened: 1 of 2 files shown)


vlbee (approved):
Looks good.

PR #1854, opened 2026-07-14
Remove midnight KO filtering on World Cup matches 

aracho1:
## What does this change?
- No live activities or push notifications were generated for Portugal vs Croatia which was scheduled for July 3, 2026 00:00 (midnight) kick off. 
- This is because we have a logic to filter out matches with 00:00 KO because PA feed sometimes uses 00:00 as a placeholder for when it doesn't know a match's kickoff time yet. PA does not have an alternative field that can be used reliably for confirmed vs estimated KO time.
- Competitions like World Cup have matches that span multiple time zones, and they publish their schedule months in advance, so midnight kickoffs are likely to be genuine.
- So this PR adds an allowlist so midnight KOs for listed competitions are no longer filtered out.

- Deployed to CODE and used football time machine to set the clock to around the Portugal vs Croatia match
- Tested that the match no longer gets filtered out and generates notification successfully

Android notification
image

Logs
image

### DIFF
--- a/football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
+++ b/football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
@@ -135,7 +135,7 @@ class FootballData(
   private val competitionsAllowingMidnightKO = Set(
     "700", // FIFA World Cup
   )
-  private[lib] def isMidnight(matchDay: MatchDay): Boolean = {
+  private[lib] def isFakeMidnightKO(matchDay: MatchDay): Boolean = {
     val isAllowlisted = matchDay.competition.exists(c => competitionsAllowingMidnightKO.contains(c.id))
     !isAllowlisted && matchDay.date.toLocalTime == LocalTime.MIDNIGHT
   }
@@ -163,7 +163,7 @@ class FootballData(
       matchesInSupportedCompetitions
         .filter(inProgress)
         .filter(paProvideAlerts)
-        .filterNot(isMidnight)
+        .filterNot(isFakeMidnightKO)
     }
   }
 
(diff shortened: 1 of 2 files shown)


football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
```diff
@@ -124,17 +124,26 @@ class FootballData(
...
+   * PA give 00:00 as the start time when they don't actually know the kickoff time yet,
+   * so those matches are unusable and should be filtered out. Competitions in
+   * `competitionsAllowingMidnightKO` are excluded from this check because a 00:00 kickoff can be
+   * genuine for them (e.g. the World Cup, whose schedule spans multiple time zones and is
+   * published months in advance).
+   */
+  private val competitionsAllowingMidnightKO = Set(
+    "700", // FIFA World Cup
+  )
+  private[lib] def isMidnight(matchDay: MatchDay): Boolean = {
```
vlbee: Maybe this could be renamed to `isFakeMidnightKickOff` for a bit of clarity when just reading the code superficially.

vlbee (approved):
🕛

vlbee:
Do we want to go ahead and include other world cup and euro cup competition ids here?

football/src/main/scala/com/gu/mobile/notifications/football/lib/FootballData.scala
aracho1: good idea, I'll change it

aracho1:
> Do we want to go ahead and include other world cup and euro cup competition ids here?

That's a good shout, though I'm inclined to leave out European competitions e.g. Euros for now as they are unlikely to have matches with midnight KO. From [this MAPI list of supported PA competitions](https://github.com/guardian/mobile-apps-api/blob/main/common-pa-feeds/src/main/scala/com/gu/mobile/football/data/pa/PaCompetitions.scala), the safe ones seem to be World Cup qualifying and Copa America, so I'll add these two.

PR #1860, opened 2026-07-22
Alternative VAR solution: Generate a payload for deleted match event goals

vlbee:
PA has indicated they may be able to add a `deleted=true` attribute flag to reversed goals on the `/match/events` endpoint we use.  More info here: https://github.com/guardian/pa-football-client/pull/123

## What does this change?

This PR tests the [Preview release](https://github.com/guardian/pa-football-client/pull/123) of the `pa-football-client` library. 

- It uses the deleted flag to disregard deleted goals when calculating goals for a paylaod
- It used the deleted flag to generate `dervidedId` for the payload UUID - this ensures that when the deleted flag changes, the same triggering event will have a unique UUID and NOT be filtered out by the event filter so we can send a follow up payload for a deleted event. 
- Parity with push notifications builder - we send "No Goal!", "Red card overturned", or "Penalty overturned" message on iOS when a deleted event payload is created. 

#### TODO before release

- Update the `pa-football-client` library

TODO 
- [ ] code testing
- [ ] e2e inegration test.

On unit tests, we can see the payload with unique UUID generated for a `deleted=true` goal at '37.

#### Live Activities
Left: `deleted=false`                  Right: `deleted=true`
image

#### Push Notifications
Left: `deleted=false`                  Right: `deleted=true`
image

## Accessibility


football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusNotificationBuilder.scala
```diff
@@ -64,7 +64,7 @@ class MatchStatusNotificationBuilder(mapiHost: String) {
      importance = importance(triggeringEvent),
      topic = topics,
      matchStatus = status,
-     eventId = UUID.nameUUIDFromBytes(triggeringEvent.eventId.getBytes).toString,
+     eventId = s"${triggeringEvent.eventId}/${triggeringEvent.isDeleted}", // used to derive the uuid in the payload trait
```
vlbee, line 67: The payload UUID is derived from the `trait Payload` derivedId def [here](https://github.com/guardian/mobile-n10n/blob/dae37651983dd92a4933bfd66e4c55588e98f3dc/api-models/src/main/scala/com/gu/mobile.notifications.client/models/Payloads.scala#L307-L309), so I don't think it's necessary to make a UUID from the event ID (which is either the PA eventID, or a UUID we already create once for synthetic events.)

build.sbt
```diff
@@ -43,7 +43,7 @@ val awsSdk2Version: String = "2.42.34"
 val doobieVersion: String = "0.13.4"
 val catsVersion: String = "2.13.0"
 val okHttpVersion: String = "4.12.0"
-val paClientVersion: String = "7.0.12"
+val paClientVersion: String = "7.0.18-PREVIEW.parse-deleted-event.2026-07-21T1518.accf3da8"
```
vlbee, line 46: Use the proper release before merging.

PR #1868, opened 2026-08-17
Ignore `status=deleted` events when calculating score, redcards, and penalites

vlbee:
## What does this change?

Depends on pa-client library release here - https://github.com/guardian/pa-football-client/pull/128

We calculated score, red cards, and penalty kick score by counting the number of relevant events in the PA match event timeline. This change uses a updated library release to ignore those events, so that when PA begin retaining "deleted" events in the timeline, our match state calculations are not impacted. 

A payload will be built for an event that is marked as "deleted", but the UUID for that payload will NOT change, so it should be deduplicated and not sent as an update. 

We will change this behaviour in a subsequent PR to be able to up date VAR reversals quickly, by changing the UUID of a "active" -> "deleted" event.

Unit tests have been added. 

CODE - tested match 4558721 and works as expected (no change) prior to PA  new status attribute change.
- [x] Android sim - 
- [x] iOS debug app - 
- [x] LiveActivity payloads logged as expected in Dynamo
- [x] PushNotifications logged as expected in Dynamo

## Accessibility


marjisound (commented):
Great work, as you already mentioned, it'd be great if we could have more unit tests to verify the deduplication of goal payload. 

I also have a question that I couldn't yet figure out how it's happening in the code. 
How does the deduplication is working? We use eventId to deduplicate data but when an event get `deleted` status, I think it'll keep the same id. I'm wondering if that would **stop** the event from being pushed?! But don't we want to push an event when that happens to reverse the goal status we had previously sent to user?

It might be best t have a walk through and chat about this in a call. Lets see when's the best time tomorrow

vlbee:
> Great work, as you already mentioned, it'd be great if we could have more unit tests to verify the deduplication of goal payload.
> 
> I also have a question that I couldn't yet figure out how it's happening in the code. How does the deduplication is working? We use eventId to deduplicate data but when an event get `deleted` status, I think it'll keep the same id. I'm wondering if that would **stop** the event from being pushed?! But don't we want to push an event when that happens to reverse the goal status we had previously sent to user?
> 
> It might be best t have a walk through and chat about this in a call. Lets see when's the best time tomorrow

yes, that is exactly the plan  to trigger a broadcast when an event is being sent (PA draft here - https://github.com/guardian/mobile-n10n/pull/1869/changes)

I though it best to just release a defensive PR first, to make sure nothing changes when the API response changes, then I can effectively test the real fix in CODE before it goes out.

football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusLiveActivityPayloadBuilder.scala
vlbee: does a status=deleted change the PA match event id?

marjisound (approved):

football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusNotificationBuilder.scala
aracho1: We should filter out deleted events here too as otherwise overturned goal scorer will appear in the text

football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusLiveActivityPayloadBuilder.scala
vlbee: Per the code example in the PA advisory email, the event id is not changing

football/src/main/scala/com/gu/mobile/notifications/football/notificationbuilders/MatchStatusNotificationBuilder.scala
vlbee: My understanding is this message is only for when the actual push notification for the triggering event goes out, BEFORE it is flagged as deleted.  In subsequent push notification for future triggering events, like another goal, this will be the scorer for that next goal.

vlbee: My understanding is this message is only for when the actual push notification for the triggering event goes out, BEFORE it is flagged as deleted. In subsequent push notification for future triggering events, like another goal, this will be the scorer for that next goal.

vlbee:
The example in PA advisory email indicate the event ID will stay the same even if the status changes.

PR #1882, opened 2026-09-18
Retry on FailedRequest dont error out

joshuaguardian:
## What does this change?

Transient APNs transport failures were being reported without retrying, including:

> Stream closed before write could take place

`FailedRequest` extends `Throwable` via `NoStackTrace`, so it did not match the existing generic `Exception` retry case.

This change explicitly marks `FailedRequest` as retryable using the existing retry policy with jitter and exponential backoff. It also adds regression coverage proving that a request succeeds when its first attempt fails transiently.

- Added a test covering a `FailedRequest` followed by a successful retry.
- `sbt notificationworkerlambda/test`: 38 tests passed.
- The repository-wide `sbt test` was attempted, but unrelated DynamoDB tests failed because the native `sqlite4java` library could not load locally.

Monitor sending metrics and logs after deployment.

We expect:

- Fewer final `DeliveryException$FailedRequest` errors.
- Fewer failed sends during bursts of HTTP/2 stream closures.
- A more consistent delivery success rate across notifications.

Transient failures may still appear in retry logs, but they should no longer be counted as final failures when a later attempt succeeds.

Retrying increases request volume briefly during APNs connection failures. The existing randomized exponential backoff limits synchronized retry pressure.

There is a small possibility that APNs accepted a notification but its response was lost, causing a retry and potential duplicate delivery. This is an inherent risk when retrying after an ambiguous transport failure.

The project currently uses Pushy `0.13.10`. Upgrading it should be considered separately to pick up newer HTTP/2 and connection-management fixes.

## Images

Not applicable. This change has no UI impact.

## Accessibility

Not applicable. This change has no user-interface impact.

- [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
- [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
- [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
- [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)


vlbee:
Thank you for this work. Upgrading Pushy is one of the next tasks the team is picking up so we can assess if there are any improvements to the connection pooling we can make. 

I think best to mitigate the risk of duplicate delivery when the FailedRequest error comes from an APNS client time out. Perhaps we could add a guard in the pattern matching to only retry a FailedRequest when it's a stream closure or not a ClientTimeout?

